PR #51, opened 2017-05-04
MainWindow.vala: GObject-style construct

danrabbit:


src/MainWindow.vala
```diff
@@ -59,14 +59,16 @@ namespace Scratch {
         delegate void HookFunc ();
 
         public MainWindow (Scratch.ScratchApp scratch_app) {
+            Object (application: scratch_app,
+                    icon_name: "accessories-text-editor",
+                    title: app.app_cmd_name);
             this.app = scratch_app;
```
Philip-Scott: Might be crashing due to this?

PR #65, opened 2017-05-05
Re-namespace Scratch.ScratchApp to Application

danrabbit:


ricotz:
So this broke the API of libscratchcore. Any thoughts on how to handle it? e.g. bumping the so-version or ignore it?

danrabbit:
Oops, thanks for catching that @ricotz. I guess the right thing to do would be to bump the so-version right? But I guess there's probably not any third-party extensions right now so probably up to you to decide :)

tintou:
Well, now that it's done, let's break it even more and bump the .so version at release time

PR #73, opened 2017-05-07
Remove line break option

danrabbit:
Since we're focusing in on code editing this option is not necessary


danrabbit:
Oops, you're right! Good catch :)

cassidyjames:
Uh, for the record line wrapping is a very common code editor feature and one I use in Scratch.

tintou:
Okay, seems like I tested the wrong executable, I just need this option to be always on, why would people even want to code with an horizontal scrollbar

PR #80, opened 2017-05-09
Enable Wrap mode by default

tintou:


jeremypw:
Does not compile:
```SourceTreePlugin.vala:222.13-222.30: error: The name `set_sort_func' does not exist in the context of `Granite.Widgets.SourceList'
            view.set_sort_func ((a, b) => {```

tintou:
Because of #86

jeremypw:
Does not compile:

> Document.vala:72.53-72.97: error: The name `data_home_folder_unsaved' does not exist in the context of `Scratch.Application'
                return file.get_path ().has_prefix (Application.instance.data_home_folder_unsaved);

> DocumentView.vala:97.20-97.64: error: The name `data_home_folder_unsaved' does not exist in the context of `Scratch.Application'
            return Application.instance.data_home_folder_unsaved + new_text_file;

PR #82, opened 2017-05-11
Use RDNN for install dirs

danrabbit:
Hopefully fix package builds by RDNNing all the things


tintou:
After this we definitely need to change the API version

PR #85, opened 2017-05-17
Add screenshot

cassidyjames:
Not sure if we want that much of a description or naw :man_shrugging:


danrabbit:
Seems a little overkill. This stuff is in the AppData file

PR #89, opened 2017-06-12
Fix #2 - handle save same name

### LINKED ISSUE #2: "Save as" destroys data when trying to overwrite a file
To reproduce:

Create one file with "Well, then", save as test.txt, and close.
Create another file with "Hello there" and attempt to overwrite test.txt.

You will see "Well, then" in your editor, and "Hello there" has been destroyed (no backup either).

This, in combination with a resurfacing of bug #1392031, caused me to lose about an hour of work. Ouchies.

Workaround: ALWAYS use a new filename when saving.

Launchpad Details: [#LP1668014](https://bugs.launchpad.net/bugs/1668014) David Berg - 2017-02-26 08:01:56 +0000

jeremypw:
Fix race condition causing content destined to be overwritten to be loaded into the overwriting document, destroying data therein.

Also confirm before overwrite.

See #2


Philip-Scott:
not a review, but i still think that this should be handled on the file dialog to be able to get this for free on all apps

jeremypw:
Do you mean Pantheon Filechooser module should *always* have do_overwrite_confirmation set to "true" so applications have no choice (or have to turn it off)?

jeremypw:
Added high priority label as it is associated with possible user data loss.

PR #105, opened 2017-06-15
Implement add file and folder

matzipan:
Depends on #104 

The lines for `newly_created_path` are currently commented out as they will be implemented in a further pull request.

PR #106, opened 2017-06-15
Folderitem monitor improvements

matzipan:
Depends on #105


jeremypw:
Blocked pending resolution of conflicts.

matzipan:
Things fixed by this branch:
* When a selected item is deleted, the sourcelist doesn't select another item (which causes another tab to open).
* When a new file or folder is created through the context menu, it gets selected and marked as editable for text input.
* It makes extensive use of `children_loaded`. A folder is not fully loaded until it is first expanded, thus, if it hasn't been loaded yet, it will do [some magic](https://github.com/elementary/scratch/pull/106/files#diff-b61b726ad8aa4cbb3d762fc573d0f224R118) to reflect the expandability state correctly (the folder can only be expanded if it has at least one child and cannot be expanded if it has no children). This fixes an issue where if you create a folder on the outside and then create a file in it, it doesn't show that folder as being expandable.

plugins/folder-manager/FileView.vala
```diff
@@ -27,15 +27,31 @@ namespace Scratch.Plugins.FolderManager {
...
         public signal void select (string file);
 
-        public FileView () {
-            this.width_request = 180;
-            this.item_selected.connect ((item) => {
-                select ((item as FileItem).path);
-            });
+        // This is a workaround for SourceList silliness: you cannot remove an item
+        // without it automatically selecting another one.
+        public bool ignore_next_select = false;
```
donadigo: Can we have this as a property?
`public bool ignore_next_select { get; set; default = false }`

plugins/folder-manager/FolderItem.vala
```diff
@@ -93,46 +98,108 @@ namespace Scratch.Plugins.FolderManager {
...
+                            add (item);
+
+                            if (source.get_path () == newly_created_path) {
+                                newly_created_path = null;
+
+                                /*
+                                 * Avoid race condition between adding and editing folder item
+                                 * (not required for file items).
+                                 */
+                                GLib.Idle.add(() => {
```
donadigo: Spaces between `add` and `(`.

plugins/folder-manager/FolderItem.vala
```diff
@@ -97,46 +102,108 @@ namespace Scratch.Plugins.FolderManager {
...
+                    case GLib.FileMonitorEvent.CREATED:
+                        if (source.query_exists () == false) {
+                            return;
+                        }
+
+                        var file = new File (source.get_path ());
+                        var exists = false;
+                        foreach (var item in children) {
+                            if ((item as Item).path == file.path) {
+                                exists = true;
```
donadigo: Since this is copied, could you break here to not go further in this `foreach`?

PR #107, opened 2017-06-15
Only show valid files as children

matzipan:


plugins/folder-manager/File.vala
```diff
@@ -145,11 +145,12 @@ namespace Scratch.Plugins.FolderManager {
         }
 
         // returns a list of all children of a directory
-        GLib.List <File> _children = null;
+        GLib.List <File>? _children = null;
```
jeremypw: I don't think you can make GList nullable.  Later you append to this list without "creating" it.  A null list is equivalent to an empty list.  This mean that for a folder that has no valid children, the folder to be searched every time "children" is accessed. Is that intended?   Maybe you should use something creatable/nullable like Gee.ArrayList instead?

matzipan: You're right about the nullable part. I am changing it back. I believe the original intent was to have this as a caching of sorts. So the `if(_children != null)` intends to check if the list has been generated yet.

I'm not sure whether to use `Gee.ArrayList` or `Gee.LinkedList`?

matzipan: I'm guessing since folder structure isn't exactly highly mutable data, I'm just gonna use an `ArrayList`

jeremypw: Yes, I was going to say it depends on what you are going to use it for - whether the data will change or need to be sorted/reordered etc.

matzipan: I'm looking at this and starting to think there isn't even a need for memoization here, as method only ever gets called once. And even in branch #106 when it gets called, it only gets called after being invalidated.

matzipan:
I've copied the is_valid_directory method over from #108 to avoid rebasing the whole thing.

plugins/folder-manager/File.vala
```diff
@@ -161,7 +137,11 @@ namespace Scratch.Plugins.FolderManager {
                     var file_info = new FileInfo ();
                     while ((file_info = enumerator.next_file ()) != null) {
                         var child = parent.get_child (file_info.get_name ());
-                        _children.append (new File (child.get_path ()));
+                        var file = new File (child.get_path ());
+
+                        if (file.is_valid_directory || file.is_valid_textfile) {
+                            _children.add (new File (child.get_path ()));
```
jeremypw: This will run every time the property "children" is accessed - is that intended?  You have lost the check for a null List object at the start of the get clause.  If the children are refreshed on every access then you need to empty or recreate the list each time.

matzipan: No. I messed up. Sorry for this. Fixing it now.

matzipan:
I've made the changes requested. The rescanning is implemented using the monitor in FolderItem although thinking of it this probably belongs to File.

PR #108, opened 2017-06-15
Simplify file and folder validity code

matzipan:
Removing lines 103-115 also fixes the bug which doesn't show empty folders.


jeremypw:
I notice that hidden files are always invalid.  Is it intended to have a "show-hidden" option at some point?

matzipan:
@jeremypw I guess it's worthy a #ux discussion. I've opened #111

jeremypw:
fwiw, gedit has this option.

PR #113, opened 2017-06-17
Cleanup context menu for FileItem

matzipan:
Moving the `trash` method to `Item` as it will be shared by `FolderItem` too.

### DIFF
--- a/plugins/folder-manager/FileItem.vala
+++ b/plugins/folder-manager/FileItem.vala
@@ -30,9 +30,7 @@ namespace Scratch.Plugins.FolderManager {
 
         public override Gtk.Menu? get_context_menu () {
             var trash_item = new Gtk.MenuItem.with_label (_("Move to Trash"));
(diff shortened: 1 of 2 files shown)


plugins/folder-manager/FileItem.vala
```diff
@@ -21,31 +21,23 @@
...
-            menu.show_all ();
-
-            item_trash.activate.connect (() => {
+            var trash_item = new Gtk.MenuItem.with_label (_("Move to Trash"));
+            trash_item.activate.connect (() => {
                 file.trash ();
             });
+            menu.append (trash_item);
+
+            menu.show_all ();
```
danrabbit: This is now messier. It should be like this:

```
var trash_item = new Gtk.MenuItem.with_label (_("Move to Trash"));
trash_item.activate.connect (() => {
    file.trash ();
});

var menu = new Gtk.Menu ();
menu.append (trash_item);
menu.show_all ();
```

PR #115, opened 2017-06-17
Implement folder context menu

matzipan:


plugins/folder-manager/FolderItem.vala
```diff
@@ -58,25 +52,24 @@ namespace Scratch.Plugins.FolderManager {
...
+            menu.append (item);
+
+            var rename_item = new Gtk.MenuItem.with_label (_("Rename"));
+            rename_item.activate.connect (() => view.start_editing_item (this));
+            menu.append (rename_item);
+
+            var trash_item = new Gtk.MenuItem.with_label (_("Move to Trash"));
+            trash_item.activate.connect (trash);
+            menu.append (trash_item);
+
```
danrabbit: can we get this like in `FileItem` so that all of the methods for `menu` are grouped? As in:

```
var menu = new Gtk.Menu ();
menu.append (item);
menu.append (rename_item);
menu.append (trash_item);
```

plugins/folder-manager/FolderItem.vala
```diff
@@ -58,25 +52,24 @@ namespace Scratch.Plugins.FolderManager {
...
-                try {
-		            FileOutputStream os = new_file.create (FileCreateFlags.NONE);
-	            } catch (Error e) {
-		            warning ("Error: %s\n", e.message);
-	            }
-            });
+        public override Gtk.Menu? get_context_menu () {
+            var menu = new Gtk.Menu ();
+
+            var item = new Gtk.MenuItem.with_label (_("Open"));
```
danrabbit: probably should be `open_item` for clarity

plugins/folder-manager/FolderItem.vala
```diff
@@ -58,25 +52,25 @@ namespace Scratch.Plugins.FolderManager {
...
-                var new_file = GLib.File.new_for_path (file.path + "/new File");
-
-                try {
-		            FileOutputStream os = new_file.create (FileCreateFlags.NONE);
-	            } catch (Error e) {
-		            warning ("Error: %s\n", e.message);
-	            }
-            });
+        public override Gtk.Menu? get_context_menu () {
+            var open_item = new Gtk.MenuItem.with_label (_("Open"));
```
danrabbit: Hm I wonder if we need a different label for this. I assumed that it was "Show in File Browser", but it actually adds this as a top level folder to the sidebar.

Is that a valuable action to take? I can't think of a good case for it right now

matzipan: Now that I think of it you're right. It doesn't make that much sense. I think this was a leftover from merging with the filemanager plugin, where I guess it made sense.

danrabbit: Ah okay that makes more sense. Yeah, my personal thought is probably to remove this

PR #143, opened 2017-07-01
Drop highlight line pref

### LINKED ISSUE #132: Drop "Highlight current line" preference
...and default to on.

It's such a subtle feature that I think having it always on would be useful for those who like it, and entirely ignorable for those who don't.

Up for discussion, though.

cassidyjames:
To be merged if #142 goes well. :stuck_out_tongue: Fixes #132.


danrabbit:
Marking this as "blocked" since we want to trial the new default a while and see if anyone complains

cassidyjames:
It seems like we haven't heard anything about #142, so we should rebase and merge this in.

cassidyjames:
Should be ready for review.

PR #166, opened 2017-08-17
Document.vala: code cleanups

Philip-Scott:


src/Services/Document.vala
```diff
@@ -27,70 +27,61 @@ namespace Scratch.Services {
...
-
-        // Widgets
-        public Scratch.Widgets.SourceView source_view;
-        private Gtk.ScrolledWindow scroll;
-        private Gtk.InfoBar info_bar;
-
-#if GTKSOURCEVIEW_3_18
-        private Gtk.SourceMap source_map;
-#endif
+        public bool is_file_temporary {
```
matzipan: This should be moved to the File class

src/Services/Document.vala
```diff
@@ -27,70 +27,61 @@ namespace Scratch.Services {
...
 
-        // Objects
         private Gtk.SourceFile source_file;
         public File file {
-            get { return source_file.location; }
+            get {
+                return source_file.location;
+            }
             set {
                 source_file.set_location (value);
```
matzipan: source_file.location = value?

src/Services/Document.vala
Philip-Scott: The GLib.File class? :P

PR #168, opened 2017-08-17
Toolbar.vala: misc cleanups

danrabbit:
* Update license header
* GObject-style construction
* Don't use ToolItems where possible
* DRY

### DIFF
--- a/src/Widgets/ToolBar.vala
+++ b/src/Widgets/ToolBar.vala
@@ -79,13 +79,8 @@ namespace Scratch.Widgets {
 
             show_all ();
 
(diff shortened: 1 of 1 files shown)


src/Widgets/ToolBar.vala
```diff
@@ -51,70 +52,58 @@ namespace Scratch.Widgets {
...
 
-            // Some signals...
+            share_menu.insert.connect (() => {
+                on_share_menu_changed ();
+            });
+
+            share_menu.remove.connect (() => {
+                on_share_menu_changed ();
+            });
+
```
jeremypw: and here

PR #170, opened 2017-08-18
Delete Source Tree plugin

danrabbit:
This seems like a shittier duplicate of folder manager


danrabbit:
Marking this as "blocked" for the above reason

danrabbit:
@jeremypw Do you have steps to reproduce the crash? I can't seem to reproduce it here

PR #174, opened 2017-08-18
Replace sidebar notebook with a stack

danrabbit:
Working towards the icon tab design like the old mockups.

screenshot from 2017-08-20 09 47 33


jeremypw:
Suggest using an interface between plugins and sidebar widget rather specifying the exact widget.  This would simplify future changes.

donadigo:
Installed the branch, but instead of icons I get labels, did I miss something?

danrabbit:
@donadigo No, I believe the icon name property is new for gtk 3.20, so on 3.18 it will show the title label

matzipan:
Is this blocked?

danrabbit:
Uh no? Blocked by what?

matzipan:
My apologies, I read "I believe the icon name property is new for gtk 3.20, so on 3.18 it will show the title label" as blocked.

PR #194, opened 2017-08-23
Convert app menu to use glib.action

danrabbit:
Adds a SimpleActionGroup and converts the app menu from using the deprecated Gtk.Action to GLib.action

### DIFF
--- a/src/Widgets/ToolBar.vala
+++ b/src/Widgets/ToolBar.vala
@@ -78,13 +78,13 @@ namespace Scratch.Widgets {
             share_app_menu.set_popup (share_menu);
 
             var new_view_menuitem = new Gtk.MenuItem.with_label (_("Add New View"));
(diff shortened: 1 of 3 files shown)


src/Widgets/ToolBar.vala
```diff
@@ -108,7 +117,8 @@ namespace Scratch.Widgets {
 
             settings.changed.connect (() => {
                 save_button.visible = !settings.autosave;
-                zoom_default.visible = Application.instance.get_last_window ().get_default_font_size () != Application.instance.get_last_window ().get_current_font_size ();
+                var last_window = Application.instance.get_last_window ();
+                zoom_default.visible = last_window.get_default_font_size () != last_window.get_current_font_size ();
```
jeremypw: This change seems unrelated to the PR description?

PR #196, opened 2017-08-24
Convert Toolbar actions to GLib.Action

danrabbit:
- [x] Open
- [x] Templates
- [x] Save
- [x] Save As
- [x] Revert
- [x] Find
- [x] Zoom


davidmhewitt:
All actions are now converted and I've tested all of them. Was unable to test the "Templates" action because I can't even figure out how to trigger it on stable.

It also seems like the "Remove Current View" menu item is broken, but this was broken in master.

I'm also unsure as to how reliable the delete whitespace at end of file on save plugin will be after this change, but I'm also unsure how useful that plugin is considering it doesn't even delete blank lines from the end of files, it only deletes literal whitespace.

danrabbit:
I can't approve since I started the PR, but I've tested these changes and it seems like everything still works as expected.

I can confirm that "remove current view" is already broken in master

PR #197, opened 2017-08-24
Convert some shortcuts to GLib.Action

danrabbit:


matzipan:
Why just some?

danrabbit:
Because these don't require changes to a lot of other files

src/MainWindow.vala
```diff
@@ -425,8 +446,8 @@ namespace Scratch {
             var fetch = (Gtk.ToggleAction) main_actions.get_action ("ShowFetch");
             fetch.sensitive = val;
             fetch.active = (fetch.active && val);
-            main_actions.get_action ("ShowGoTo").sensitive = val;
-            main_actions.get_action ("ShowReplace").sensitive = val;
+            ((SimpleAction) actions.lookup_action (ACTION_GO_TO)).set_enabled (val);
```
davidmhewitt: Can we have a method called `get_action` or something that does `actions.lookup_action (x)` and returns it as a `SimpleAction`. We'll need to use that quite a few times and it ought to make the code look a bit cleaner.

PR #198, opened 2017-08-24
Remove unnecessary GLib.File creation

matzipan:
`parent` is pointing to the same path as `file` and there's no reason why they cannot be the same object.


matzipan:
To verify this branch simply use the folder manager sidebar and see if the correct folder children are displayed in the tree.

PR #203, opened 2017-08-29
Brackets completion preserves selection

### LINKED ISSUE #31: Brackets completion should also un-complete
If I type an opening bracket/paren/quote, then backspace it, the Brackets Completion plugin should also removed its added closing bracket/paren/quote. Right now it just leaves it.

Atom and Sublime have great behavior here if we need to investigate edge cases.

Launchpad Details: [#LP1547635](https://bugs.launchpad.net/bugs/1547635) Cassidy James Blaede - 2016-02-19 18:55:34 +0000

### LINKED ISSUE #151: Quotation marks appended when paste
I find out that when I paste some html, quotation marks are appended at the end of text in number equal to ( quotation marks in text )/2.

It appens also with angular brackets.

screenshot from 2017-07-11 12 03 06

Auto-completion problem?

### LINKED ISSUE #183: Bracket Completion should preserve selection
The current plugin destroys whatever is selected, replacing with the inputted brackets. The expected behavior, as per other text/code editors, is to preserve the selection.

Attached sample of how it looks in other editors
bracket

### LINKED ISSUE #215: Don't complete brackets in front of letters
Like in Sublime Code /Scratch shouldnt complete brackets if cursor is left of letters and numbers.

OttoRobba:
Fixes #183 and fixes #151. EDIT: Also fixes #31 and fixes #215 now, thanks to @snowparrot :+1: 

I'm not 100% happy with this, I wish I could make into a single action - right now it takes two undo commands to revert the completion.

I'm open for input on how to make this better - it is my first time using Vala.

### DIFF
--- a/plugins/brackets-completion/brackets-completion.vala
+++ b/plugins/brackets-completion/brackets-completion.vala
@@ -73,41 +73,49 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
     public void deactivate () {
     }
 
-    string get_next_char () {
+    string get_next_char () { // This breaks on the last character
         Gtk.TextIter start, end;
 
         current_buffer.get_selection_bounds (out start, out end);
         end.forward_char ();
-        return current_buffer.get_text (start, end, true) ;
+        string current_text = current_buffer.get_text (start, end, true) ;
+        int len = current_text.length;
+        return current_text[len - 1:len];
     }
 
-    void on_backspace () {
+    string get_previous_char () { // This breaks when it is first character
         Gtk.TextIter start, end;
 
         current_buffer.get_selection_bounds (out start, out end);
-        current_selection = current_buffer.get_text (start, end, true);
+        start.backward_char ();
+        string current_text = current_buffer.get_text (start, end, true);
+        int len = current_text.length;
 
-        if (current_selection.length == 0) {
-            start.backward_char ();
-            var left_char = current_buffer.get_text (start, end, true);
+        return current_text[0:1];
+    }
 
-            current_buffer.get_selection_bounds (out start, out end);
-            end.forward_char ();
-            var right_char = current_buffer.get_text (start, end, true);
+    void on_backspace () {
+        if (!current_buffer.has_selection) {
+            var left_char = get_previous_char ();
+            var right_char = get_next_char ();
+
             if (left_char in opening_brackets && right_char in closing_brackets) {
+                Gtk.TextIter start, end;
+
+                current_buffer.get_selection_bounds (out start, out end);
                 start.backward_char ();
+                end.forward_char ();
                 current_buffer.select_range (start, end);
             }
         }
     }
 
-    void complete_brackets () {
+    void complete_brackets (string opening_bracket) {
         Gtk.TextIter start, end;
 
         current_buffer.get_selection_bounds (out start, out end);
         current_selection = current_buffer.get_text (start, end, true);
 
-        string opening_bracket = keys[event.keyval];
         string closing_bracket = brackets[opening_bracket];
         string text = opening_bracket + current_selection + closing_bracket;
 
@@ -124,11 +132,11 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
     }
 
     bool on_key_press (Gdk.EventKey event) {
-        string bracket = keys[event.keyval];
         if (current_view.has_focus && event.keyval in keys) {
-            if (keys[event.keyval] in opening_brackets) {
-                complete_brackets ();
-            } else if (keys[event.keyval] in closing_brackets) {
+            string bracket = keys[event.keyval];
+            if (bracket in opening_brackets) {
+                complete_brackets (bracket);
+            } else if (bracket in closing_brackets) {
                 print("nope");
             }
 
--- a/plugins/brackets-completion/brackets-completion.vala
+++ b/plugins/brackets-completion/brackets-completion.vala
@@ -141,7 +141,6 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
     }
 
     bool on_key_press (Gdk.EventKey event) {
-        print("I happened");
         if (event.keyval in keys &&
             !(Gdk.ModifierType.MOD1_MASK in event.state) &&
             !(Gdk.ModifierType.CONTROL_MASK in event.state)) {
--- a/plugins/brackets-completion/brackets-completion.vala
+++ b/plugins/brackets-completion/brackets-completion.vala
@@ -137,9 +137,9 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
     }
 
     bool on_key_press (Gdk.EventKey event) {
-        if (event.keyval in keys
-        && !(Gdk.ModifierType.MOD1_MASK in event.state)
-        && !(Gdk.ModifierType.CONTROL_MASK in event.state)) {
+        if (event.keyval in keys &&
+            !(Gdk.ModifierType.MOD1_MASK in event.state) &&
+            !(Gdk.ModifierType.CONTROL_MASK in event.state)) {
             string bracket = keys[event.keyval];
             string next_char = get_next_char ();
 
--- a/plugins/brackets-completion/brackets-completion.vala
+++ b/plugins/brackets-completion/brackets-completion.vala
@@ -26,8 +26,7 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
     Gee.HashMap<uint, string> keys;
     const string[] valid_next_chars = {"", " ", "\b", "\r", "\n", "\t"};
 
-    Gtk.TextBuffer current_buffer;
-    Gtk.SourceView current_view;
+    Gtk.TextBuffer buffer;
 
     Scratch.Services.Interface plugins;
     public Object object { owned get; construct; }
@@ -55,7 +54,6 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
         keys[Gdk.Key.grave] = "`";
 
         plugins = (Scratch.Services.Interface) object;
-
         plugins.hook_document.connect (on_hook_document);
     }
 
@@ -64,74 +62,73 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
     }
 
     void on_hook_document (Scratch.Services.Document doc) {
-        current_buffer = doc.source_view.buffer;
-        current_view = doc.source_view;
+        buffer = doc.source_view.buffer;
 
-        current_view.backspace.connect (on_backspace);
-        current_view.key_press_event.connect (on_key_press);
+        doc.source_view.backspace.connect (on_backspace);
+        doc.source_view.key_press_event.connect (on_key_press);
     }
 
     string get_next_char () {
         Gtk.TextIter start, end;
 
-        current_buffer.get_selection_bounds (out start, out end);
+        buffer.get_selection_bounds (out start, out end);
         end.forward_char ();
 
-        return current_buffer.get_text (start, end, true) ;
+        return buffer.get_text (start, end, true) ;
     }
 
     string get_previous_char () {
         Gtk.TextIter start, end;
 
-        current_buffer.get_selection_bounds (out start, out end);
+        buffer.get_selection_bounds (out start, out end);
         start.backward_char ();
 
-        return current_buffer.get_text (start, end, true);
+        return buffer.get_text (start, end, true);
     }
 
     void on_backspace () {
-        if (!current_buffer.has_selection) {
+        if (!buffer.has_selection) {
             string left_char = get_previous_char ();
             string right_char = get_next_char ();
 
             if (left_char in brackets && right_char in brackets.values) {
                 Gtk.TextIter start, end;
 
-                current_buffer.get_selection_bounds (out start, out end);
+                buffer.get_selection_bounds (out start, out end);
                 start.backward_char ();
                 end.forward_char ();
-                current_buffer.select_range (start, end);
+                buffer.select_range (start, end);
             }
         }
     }
 
     void complete_brackets (string opening_bracket) {
         Gtk.TextIter start, end;
-        current_buffer.get_selection_bounds (out start, out end);
+        buffer.get_selection_bounds (out start, out end);
 
-        string current_selection = current_buffer.get_text (start, end, true);
+        string current_selection = buffer.get_text (start, end, true);
         string closing_bracket = brackets[opening_bracket];
         string text = opening_bracket + current_selection + closing_bracket;
 
-        current_buffer.begin_user_action();
+        buffer.begin_user_action();
 
-        current_buffer.delete (ref start, ref end);
-        current_buffer.insert (ref start, text, text.length);
+        buffer.delete (ref start, ref end);
+        buffer.insert (ref start, text, text.length);
 
-        current_buffer.get_selection_bounds (out start, out end);
+        buffer.get_selection_bounds (out start, out end);
         end.backward_char ();
         start.backward_chars (current_selection.length + 1);
-        current_buffer.select_range (start, end);
+        buffer.select_range (start, end);
 
-        current_buffer.end_user_action();
+        buffer.end_user_action();
     }
 
     void skip_char () {
         Gtk.TextIter start, end;
 
-        current_buffer.get_selection_bounds (out start, out end);
+        buffer.get_selection_bounds (out start, out end);
         end.forward_char ();
-        current_buffer.place_cursor (end);
+        buffer.place_cursor (end);
     }
 
     bool has_valid_next_char (string next_char) {
@@ -149,8 +146,7 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
             string next_char = get_next_char ();
 
             if (bracket in brackets &&
-                (current_buffer.has_selection ||
-                has_valid_next_char (next_char))) {
+                (buffer.has_selection || has_valid_next_char (next_char))) {
                 complete_brackets (bracket);
                 return true;
             } else if (bracket in brackets.values && next_char == bracket) {
--- a/plugins/brackets-completion/brackets-completion.vala
+++ b/plugins/brackets-completion/brackets-completion.vala
@@ -28,7 +28,8 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
         "", " ", "\b", "\r", "\n", "\t", ",", ".", ";", ":"
     };
 
-    Gtk.TextBuffer buffer;
+    Gtk.TextBuffer current_buffer;
+    Scratch.Widgets.SourceView current_source_view;
 
     Scratch.Services.Interface plugins;
     public Object object { owned get; construct; }
@@ -64,73 +65,80 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
     }
 
     void on_hook_document (Scratch.Services.Document doc) {
-        buffer = doc.source_view.buffer;
+        current_buffer = doc.source_view.buffer;
 
-        doc.source_view.backspace.connect (on_backspace);
-        doc.source_view.key_press_event.connect (on_key_press);
+        if (current_source_view != null) {
+            current_source_view.key_press_event.disconnect (on_key_press);
+            current_source_view.backspace.disconnect (on_backspace);
+        }
+
+        current_source_view = doc.source_view;
+
+        current_source_view.key_press_event.connect (on_key_press);
+        current_source_view.backspace.connect (on_backspace);
     }
 
     string get_next_char () {
         Gtk.TextIter start, end;
 
-        buffer.get_selection_bounds (out start, out end);
+        current_buffer.get_selection_bounds (out start, out end);
         end.forward_char ();
 
-        return buffer.get_text (start, end, true) ;
+        return current_buffer.get_text (start, end, true) ;
     }
 
     string get_previous_char () {
         Gtk.TextIter start, end;
 
-        buffer.get_selection_bounds (out start, out end);
+        current_buffer.get_selection_bounds (out start, out end);
         start.backward_char ();
 
-        return buffer.get_text (start, end, true);
+        return current_buffer.get_text (start, end, true);
     }
 
     void on_backspace () {
-        if (!buffer.has_selection) {
+        if (!current_buffer.has_selection) {
             string left_char = get_previous_char ();
             string right_char = get_next_char ();
 
             if (left_char in brackets && right_char in brackets.values) {
                 Gtk.TextIter start, end;
 
-                buffer.get_selection_bounds (out start, out end);
+                current_buffer.get_selection_bounds (out start, out end);
                 start.backward_char ();
                 end.forward_char ();
-                buffer.select_range (start, end);
+                current_buffer.select_range (start, end);
             }
         }
     }
 
     void complete_brackets (string opening_bracket) {
         Gtk.TextIter start, end;
-        buffer.get_selection_bounds (out start, out end);
+        current_buffer.get_selection_bounds (out start, out end);
 
-        string current_selection = buffer.get_text (start, end, true);
+        string current_selection = current_buffer.get_text (start, end, true);
         string closing_bracket = brackets[opening_bracket];
         string text = opening_bracket + current_selection + closing_bracket;
 
-        buffer.begin_user_action();
+        current_buffer.begin_user_action();
 
-        buffer.delete (ref start, ref end);
-        buffer.insert (ref start, text, text.length);
+        current_buffer.delete (ref start, ref end);
+        current_buffer.insert (ref start, text, text.length);
 
-        buffer.get_selection_bounds (out start, out end);
+        current_buffer.get_selection_bounds (out start, out end);
         end.backward_char ();
         start.backward_chars (current_selection.length + 1);
-        buffer.select_range (start, end);
+        current_buffer.select_range (start, end);
 
-        buffer.end_user_action();
+        current_buffer.end_user_action();
     }
 
     void skip_char () {
         Gtk.TextIter start, end;
 
-        buffer.get_selection_bounds (out start, out end);
+        current_buffer.get_selection_bounds (out start, out end);
         end.forward_char ();
-        buffer.place_cursor (end);
+        current_buffer.place_cursor (end);
     }
 
     bool has_valid_next_char (string next_char) {
@@ -148,7 +156,7 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
             string next_char = get_next_char ();
 
             if (bracket in brackets &&
-                (buffer.has_selection || has_valid_next_char (next_char))) {
+                (current_buffer.has_selection || has_valid_next_char (next_char))) {
                 complete_brackets (bracket);
                 return true;
             } else if (bracket in brackets.values && next_char == bracket) {
--- a/plugins/brackets-completion/brackets-completion.vala
+++ b/plugins/brackets-completion/brackets-completion.vala
@@ -1,104 +1,104 @@
 // -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
 /***
   BEGIN LICENSE
-	
+
   Copyright (C) 2013 Mario Guerriero <mario@elementaryos.org>
-  This program is free software: you can redistribute it and/or modify it	
-  under the terms of the GNU Lesser General Public License version 3, as published	
+  This program is free software: you can redistribute it and/or modify it
+  under the terms of the GNU Lesser General Public License version 3, as published
   by the Free Software Foundation.
-	
-  This program is distributed in the hope that it will be useful, but	
-  WITHOUT ANY WARRANTY; without even the implied warranties of	
-  MERCHANTABILITY, SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR	
+
+  This program is distributed in the hope that it will be useful, but
+  WITHOUT ANY WARRANTY; without even the implied warranties of
+  MERCHANTABILITY, SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR
   PURPOSE.  See the GNU General Public License for more details.
-	
-  You should have received a copy of the GNU General Public License along	
-  with this program.  If not, see <http://www.gnu.org/licenses/>	
-  
-  END LICENSE	
+
+  You should have received a copy of the GNU General Public License along
+  with this program.  If not, see <http://www.gnu.org/licenses/>
+
+  END LICENSE
 ***/
 
 public const string NAME = _("Brackets Completion");
 public const string DESCRIPTION = _("Complete brackets while typing");
 
 public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Activatable {
-    Gee.HashMap<string, string> brackets;
 
     Gee.TreeSet<Gtk.TextBuffer> buffers;
     Gtk.TextBuffer current_buffer;
-    string last_inserted;
+    MainWindow window;
+
+    Gee.HashMap<uint, int> keys;
+    string[] opening_brackets;
+    string[] closing_brackets;
 
     Scratch.Services.Interface plugins;
     public Object object { owned get; construct; }
 
     public void update_state () {
-        
+
     }
 
     public void activate () {
-        this.buffers = new Gee.TreeSet<Gtk.TextBuffer> ();
-        this.brackets = new Gee.HashMap<string, string> ();
-        this.brackets.set ("(", ")");
-        this.brackets.set ("[", "]");
-        this.brackets.set ("{", "}");
-        this.brackets.set ("<", ">");
-        this.brackets.set ("⟨", "⟩");
-        this.brackets.set ("｢", "｣");
-        this.brackets.set ("⸤", "⸥");
-        this.brackets.set ("‘", "‘");
-        this.brackets.set ("'", "'");
-        this.brackets.set ("\"", "\"");
+        buffers = new Gee.TreeSet<Gtk.TextBuffer> ();
+        opening_brackets = {"{", "[", "(", "'", "\"", "`"};
+        closing_brackets = {"}", "]", ")", "'", "\"", "`"};
+
+        keys = new Gee.HashMap<uint, int> ();
+        keys.set (Gdk.Key.braceleft, 0);
+        keys.set (Gdk.Key.bracketleft, 1);
+        keys.set (Gdk.Key.parenleft, 2);
+        keys.set (Gdk.Key.quoteright, 3);
+        keys.set (Gdk.Key.quotedbl, 4);
+        keys.set (Gdk.Key.grave, 5);
 
         plugins = (Scratch.Services.Interface) object;
+
+        plugins.hook_window.connect ((w) => {
+            window = w;
+            window.key_press_event.connect (on_key_press);
+        });
+
         plugins.hook_document.connect ((doc) => {
             var buf = doc.source_view.buffer;
-            buf.insert_text.disconnect (on_insert_text);
-            buf.insert_text.connect (on_insert_text);
-            this.buffers.add (buf);
-            this.current_buffer = buf;
+            buffers.add (buf);
+            current_buffer = buf;
         });
     }
 
     public void deactivate () {
-        foreach (var buf in buffers) {
-            buf.insert_text.disconnect (on_insert_text);
-        }
+        window.key_press_event.disconnect (on_key_press);
     }
 
-    void on_insert_text (ref Gtk.TextIter pos, string new_text, int new_text_length) {
-        // If you are copy/pasting a large amount of text...
-        if (new_text_length > 1) {
-            return;
-        }
+    bool on_key_press (Gdk.EventKey event) {
+        var doc = window.get_current_document ();
+        if (doc != null && doc.source_view.has_focus) {
+            GLib.print(Gdk.keyval_name (event.keyval));
+            if (event.keyval in keys) {
 
-        // To avoid infinite loop
-        if (this.last_inserted == new_text) {
-            return;
-        }
+                var buf = this.current_buffer;
+                Gtk.TextIter start, end;
 
-        if (new_text in this.brackets.keys) {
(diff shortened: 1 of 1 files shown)


plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,110 +1,109 @@
...
-        }
+        window.key_press_event.disconnect (on_key_press);
     }
 
-    void on_insert_text (ref Gtk.TextIter pos, string new_text, int new_text_length) {
-        // If you are copy/pasting a large amount of text...
-        if (new_text_length > 1) {
-            return;
-        }
+    bool on_key_press (Gdk.EventKey event) {
```
jeremypw: Deletion of "on_insert_text" conflicts with PR#193.  These two PRs need to be resolved/combined in some way.

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,110 +1,109 @@
...
-        this.brackets.set ("⟨", "⟩");
-        this.brackets.set ("｢", "｣");
-        this.brackets.set ("⸤", "⸥");
-        this.brackets.set ("‘", "‘");
-        this.brackets.set ("'", "'");
-        this.brackets.set ("\"", "\"");
+        buffers = new Gee.TreeSet<Gtk.TextBuffer> ();
+        opening_brackets = {"{", "[", "(", "'", "\"", "`"};
+        closing_brackets = {"}", "]", ")", "'", "\"", "`"};
+
```
jeremypw: Not all types of brackets catered for, e.g. <> - why is this?

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,110 +1,109 @@
...
-            this.last_inserted = text;
-            buf.insert (ref pos, text, len);
+                var index = keys.get (event.keyval);
+                var open_bracket = opening_brackets[index];
+                var close_bracket = closing_brackets[index];
+                var text = open_bracket + current_text + close_bracket;
 
-            //To make " and ' brackets work correctly (opening and closing chars are the same)
-            this.last_inserted = null;
+                buf.delete_selection (false, false);
```
jeremypw: This deletion gets recorded as a separate operation so undoing with Ctrl-Z has unexpected effect of deleting and then reinserting the selection with the next Ctrl-Z. Please try to avoid this.

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,110 +1,109 @@
...
 
-    void on_insert_text (ref Gtk.TextIter pos, string new_text, int new_text_length) {
-        // If you are copy/pasting a large amount of text...
-        if (new_text_length > 1) {
-            return;
-        }
+    bool on_key_press (Gdk.EventKey event) {
+        var doc = window.get_current_document ();
+        if (doc != null && doc.source_view.has_focus) {
+            if (event.keyval in keys) {
```
jeremypw: This if clause can be combined with the previous one.

plugins/brackets-completion/brackets-completion.vala
OttoRobba: <> completion seems largely to be language dependent - ideally we want to have it when dealing with types in vala or tags in html but for many languages it interferes normal usage (for things like <=, >=, et al...).

Other bracket completion packages (as per Atom and Gedit) seem to forego completing <> due to this.

plugins/brackets-completion/brackets-completion.vala
OttoRobba: I would love that - I really hate how it results in two undo actions. Is there a way to delete the selection without recording the operation?

plugins/brackets-completion/brackets-completion.vala
OttoRobba: Will do when I get home.

plugins/brackets-completion/brackets-completion.vala
OttoRobba: I was thinking of solving the issue addressed on PR#193 differently:

- Attach a function to backspace
- Is the character being deleted in opening_brackets?
- If true, check if the character to the right of the cursor is the matching twin in closing_brackets.
- If true, delete it as well.
- Else, do nothing.

Not sure it would work 100% though, so I'm willing to adapt this PR. :-)

snowparrot:
To answer some questions:

- My pull request works like:
1. If user types bracket, the bool attention_bracket becomes true.
2. If user does anything after like moving cursor, insert text except hitting backspace,
then attention_bracket becomes false.
3. If user hits backspace and attention_bracket is true, then
delete the right letter true.

So my code doesnt rely on insert_text but on setting attention_bracket false or true.

plugins/brackets-completion/brackets-completion.vala
snowparrot: I wanted to ask that question in an issue, because some web stuff is done by scratch. Maybe it depends on the input of document.

snowparrot:
I'm thinking to start rewriting the plugin new because in the close future I wanted to improve the code:
- If someone types a bracket, there should be no completion in front of a letter. (At least sublime does it like this)

If I see the plugin adding features isnt very easy and it needs an rewriting. What do you think?

matzipan:
@OttoRobba You can [join](https://elementarycommunity.herokuapp.com/) the elementary community slack if you wanna discuss this at greater length. Given the overlap with @snowparrot's branch, I think you'd find it more productive to collaborate.

plugins/brackets-completion/brackets-completion.vala
jeremypw: Is there a way of not deleting the selection and then inserting it again with brackets? e.g. just insert the brackets? I have not had time to look in detail I am afraid.

OttoRobba: Not as a single operation I'm afraid. I can try to find a different way during the weekend.

jeremypw: Its not so bad if the operations are the insertion of brackets. Undoing -> removing one bracket at a time is less "unexpected" then deleting and then replacing the selection.

plugins/brackets-completion/brackets-completion.vala
jeremypw: Good point, although this change is unrelated to the purpose of the PR and probably needs UX team input.  It is better to keep the PR focussed on one aspect at a time.

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,110 +1,105 @@
...
 public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Activatable {
-    Gee.HashMap<string, string> brackets;
 
     Gee.TreeSet<Gtk.TextBuffer> buffers;
     Gtk.TextBuffer current_buffer;
-    string last_inserted;
+    MainWindow window;
+
+    Gee.HashMap<uint, int> keys;
+    string[] opening_brackets;
```
donadigo: This string array could as well be a constant here. You don't change it anywhere in the code.

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,110 +1,105 @@
...
+    string[] closing_brackets;
```
donadigo: The same here.

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,110 +1,105 @@
...
-        this.brackets.set ("⸤", "⸥");
-        this.brackets.set ("‘", "‘");
-        this.brackets.set ("'", "'");
-        this.brackets.set ("\"", "\"");
+        buffers = new Gee.TreeSet<Gtk.TextBuffer> ();
+        opening_brackets = {"{", "[", "(", "'", "\"", "`"};
+        closing_brackets = {"}", "]", ")", "'", "\"", "`"};
+
+        keys = new Gee.HashMap<uint, int> ();
+        keys.set (Gdk.Key.braceleft, 0);
```
donadigo: You can use a native Vala syntax for setting those pairs like this:
`keys[Gdk.Key.braceleft] = 0;`

I would like to see this done for all of the `set` calls below but it is not required.

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,110 +1,105 @@
...
-            string text = this.brackets.get (new_text);
-            int len = text.length;
-            this.last_inserted = text;
-            buf.insert (ref pos, text, len);
+            buf.get_selection_bounds (out start, out end);
+            var current_text = buf.get_text (start, end, true);
 
-            //To make " and ' brackets work correctly (opening and closing chars are the same)
-            this.last_inserted = null;
+            var index = keys.get (event.keyval);
```
donadigo: The same here. You can access the index value by doing:
`var index = keys[event.keyval];`

OttoRobba:
Merged bracket deletion from https://github.com/elementary/scratch/pull/193 and refactored so that completion is always a single action (no more weird undo behaviour!).

OttoRobba:
@jeremypw Sounds good, I will modify that when I get home :)

OttoRobba:
Implemented the changes suggested by @donadigo and by @jeremypw (good call by the way, I had no idea I could use `start_user_action` like that! :D).

Code is quite a bit cleaner as well. Let me know if you see anything else that can be improved.

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,104 +1,158 @@
...
-        this.brackets.set ("‘", "‘");
-        this.brackets.set ("'", "'");
-        this.brackets.set ("\"", "\"");
+        brackets = new Gee.HashMap<string, string> ();
+        brackets["("] = ")";
+        brackets["["] = "]";
+        brackets["{"] = "}";
+        brackets["'"] = "'";
+        brackets["\""] = "\"";
+        brackets["`"] = "`";
```
matzipan: Well no you're duplicating `brackets` as well as the `opening/closing_brackets` attributes on the object. Maybe a mistake?

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,104 +1,158 @@
...
+        keys = new Gee.HashMap<uint, string> ();
+        keys[Gdk.Key.braceleft] = "{";
+        keys[Gdk.Key.bracketleft] = "[";
+        keys[Gdk.Key.parenleft] = "(";
+        keys[Gdk.Key.braceright] = "}";
+        keys[Gdk.Key.bracketright] = "]";
+        keys[Gdk.Key.parenright] = ")";
+        keys[Gdk.Key.quoteright] = "'";
+        keys[Gdk.Key.quotedbl] = "\"";
+        keys[Gdk.Key.grave] = "`";
```
matzipan: I'm half certain this information can be obtained from `Gdk.EventKey`. Unsure about the exact method

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,104 +1,158 @@
...
-            return;
+        return current_buffer.get_text (start, end, true);
+    }
+
+    void on_backspace () {
+        if (!current_buffer.has_selection) {
+            string left_char = get_previous_char ();
+            string right_char = get_next_char ();
+
+            if (left_char in opening_brackets && right_char in closing_brackets) {
```
matzipan: Do we need to consider special cases when there is a whitespace between the brackets (`{ }`) or when the closing bracket has not been added by the plugin but has been there through other means?

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,104 +1,158 @@
...
+        Gtk.TextIter start, end;
+
+        current_buffer.get_selection_bounds (out start, out end);
+        end.forward_char ();
+        current_buffer.place_cursor (end);
+    }
+
+    bool on_key_press (Gdk.EventKey event) {
+        if (event.keyval in keys
+        && !(Gdk.ModifierType.MOD1_MASK in event.state)
```
matzipan: Generally you leave the operator (eg `&&`) on the previous line so that it's obvious that there is a continuation and you also add a tab on the next lines to make it visually obvious that it's part of the same block. So
```
if (event.keyval in keys &&
    !(.....) &&
    !(....)) {

File: plugins/brackets-completion/brackets-completion.vala [hunk 13]
```diff
@@ -1,104 +1,158 @@
...
+        && !(Gdk.ModifierType.CONTROL_MASK in event.state)) {
+            string bracket = keys[event.keyval];
+            string next_char = get_next_char ();
+
+            if (bracket in opening_brackets) {
+                complete_brackets (bracket);
+                return true;
```
matzipan: Is there a reason why you want to stop other handlers from being fired?

plugins/brackets-completion/brackets-completion.vala
OttoRobba: I agree, definitely not ideal - I need to be able to check if a bracket is used for opening and/or closing, as well as getting the matching bracket for it. Hmm I think I know a way to improve this. Sorry, still learning Vala :)

plugins/brackets-completion/brackets-completion.vala
jeremypw: I think was intended as a quick way of defining the set of handled events as well as translating to string. Also looking up the string from the event struct is deprecated because of im issues, as I understand it.

plugins/brackets-completion/brackets-completion.vala
jeremypw: The two arrays should be replaced with brackets.get_keys () and brackets.get_values ().

plugins/brackets-completion/brackets-completion.vala
OttoRobba: I was copying what was done in the terminal plugin, will fix it now.

plugins/brackets-completion/brackets-completion.vala
OttoRobba: Yes. We don't want to further propagate the bracket insertion action else we end up with funky results.
Instead of `[]` we would get `[[]`

plugins/brackets-completion/brackets-completion.vala
OttoRobba: Not really. Going by the behavior of other editors, deleting an opening bracket defaults to also deleting the closing bracket but only if it is the next character. Anything else would be highly impractical, I think.

plugins/brackets-completion/brackets-completion.vala
OttoRobba: Hmm I think you are correct, I'll take a look at it. :)

plugins/brackets-completion/brackets-completion.vala
OttoRobba: Done. :smile:

plugins/brackets-completion/brackets-completion.vala
OttoRobba: After testing and reading the docs, it does seem possible to easily get the utf8 from the EventKey - at most I can get the unicode value but then I would have to find a way to translate it to a string, which seems more work and more obscure than the current implementation.

OttoRobba: Done! Using `event.str`.

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,104 +1,142 @@
...
+
+        current_buffer.get_selection_bounds (out start, out end);
+        end.forward_char ();
+        current_buffer.place_cursor (end);
+    }
+
+    bool on_key_press (Gdk.EventKey event) {
+        if (!(Gdk.ModifierType.MOD1_MASK in event.state) &&
+            !(Gdk.ModifierType.CONTROL_MASK in event.state)) {
+            string bracket = event.str;
```
jeremypw: You should not use event.str - it is deprecated - see https://people.gnome.org/~gcampagna/docs/Gdk-3.0/Gdk.EventKey-string.html

OttoRobba: oh :|
Valadocs really should have deprecation warnings. I will revert that commit later today.

jeremypw: Generally it does for functions, but this appears to have been missed - I got it from from devhelp tbh.

matzipan: Documentation seems to recommend `Gdk.unicode_to_keyval`. Perhaps it's a good approximation for the beginning.

OttoRobba: We can skip that conversion step using `event.keyval`, which was what I was doing (and what will be when I undo the `event.str` commit).

OttoRobba: Done, reverted to older commit.

plugins/brackets-completion/brackets-completion.vala
jeremypw: I think there was a misunderstanding here.  I was intending to indicate that I though the map between Gdk.Key and the strings was an OK solution (of course there may be a more elegant one but it was no big deal for me).

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,104 +1,157 @@
...
 public const string DESCRIPTION = _("Complete brackets while typing");
 
 public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Activatable {
     Gee.HashMap<string, string> brackets;
+    Gee.HashMap<uint, string> keys;
 
-    Gee.TreeSet<Gtk.TextBuffer> buffers;
     Gtk.TextBuffer current_buffer;
-    string last_inserted;
+    Gtk.SourceView current_view;
```
matzipan: If you only have a `current_view`, how does this work with split view? I'm assuming it works with multiple tabs since the the main view is reused.

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,104 +1,157 @@
...
         plugins.hook_document.connect ((doc) => {
-            var buf = doc.source_view.buffer;
-            buf.insert_text.disconnect (on_insert_text);
-            buf.insert_text.connect (on_insert_text);
-            this.buffers.add (buf);
-            this.current_buffer = buf;
+            current_buffer = doc.source_view.buffer;
+            current_view = doc.source_view;
+
+            current_view.key_press_event.disconnect (on_key_press);
```
matzipan: Why is this needed?

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,104 +1,157 @@
...
+            current_view.key_press_event.connect (on_key_press);
+            current_view.backspace.disconnect (on_backspace);
```
matzipan: Same here?

plugins/brackets-completion/brackets-completion.vala
```diff
@@ -1,104 +1,157 @@
...
+
+            string bracket = keys[event.keyval];
+            string next_char = get_next_char ();
+
+            if (bracket in brackets) {
+                complete_brackets (bracket);
+                return true;
+            } else if (bracket in brackets.values && next_char == bracket) {
+                skip_char ();
+                return true;
```
matzipan: You said:

> Yes. We don't want to further propagate the bracket insertion action else we end up with funky results.
Instead of [] we would get [[]

Well this is only a limitation of your implementation. I guess it's fine leaving it as it is, since we don't have any plugins that might want to intervene in this case, but in my opinion it's bad behaviour.

plugins/brackets-completion/brackets-completion.vala
OttoRobba: Mostly something that is done in other plugins as well, I thought it might be a valid standard? Likely to be used in case the plugin has not been properly turned off. I don't see harm in removing it though, was just trying to keep it reasonably similar to other plugins.

plugins/brackets-completion/brackets-completion.vala
OttoRobba: Same as above.

plugins/brackets-completion/brackets-completion.vala
OttoRobba: Yep. But that is because using the on_insert event requires us to do a lot of book-keeping, which is why I chose to take this path. I simply would rather not have so much state in a simple plugin.

plugins/brackets-completion/brackets-completion.vala
OttoRobba: It works fine, I think it is because Scratch plugins uses the SourceView widget.

Would be nice if someone could confirm that though, I'm just guessing.

matzipan:
Ok, I'm doing some more extensive checking because I'm not sure about the `current_view` stuff. In the meanwhile I stumbled upon this fatal error: `[FATAL 17:49:59.640565] enchant_dict_check: assertion 'len' failed`.

Its backtrace is: 
```
#0  0x00007ffff5562a5b in g_logv () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff5562bcf in g_log () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007fffe0e52e06 in enchant_dict_check () from /usr/lib/x86_64-linux-gnu/libenchant.so.1
#3  0x00007fffe105ef83 in ?? () from /usr/lib/x86_64-linux-gnu/libgtkspell3-3.so.0
#4  0x00007fffe105fae6 in ?? () from /usr/lib/x86_64-linux-gnu/libgtkspell3-3.so.0
#5  0x00007fffe105ff76 in ?? () from /usr/lib/x86_64-linux-gnu/libgtkspell3-3.so.0
#6  0x00007ffff5832fa5 in g_closure_invoke () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#7  0x00007ffff5845264 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#8  0x00007ffff584dd5c in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#9  0x00007ffff584e08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#10 0x00007fffe0a365d3 in scratch_plugins_brackets_completion_complete_brackets () from /usr/lib/x86_64-linux-gnu/io.elementary.code/plugins/brackets-completion/libbrackets-completion.so
#11 0x00007fffe0a36e13 in scratch_plugins_brackets_completion_on_key_press () from /usr/lib/x86_64-linux-gnu/io.elementary.code/plugins/brackets-completion/libbrackets-completion.so
#12 0x00007fffe0a34c32 in _scratch_plugins_brackets_completion_on_key_press_gtk_widget_key_press_event ()
   from /usr/lib/x86_64-linux-gnu/io.elementary.code/plugins/brackets-completion/libbrackets-completion.so
#13 0x00007ffff6722fac in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#14 0x00007ffff5832fa5 in g_closure_invoke () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#15 0x00007ffff5844fc1 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#16 0x00007ffff584d7f9 in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#17 0x00007ffff584e08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#18 0x00007ffff6860c3c in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#19 0x00007ffff6880cbb in gtk_window_propagate_key_event () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#20 0x00007ffff688413b in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#21 0x00007ffff6722fac in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#22 0x00007ffff5832fa5 in g_closure_invoke () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#23 0x00007ffff584556e in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#24 0x00007ffff584d7f9 in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#25 0x00007ffff584e08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#26 0x00007ffff6860c3c in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#27 0x00007ffff67204c9 in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#28 0x00007ffff67221bc in gtk_main_do_event () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#29 0x00007ffff628fd92 in ?? () from /usr/lib/x86_64-linux-gnu/libgdk-3.so.0
#30 0x00007ffff555c197 in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#31 0x00007ffff555c3f0 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#32 0x00007ffff555c49c in g_main_context_iteration () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#33 0x00007ffff6eecd6d in g_application_run () from /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#34 0x00007ffff74b0450 in granite_application_run () from /usr/lib/x86_64-linux-gnu/libgranite.so.3
#35 0x000000000041f1eb in scratch_application_main ()
#36 0x000000000041f234 in main ()
```

jeremypw:
Do you recall how you produced this crash?

matzipan:
If I activate and then deactivate the plugin, I get the following `[WARNING 17:58:38.232120] [GLib-GObject] invalid (NULL) pointer instance`

With the following backtrace:
```
#0  0x00007ffff5562a5b in g_logv () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff5562bcf in g_log () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007ffff5857e2a in g_type_check_instance () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#3  0x00007ffff584c514 in g_signal_handlers_disconnect_matched () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#4  0x00007fffcf9763d2 in scratch_plugins_brackets_completion_real_deactivate () from /usr/lib/x86_64-linux-gnu/io.elementary.code/plugins/brackets-completion/libbrackets-completion.so
#5  0x00000000004444e6 in __lambda41_ ()
#6  0x0000000000444542 in ___lambda41__peas_extension_set_extension_removed ()
#7  0x00007ffff5832fa5 in g_closure_invoke () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#8  0x00007ffff5844fc1 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#9  0x00007ffff584dd5c in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#10 0x00007ffff584e08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#11 0x00007ffff79beaaa in ?? () from /usr/lib/x86_64-linux-gnu/libpeas-1.0.so.0
#12 0x00007ffff79beb18 in ?? () from /usr/lib/x86_64-linux-gnu/libpeas-1.0.so.0
#13 0x00007ffff5832fa5 in g_closure_invoke () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#14 0x00007ffff5844fc1 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#15 0x00007ffff584dd5c in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#16 0x00007ffff584e08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#17 0x00007ffff79bd16b in peas_engine_unload_plugin () from /usr/lib/x86_64-linux-gnu/libpeas-1.0.so.0
#18 0x00007ffff7bd1ce4 in ?? () from /usr/lib/x86_64-linux-gnu/libpeas-gtk-1.0.so.0
#19 0x00007ffff5835a5a in g_cclosure_marshal_VOID__STRINGv () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#20 0x00007ffff58331d4 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#21 0x00007ffff584d9a6 in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#22 0x00007ffff584e08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#23 0x00007ffff6650191 in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#24 0x00007ffff663d742 in gtk_cell_area_activate_cell () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#25 0x00007ffff663dafe in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#26 0x00007ffff684f089 in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#27 0x00007fffef4e5e40 in ffi_call_unix64 () from /usr/lib/x86_64-linux-gnu/libffi.so.6
#28 0x00007fffef4e58ab in ffi_call () from /usr/lib/x86_64-linux-gnu/libffi.so.6
#29 0x00007ffff5833cf5 in g_cclosure_marshal_generic_va () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#30 0x00007ffff58331d4 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#31 0x00007ffff584d9a6 in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#32 0x00007ffff584e08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#33 0x00007ffff66dcc16 in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#34 0x00007ffff5835dbe in g_cclosure_marshal_VOID__BOXEDv () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#35 0x00007ffff58331d4 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#36 0x00007ffff584d9a6 in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#37 0x00007ffff584e08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#38 0x00007ffff66d9bab in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#39 0x00007ffff66db1ab in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#40 0x00007ffff66ddde5 in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#41 0x00007ffff66adbeb in gtk_event_controller_handle_event () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#42 0x00007ffff685ebfb in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#43 0x00007ffff7bd2dcd in ?? () from /usr/lib/x86_64-linux-gnu/libpeas-gtk-1.0.so.0
#44 0x00007ffff672309a in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#45 0x00007ffff58331d4 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#46 0x00007ffff584d4b8 in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#47 0x00007ffff584e08f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#48 0x00007ffff6860c3c in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#49 0x00007ffff67203be in ?? () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#50 0x00007ffff67221bc in gtk_main_do_event () from /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#51 0x00007ffff628fd92 in ?? () from /usr/lib/x86_64-linux-gnu/libgdk-3.so.0
```

matzipan:
@jeremypw  the first issue seems to only pop up when your bracket is the first thing on the line.

jeremypw:
@matzipan:  It appears to be to do with spellchecking - does it happen with the spellcheck plugin disabled?
So far I have only tested this as the only plugin.

matzipan:
You are correct about the first issue.

matzipan:
Ok so I can't see any issues apart from [this one right now](https://github.com/elementary/scratch/pull/203#issuecomment-329977187)

OttoRobba:
@matzipan Nice catch, had not noticed that. It is quite weird. If I remove the calls out of the deactivation method, the error no longer happens and things still work. If I deactivate the plugin, it stops completion as normal. If I activate it back, works normally as well.

I'm honestly confused by it. I'll take a deeper look to see if I understand what is happening.

OttoRobba:
@matzipan I refactored deactivation, so the fatal errors do not occur and I renamed the variables as the names were wrong. I was not connecting to the current view but rather to the SourceView, which seems to be instanced only once by Scratch and then reused? I'm really unsure. Is there someone who knows? I hate not understanding why things happen :|

matzipan:
> which seems to be instanced only once by Scratch and then reused

Well this is why I was asking if it works when adding another view (press on settings button in top right and then "Add view"). 

> Is there someone who knows? I hate not understanding why things happen :| 

Welp, we're gonna have to look through the source code and find out. I'll do this tomorrow.

matzipan:
The `hook_document` seems to be invoked every time a view gets focus. So I think we're alright.

OttoRobba:
@matzipan Would that be the reason why other plugins first try to disconnect any hooks? Should we adhere to that?

matzipan:
I think that is expected behavior. I will have a second look though.

matzipan:
I think you are right that this might cause problems. So for example right now, everytime on_hook_document is called, it binds a new key handler - but they never get unbinded! I don't think there is any check against binding the same thing multiple times inside vala/glib. So that's why the disconnect is there.

Can somebody else confirm this?

matzipan:
Ok after a bit of investigation I actually think the code should look like this:
```
if (current_view != null) {  
    current_view.key_press_event.disconnect (on_key_press);
    current_view.backspace.disconnect (on_backspace);
}

current_view = doc.source_view;

current_view.key_press_event.connect (on_key_press);
current_view.backspace.connect (on_backspace);
```

This way it unbinds from the previous view (whether it's the same or different to the new one), and it binds to the new one.

matzipan:
Similarly, the deactivate should be 
```
if (current_view != null) {  
    current_view.key_press_event.disconnect (on_key_press);
    current_view.backspace.disconnect (on_backspace);
}
```

matzipan:
Ok, so after  a short code investigation it turns out I was right. With the current code, adding `message("BLALBLALALL");` at the start of the `on_backspace` method, when I press backspace, with around 5 tabs open I get the following print:
```
[INFO 18:14:41.100591] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100642] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100672] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100699] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100726] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100752] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100777] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100803] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100828] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100853] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100878] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100903] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100927] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100952] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.100976] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101002] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101027] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101051] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101076] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101101] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101126] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101150] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101175] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101199] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101224] brackets-completion.vala:93: BLALBLALALL
[INFO 18:14:41.101249] brackets-completion.vala:93: BLALBLALALL
```

By adding the following patch:
```
diff --git a/plugins/brackets-completion/brackets-completion.vala b/plugins/brackets-completion/brackets-completion.vala
index f25e3a9..1b9080c 100644
--- a/plugins/brackets-completion/brackets-completion.vala
+++ b/plugins/brackets-completion/brackets-completion.vala
@@ -29,6 +29,7 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
     };
 
     Gtk.TextBuffer buffer;
+    Scratch.Widgets.SourceView current_source_view;
 
     Scratch.Services.Interface plugins;
     public Object object { owned get; construct; }
@@ -66,8 +67,15 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
     void on_hook_document (Scratch.Services.Document doc) {
         buffer = doc.source_view.buffer;
 
-        doc.source_view.backspace.connect (on_backspace);
-        doc.source_view.key_press_event.connect (on_key_press);
+        if (current_source_view != null) {
+            current_source_view.key_press_event.disconnect (on_key_press);
+            current_source_view.backspace.disconnect (on_backspace);
+        }
+
+        current_source_view = doc.source_view;
+
+        current_source_view.key_press_event.connect (on_key_press);
+        current_source_view.backspace.connect (on_backspace);
     }
 
     string get_next_char () {
```

Pressing backspace only one line per press:
```
[INFO 18:21:08.897775] brackets-completion.vala:100: BLALBLALALL
```

matzipan:
I've pushed a commit with this patch.

PR #209, opened 2017-09-03
No restore failed doc

### LINKED ISSUE #5: Scratch hanging on large files
When opening large files (in my case a 35MB ".run" file) scratch will hang on startup until the session is deleted by hand. Scratch should warn the user if a file exceeds a certain size or load the file asycrounously.

Launchpad Details: [#LP1368726](https://bugs.launchpad.net/bugs/1368726) Alexander Schittler - 2014-09-12 12:09:53 +0000

### LINKED ISSUE #81: When failing to open a big file (1 GB) Scratch gets stucked forever
The point is that as Scratch remembers its status, it will keep on trying loading such file. It could be interesting to remember whether it got closed unexpectedly to let the user select whether he wants to reload the previous session to avoid this issue.

jeremypw:
Fixes #81.
Fixes #5 

This branch delays various operations until a document has successfully loaded asynchronously rather than assuming success. This includes adding the document to the list to be saved and restored, thereby fixing the linked issue.

Files are checked to be of the supertype "text/plain" before opening to avoid problems opening inappropriate files.

Speed of loading of large files is increased by first loading into a separate buffer to avoid continuous reformatting of the view.

The load error dialog is minimally fixed in passing (but needs redesign).

TO TEST:

Attempt to load a binary or other non-text file under various conditions.  The app should not hang nor try to restore the file.


src/MainWindow.vala
```diff
@@ -143,6 +143,12 @@ namespace Scratch {
 
             Unix.signal_add (Posix.SIGINT, quit_source_func, Priority.HIGH);
             Unix.signal_add (Posix.SIGTERM, quit_source_func, Priority.HIGH);
+
+            /* Showing welcome sets widgets insensitive.
+             * Welcome view will be removed and the widgets set sensitive
+             * if a document is successfully loaded.
+             */
+            split_view.show_welcome ();
```
matzipan: In the original code, show_welcome was being called by init_actions, which is somewhere before the current invocation, on line 111.  Does the late invocation of this not affect this in any way or introduce any race conditions? 

Furthermore, given your comments about the widgets being set to insensitive, should this not be made more clear somewhere, maybe in the name of the function?

src/Services/Document.vala
```diff
@@ -556,12 +582,17 @@ namespace Scratch.Services {
...
+            /* NOTE: This dialog needs changing to Elementary HIG style */
+            var dialog = new Gtk.MessageDialog  (parent_window,
+                                                 Gtk.DialogFlags.MODAL,
                                                  Gtk.MessageType.ERROR,
-                                                 Gtk.ButtonsType.CLOSE,
-                                                 message);
+                                                 Gtk.ButtonsType.CLOSE, null);
+
+            /* Setting markup now works */
+            dialog.set_markup ( _("File \"%s\" cannot be read. Maybe it is corrupt\nor you do not have the necessary permissions to read it.").printf ("<b>%s</b>".printf (get_basename ())));
```
matzipan: Which way is better phrasing? "File cannot" or "File could not"? 

Would it be in the scope of this branch to distinguish between corruption/slow loading/unsupported data type/permission failure?

src/Widgets/DocumentView.vala
```diff
@@ -156,9 +156,21 @@ public class Scratch.Widgets.DocumentView : Gtk.Box {
         }
 
         doc.create_page ();
-        notebook.insert_tab (doc, -1);
+        doc.working = true;
         current_document = doc;
-        doc.focus ();
+
+        Idle.add (() => { // This ensures new tab is drawn before opening document.
```
matzipan: Should it show a loading spinner or anything?

src/MainWindow.vala
jeremypw: Since restore_opened_documents () is called after the window has been constructed and presented, the late timing of show_welcome () within window construction is not expected to cause problems but let me know if you see anything.

Renaming of existing function which has not been changed is not really within scope of this branch.

src/Services/Document.vala
jeremypw: It is questionable whether I should have touched this dialog at all within this branch! I made minimum changes for it to not to display <b> and </b> to the user which was happening in trunk.

A discussion can certainly be had as to what a suitable message is and whether the exact reason needs to be determined/displayed, but it is outside the scope of this PR.

src/Widgets/DocumentView.vala
jeremypw: The does (or should) display a spinner in the tab since the document's "working" property is set to <b>true</b>, but the loading process seems to block the process that causes is to spin. I am not sure what can be done about that.

jeremypw:
Working on bug when first document opened from welcome screen fails.

src/Widgets/DocumentView.vala
Philip-Scott: Multiple threads :D

src/Services/Document.vala
matzipan: Ah, I see you did not change the message - for some reason I thought this was an entirely new addition (I was using split diff view). 

Well if the intent was to not show the `<b>` remove them from the `<b>%s</b>` string too.

src/Widgets/DocumentView.vala
matzipan: Ah true, there's a loading spinner already!

src/Services/Document.vala
jeremypw: The <b> and </b> no longer show but have the intended effect (bold font).

jeremypw:
Latest commits should fix observed regression as well as improving loading of very large but valid text files.

jeremypw:
OK, thanks for finding that. I'll fix it.

jeremypw:
Latest revisions simplify things by checking for content type before attempting to load rather than trying to detect load failure by rather hackish methods. Tests so far indicate that the method used passes valid files and blocks invalid ones but counter examples should be sought. It should still be possible to close the tab if loading takes too long for any reason.

In addition, it is no longer possible to edit a document until it is fully loaded and formatted (formatting can take considerably longer than loading for very large files).

matzipan:
Ok, it worked with whatever I threw at it. Would it be possible to not do show_all () on the source view until the file has actually loaded?

src/Services/Document.vala
```diff
@@ -160,6 +160,29 @@ namespace Scratch.Services {
                 load_cancellable.cancel ();
             }
 
+            string content_type = ContentType.from_mime_type (get_mime_type ());
+
+            if (!(ContentType.is_a (content_type, "text/plain"))) {
+                // Create a GtkDialog in app style. TODO: Style for elementary
+                // TODO: Give option to continue and try to load?
+                var dialog = new Gtk.MessageDialog ((Gtk.Window?)source_view.get_toplevel (),
```
matzipan: I believe you're looking for `Granite.Widgets.MessageDialog`.

src/Services/Document.vala
```diff
@@ -160,6 +160,29 @@ namespace Scratch.Services {
                 load_cancellable.cancel ();
             }
 
+            string content_type = ContentType.from_mime_type (get_mime_type ());
+
+            if (!(ContentType.is_a (content_type, "text/plain"))) {
```
matzipan: Perhaps we should do this in a more centralized fashion? Have a static method on a file class which checks if a given content type is supported? I'm sayingthis because it might be also useful in the file viewer.

src/Services/Document.vala
```diff
@@ -160,6 +160,29 @@ namespace Scratch.Services {
...
+                var dialog = new Gtk.MessageDialog ((Gtk.Window?)source_view.get_toplevel (),
+                                                    Gtk.DialogFlags.MODAL,
+                                                    Gtk.MessageType.WARNING,
+                                                    Gtk.ButtonsType.CANCEL,
+                                                    "");
+
+                dialog.deletable = false;
+                dialog.use_markup = true;
+
+                dialog.text = ("<b>" + _("%s is not a text document.") +
```
matzipan: Does "text file" explain it better? A text document is more like a letter in Word

src/Services/Document.vala
```diff
@@ -160,6 +160,29 @@ namespace Scratch.Services {
...
+                               "</b>").printf (this.get_basename ());
+                dialog.format_secondary_markup (_("Cannot load this type of file"));
```
matzipan: Code may not be able to load this type of file

src/Services/Document.vala
jeremypw: Will change

src/Services/Document.vala
jeremypw: I think "may" should not be used unless we offer a chance to go ahead anyway.  As, currently, the load fails something like "Cannot load" is more appropriate.

src/Services/Document.vala
jeremypw: Thanks - my granite.vapi was out of date. Will use and bump the minimum granite version.

danrabbit:
If we use granite messagedialog that will mean we're blocked on release until a new version of Granite is out. I'd prefer not to merge things that rely on unreleased versions of libraries

matzipan:
I guess this might be why the build is failing.

Shall we revert back to Gtk, and when the Granite dialog lands we can make a concentrated effort to change them all?

src/Services/Document.vala
matzipan: What do you think about this one?

matzipan:
[This](https://github.com/elementary/scratch/pull/209/commits/b89bbda1d8677167291f11ca7f3e85cb2dc73e10) doesn't seem to work for me.

jeremypw:
Strange - it works for me when loading or restoring a large valid document, for example.  Until the document is fully loaded and editable, the tab is blank (grey).  Could you describe an exact scenario and what you see?

jeremypw:
@matzipan: The side-effect of your last commit is that when loading a very large text file, the text appears and looks like it is editable even though it is still be formated. Previously the document remained hidden until it was fully loaded and editable.  It is probably not a blocking issue though.

matzipan:
I guess you do have a point since this branch initially tried to address large files being loaded. Perhaps we should merge the binary stuff now and continue work on the large text file stuff, perhaps we can find another way, some sort of progressive loading?

matzipan:
I think I have seen some of this progressive loading on Ubuntu 16.04 gedit, but I can't see the same thing in the elementary gedit - are we shipping a different thing?

danrabbit:
We don't patch gedit in any way or have anything to do with it. It comes straight from the Ubuntu repos

jeremypw:
@matzipan:  What was your objection to hiding the document until in an editable state?  Tbh I am not sure how big a problem it is either way i.e. if Code is aimed primarily at coding, most code files are relatively small.

matzipan:
> What was your objection to hiding the document until in an editable state? 

The way you were doing visibility toggling, it introduced a nasty flicker when changing tabs, and also when you loaded a binary file, the sourceview would actually be visible. 

I guess technically it fixes #5 and #81 since it doesn't hang up anymore. However, perhaps we should open a new issue regarding progressive loading of a file.

PR #217, opened 2017-09-15
Duplicate document properly.

### LINKED ISSUE #216: Duplicate document does not copy text
Using the "duplicate document" option on the tab context menu results in a blank new document which is not the expected result.

jeremypw:
Fixes #216 

* Duplicate text from original
* Save if autosave.

### DIFF
--- a/src/Services/Document.vala
+++ b/src/Services/Document.vala
@@ -282,8 +282,8 @@ namespace Scratch.Services {
             return ret_value;
         }
 
-        public async bool save () {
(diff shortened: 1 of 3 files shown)


jeremypw:
I was concerned that having two tabs potentially modifying the same file might cause problems (moreover, the existing behaviour was to create a new document).  Need to refer to UX team  I guess.

matzipan:
Alternatively, we can remove the option altogether?

jeremypw:
Query left in #ux on Slack

PR #222, opened 2017-09-23
Application.vala: Update copyright header

danrabbit:
* Update copyright header
* Add brackets
* remove an unnecessary comment

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -1,26 +1,26 @@
 // -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
-/***
-  BEGIN LICENSE
-
-  Copyright (C) 2011-2012 Giulio Collura <random.cpp@gmail.com>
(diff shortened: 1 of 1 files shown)


src/Application.vala
```diff
@@ -1,26 +1,26 @@
...
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+* General Public License for more details.
+*
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
```
jeremypw: This version of the copyright/license header does not match that appearing at https://www.gnu.org/licenses/gpl-3.0.en.html.

PR #236, opened 2017-10-16
Cleanup some components, create a WelcomeView

tintou:


danrabbit:
screenshot from 2017-10-16 10 26 28

You broke it :p

PR #240, opened 2017-10-16
Document.vala: replace get_mimetype with mimetype property

danrabbit:
Just the mimetype property changes from #237


danrabbit:
This seems to break installations which have old plugins installed. Pinging @codygarver to verify that packaging will clean up those old plugins

PR #249, opened 2017-10-22
Show one loading spinner per file

tintou:
This makes the whole loading process feeling smoother


tintou:
This makes the whole loading process feeling smoother

tintou:
@jeremypw I get fewer critical warnings in my branch than in trunk but I agree that the other warnings should be fixed in another branch.
I haven't changed the loading behavior, it's probably up to the @elementary/ux team to decide

jeremypw:
OK, well in any case an appropriate message should show depending on the situation.  Some critical errors may be due to races so could very depending on number of tabs, document sizes, cpu speed etc I suppose.  I agree design team input is required.

tintou:
@jeremypw On error there is still the error dialog closing the tab, I haven't changed the actual behavior and don't want to change it in this PR

danrabbit:
This seems worse to me. With this branch as each file is loaded it's flashing and I still have to wait for all of the files to load before my last opened file is selected anyways.

PR #252, opened 2017-10-28
Use OptionContext.parse_strv instead of OptionContext.parse

JMoerman:
Currently OptionContext.parse is used in Application.vala, OptionContext.parse removes parsed strings from args, but doesn't free them.
This is unlikely to cause any real issues, but I wouldn't keep memory leaks around.
You either need to make sure that the references to the parsed strings are not lost, or you can use parse_strv.
See: [OptionContext.parse_strv](https://valadoc.org/glib-2.0/GLib.OptionContext.parse_strv.html)

(Right now an array of owned strings is passed to parse (that you pass an unowned reference _temp_ of this array changes nothing), _if_ you want to use OptionContext.parse pass an array of unowned strings.)


JMoerman:
This same issue, is also present in files (and probably every single other elementary Vala application that overrides `GLib.Application.command_line`, because the documentation of `OptionContext.parse` is not very clear about this behavior.)

Issue #242, opened 2017-10-16
Browser preview plugin generates critical warnings on start up


jeremypw:
With the Browser preview extension activated a large number of critical warnings are shown when launching from the commandline.

danrabbit:
Browser plugin was removed, so this is fixed yay :tada:

PR #253, opened 2017-10-29
Reduce the amount of criticial warnings at startup

tintou:


jeremypw:
Warnings reduced but browser preview plugin still giving a fatal warning at start up.

tintou:
@jeremypw don't hesitate to open a bug ;)

jeremypw:
See #242

PR #262, opened 2017-11-06
Set preferred style scheme to solarized-light

codygarver:


codygarver:
This is actually meant to be set by default settings branch, I only noticed it wasn't in master because I renamed the schema so the OS settings wasn't applied.

PR #266, opened 2017-11-07
Rename scratchcore

codygarver:


codygarver:
This PR wasn't getting any attention so I've included this change in PR #292

Issue #246, opened 2017-10-17
Add a filter entry for syntax highlighting


danrabbit:
because "Fuck dropdowns"

aleksandar-stefanovic:
Something like this?
![screenshot from 2017-12-11 14 52 04](https://user-images.githubusercontent.com/6058082/33834382-0f509b32-de83-11e7-93f6-07915a8baccd.png)

If so, should the `Entry` float upon scrolling, or be scrolled with the rest of the content?

davidmhewitt:
Fixed in https://github.com/elementary/code/pull/287

PR #287, opened 2017-12-11
FormatBar.vala: Added language list filtering (issue #246)

aleksandar-stefanovic:
- Added the `Entry` above the list to be used when filtering.
- Used the  built-in filtering of the `ListBox` to filter the entries.

Not sure if I've created the layout perfectly (regarding the margins and the spacing), so please, comment if the values need changing.

Also, should the search bar float or be scrolled with the content? Currently, it scrolls with the content.


src/Widgets/FormatBar.vala
```diff
@@ -84,13 +85,31 @@ public class Code.FormatBar : Gtk.Grid {
...
             return ((LangEntry) row1).lang_name.collate (((LangEntry) row2).lang_name);
         });
+        
+        lang_selection_filter = new Gtk.Entry();
+        lang_selection_filter.margin_left = 4;
+        lang_selection_filter.margin_right = 4;
+        lang_selection_filter.margin_top = 4;
+        lang_selection_filter.placeholder_text = _("Filter languages");
+        lang_selection_filter.changed.connect(() => {
+            lang_selection_listbox.set_filter_func((row) => {
```
donadigo: You should not set up the filter function everytime the filter changes. Call `set_filter_func` once and check if the entry contains any text

src/Widgets/FormatBar.vala
```diff
@@ -84,13 +85,31 @@ public class Code.FormatBar : Gtk.Grid {
         lang_selection_listbox.set_sort_func ((row1, row2) => {
             return ((LangEntry) row1).lang_name.collate (((LangEntry) row2).lang_name);
         });
+        
+        lang_selection_filter = new Gtk.Entry();
+        lang_selection_filter.margin_left = 4;
+        lang_selection_filter.margin_right = 4;
+        lang_selection_filter.margin_top = 4;
+        lang_selection_filter.placeholder_text = _("Filter languages");
+        lang_selection_filter.changed.connect(() => {
```
donadigo: Space before braces.

src/Widgets/FormatBar.vala
```diff
@@ -84,13 +85,31 @@ public class Code.FormatBar : Gtk.Grid {
...
+            lang_selection_listbox.set_filter_func((row) => {
+                //Both are uppercased so that the case doesn't matter when comparing.
+                return (((LangEntry) row).lang_name.up ().contains (lang_selection_filter.text.up ()));
```
donadigo: I think we should use use down here, since that's what we use elsewhere

aleksandar-stefanovic: Can you explain what do you mean by that? What should we use, exactly?

donadigo: Yes, sorry, I didn't phrase this well. We should use the `string.down ()` method instead of `string.up ()`. I think it's just more natural.

src/Widgets/FormatBar.vala
aleksandar-stefanovic: But, the lambda expression values are calculated once, and so, if the `set_filter_func` is used only once, it doesn't use the new value of the entry. Am I wrong? Is there a workaround?

I've tried setting once and invalidating, but it doesn't work (it doesn't filter the entries).

aleksandar-stefanovic:
I'm working on getting the search bar separated from the `ScrollWindow`, but I'm stuck, because every time I change the child widget of a popup, it just bugs out, and doesn't display any child widget, and language toggle doesn't display the language, and I can't understand why... I will post the commit as soon as I fix that...

aleksandar-stefanovic:
The bug fixed in the last commit was that, when the last character was deleted, it wouldn't re-run the filtering, and instead left the list with the items filtered by the last remaining character.

aleksandar-stefanovic:
Polite ping, can anyone review this?

src/Widgets/FormatBar.vala
```diff
@@ -84,12 +85,24 @@ public class Code.FormatBar : Gtk.Grid {
         lang_selection_listbox.set_sort_func ((row1, row2) => {
             return ((LangEntry) row1).lang_name.collate (((LangEntry) row2).lang_name);
         });
+        lang_selection_listbox.set_filter_func ((row) => {
+            //Both are lowercased so that the case doesn't matter when comparing.
+            return (((LangEntry) row).lang_name.down ().contains (lang_selection_filter.text.down ().strip ()));
+        });
+        
+        lang_selection_filter = new Gtk.Entry();
+        lang_selection_filter.margin = 4;
```
donadigo: The general to rule is that margins should be a multiplier of 3, so I think this should be just `3`.

src/Widgets/FormatBar.vala
```diff
@@ -84,12 +85,24 @@ public class Code.FormatBar : Gtk.Grid {
         lang_selection_listbox.set_sort_func ((row1, row2) => {
             return ((LangEntry) row1).lang_name.collate (((LangEntry) row2).lang_name);
         });
+        lang_selection_listbox.set_filter_func ((row) => {
+            //Both are lowercased so that the case doesn't matter when comparing.
+            return (((LangEntry) row).lang_name.down ().contains (lang_selection_filter.text.down ().strip ()));
+        });
+        
+        lang_selection_filter = new Gtk.Entry();
```
donadigo: Space before brackets.

src/Widgets/FormatBar.vala
```diff
@@ -21,6 +21,7 @@ public class Code.FormatBar : Gtk.Grid {
     private Gtk.SourceLanguageManager manager;
     private FormatButton lang_toggle;
     private Gtk.ListBox lang_selection_listbox;
+    private Gtk.Entry lang_selection_filter;
```
donadigo: I think`Gtk.SearchEntry` would be more appropriate here than a regular entry.

src/Widgets/FormatBar.vala
aleksandar-stefanovic: I've found `6` looking better, but if you think that we should use `3`, feel free to say it.

PR #300, opened 2018-01-09
Add brackets completion by default

danrabbit:
This plugin got majorly better recently


tintou:
I'm generally not a fan of completing features as it's invasive and prevents linear coding

jeremypw:
I am also against making an intrusive feature on by default (at least until it can guess the writers intention 99+% if the time)

danrabbit:
Gonna close this since there's objections and we've identified some issues with the bracket completion

cassidyjames:
Reopening to fix #667

tintou:
Wow, this has been reopened after objections, got 0 discussion, 0 tests and breaks the app :clap:

cassidyjames:
@tintou objections from 16 months ago before major efforts to improve the plugin. Instead of passive-aggression, could you point out what is broken so we can fix it?

PR #306, opened 2018-01-13
Add quick style switcher like Terminal

### LINKED ISSUE #127: Some way to use dark mode
I've been thinking about this a lot and I really think we should offer some method of using the dark stylesheet in Scratch. Every code editor I can think of offers a dark theme, and almost everyone I see using a code editor is using a dark theme.

I personally have resorted to using the Inspector to toggle it on at the beginning of my session because staring at a bright UI all day hurts my head, but that's a pretty hacky solution :stuck_out_tongue:

I know back in the day someone wrote a patch to automatically request the dark theme if "Dark" appeared in the syntax theme name, but that wasn't very smart (i.e. it didn't work on dark syntax themes like "Oblivion"). An alternative would be to just carry a list of dark/light syntax themes and toggle the stylesheet based on that, but at that point I think it might just be better to have a dark mode toggle.

danrabbit:
Todo:
- [x] restore the gtk light/dark theme

Fixes #127


cassidyjames:
Oh yes please. I like where this is going. :D

data/io.elementary.code.gschema.xml
```diff
@@ -144,6 +144,11 @@
 			<summary>Remember the last focused document.</summary>
 			<description>Restore the focused document from a previous session when opening Scratch.</description>
 		</key>
+		<key name="prefer-dark-style" type="b">
+			<default>false</default>
+			<summary>Defined whether to use dark stylesheet from Gtk</summary>
```
cassidyjames: This is worded a little awkwardly. Perhaps "Request dark Gtk stylesheet variant" or something?

Issue #305, opened 2018-01-12
"Strip trailing whitespace" plugin occasionally eats files


cassidyjames:
If I enable the "Strip trailing whitespace" plugin, sometimes when Code saves my files, the contents are actually completely deleted from the disk, even though the code is still visible in the editor. Changing a whitespace character and re-saving appears to restore the contents of the file. 

This is observable when building an app; if you make a change in Code and then go to compile, you get errors about classes from the open file not existing. If you commit the code, you see that the entire contents of the file has been deleted, even though it's still visible in the open Code window.

Here's an example of one such case; I changed a line, switched to Terminal, and committed it. This is the resulting commit:
https://github.com/cassidyjames/dippi/commit/ef5778d325e19ec8097ad7de937f51c00ef2f2a2

I changed a whitespace character and re-committed it. This is the resulting commit: 
https://github.com/cassidyjames/dippi/commit/5c45d39fe2f0d7acd0bb9c4116f5d3fa6c025888

cassidyjames:
The scary thing is if I would have closed Code after changing a line, I'm not sure if I would have completely lost the file or not. Luckily in this case I could see from the Terminal that something was weird while Code was still open.

danrabbit:
@cassidyjames can you see if you can still reproduce the issue after David's branch just got merged?

cassidyjames:
I'm pretty sure it was fixed, but I'll keep my eye out for it.

davidmhewitt:
Going to close this as I've not seen this happen for a while and I'm guessing nobody else has either?

Can always be reopened!

PR #316, opened 2018-01-17
Use a regular expression to strip trailing whitespace

davidmhewitt:
This might help with #305 , it definitely contains 100% less magic now. So we'll see.


tintou:
Is there a reason to not use https://valadoc.org/gtk+-3.0/Gtk.TextBuffer.text.html directly instead of using first and last iter ?

davidmhewitt:
@tintou That sounds like it might be better! I'll check that later when I'm back from work. I just assumed that there was a reason the existing code needed to get the text using the iterators. But I shouldn't have assumed that any of that code was reasonable 😉

PR #318, opened 2018-01-20
Draw tabs/spaces using native GtkSourceView methods

davidmhewitt:
Since the version of GtkSourceView we had in Loki, there have been some additions that allow you to draw tabs/spaces between two points which means we can quite nicely implement the "Draw spaces for selection" functionality using that.


cassidyjames:
How much of a maintenance burden would carrying our own solarized styles be? If it's not high and it lets us do this, I'm up for it.

davidmhewitt:
@tintou You're right, I can't remember why or when I touched that line, but it's definitely not necessary for this so I've reverted the change.

PR #319, opened 2018-01-20
Don't set prefs dialog modal

danrabbit:
https://elementary.io/docs/human-interface-guidelines#preference-dialogs

"Preference dialogs should be made Transient, but not Modal. When a user makes a change in a preference dialog, the change should be immediately visible in the UI. If the dialog is modal, the user may be blocked from seeing (and especially from interacting with) the change. This means they will have to close the dialog, evaluate the change, then possibly re-open the dialog."

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -690,6 +690,10 @@ namespace Scratch {
             if (preferences_dialog == null) {
                 preferences_dialog = new Scratch.Dialogs.Preferences (this, plugins);
                 preferences_dialog.show_all ();
+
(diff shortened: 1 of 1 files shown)


danrabbit:
Oops you’re right

davidmhewitt:
I know you were confused as to why your pointer wasn't being set to `null` again after destroying the dialog.

You probably want to use this:
https://valadoc.org/gtk+-3.0/Gtk.Widget.destroyed.html

davidmhewitt:
Or... Connect to the destroy signal yourself and set the pointer to null. Same end result.

PR #320, opened 2018-01-20
Toggle code comments support

### LINKED ISSUE #313: Toggle line comment
A feature that I miss from Atom is the shortcut to toggle a line-level comment. It's super handy for commenting/uncommenting a line of code quickly while testing things.

In Atom, it's `Ctrl`+`/`. In Code that appears to select all text, so I'm not sure if we want to override that or come up with a different shortcut.

davidmhewitt:
Fixes #313 

Adds a plugin that, when enabled, allows you to toggle comments for the selected line(s) on and off using the Ctrl+M shortcut. Should in theory work with any language that GtkSourceView has the right metadata for, so it knows what the comments look like.

I built this as a plugin initially so I could keep the functionality nice and clean and separate, but would be happy to build this into the main functionality of the editor if desired.

The logic goes as follows:

* If you don't have a selection and the shortcut is pressed, comment/uncomment the line the cursor is on.
* If you have a section of the file selected and press the shortcut, the plugin should comment/uncomment all of the lines that selection touches based on the following logic:
    * If all of the non-blank lines in the selection are already commented, we assume the user wants to uncomment this section and remove the comments from all lines in the selection.
    * If none or some of the lines in the selection are commented, we assume the user wants to comment out that section. For example, they may have selected a block of code that had some comments in amongst the lines already.
* All of this should work for both line comments and block comments (i.e `//` and `/* */` in Vala).
* When automatically commenting code, this plugin automatically prefers to create one comment per line rather than block comments, even for languages that *only* have block comment tags. For example, in HTML, it will create one `<!-- -->` section per line rather than over the whole block.
* The above point helps the user in selectively uncommenting lines as the line comments can be removed from the lines in question rather than trying to recalculate how the block comments should work when removing a couple of lines from the block.
* This uses the metadata for each programming language shipped by GtkSourceView. I don't know how extensive the metadata is, so it may not work with every language, but I've tested Vala, Python and HTML.


davidmhewitt:
@cassidyjames I'd agree with those comments, so I've made some changes towards that. Still some work to do around unselecting the text after toggling the comment. But, I've switched to preferring single line comments now and the comments should start at the indent point of the code.

A question came up while doing this though:

For languages that don't have a way of doing single line comments (i.e. HTML). Do we change the logic to (ab)use the block comment features of those languages to do single line comments. For example, when hitting the toggle comment shortcut while having 10 lines selected in a HTML document, do we make 10 block comments, one for each line. Or, do we just make one (how the plugin currently works)?

davidmhewitt:
I answered some of my own questions when working on this.

I think for languages like HTML that don't have a way of doing a single line comment, it makes sense to use a block comment for each line when using this plugin to automatically toggle comments.

If the user is using this functionality, they'll know they can also remove comments using it. It's much easier to remove comments programatically when there's one comment per line.

I have however, left in support for detecting and removing multi-line block comments so if someone comes across a file with some in, they can remove them using this plugin and then when they get toggled back on, it'll switch to using one comment per line.

So, this is ready for review now. Test it out in a variety of situations and let me know if it does any unexpected things anywhere :smile:

cassidyjames:
@davidmhewitt I'm into nitpick territory now, but

- would it be possible to preserve/restore the selection state after toggling the comment? I.e. if I comment a block out, right now it deselects whatever I had highlighted.
- it appears to double-comment the commented line if you select a commented and non-commented line, then hit Ctrl+M (tested in Python)

cassidyjames:
- I also found something else, hopefully I can describe. If you select a multi-line comment, but the contents of the first line is just whitespace (like this in Vala/countless languages):

```/*
 * Block comment
 */
```

It doesn't detect it as if I selected the first line, so it adds line-level comments on the left of all but the first line. It looks like Atom just doesn't do anything fancy with block comments which is interesting, it just throws the `//` at the beginning of all the lines.

If you actually select the `/*` on the first line it works as expected.

cassidyjames:
It looks like if you have the last character on a line highlighted (manually or automatically), it will affect the next line as well. So if you comment out two lines, then the plugin selects both lines, then you try to toggle comments off, you end up re-commenting those two lines plus the next line. Ad infinitum.

danielpinto8zz6:
Hi guys, 
there is a bug, when you try to comment the last line

davidmhewitt:
@danielpinto8zz6 Hey! Thanks for testing and spotting that, should be resolved now.

PR #321, opened 2018-01-21
Don't try to load gjs libpeas loader

TingPing:
- The `gjs` loader has been removed from upstream for many years
- Enabling two dynamic loaders (python and gjs) is broken in practice as they both leak references
- None of the plugins in-tree use it

(Sidenote it is probably time to move `python` to `python3`)


tintou:
Thanks, this was also bugging me or years :)

PR #324, opened 2018-01-28
Port to meson build

### LINKED ISSUE #156: plug-in description aren't translatable
The description of Scratch's plugins aren't transable so they always appear in English.

davidmhewitt:
Also fixes #156 

It did get smaller with meson instead of CMake, but then I added a bunch of `.po` files to fix the above issue.


danrabbit:
In case anybody is looking at translations and seeing that the desktop file doesn't seem to be translated, this will be fixed after running a pot and po update and isn't a bug in this branch

Issue #323, opened 2018-01-26
 Support .editorconfig 


cassidyjames:
From Launchpad [bug 1547574](https://bugs.launchpad.net/scratch/+bug/1547574)
>Several projects I work on use .editorconfig, a file defined at http://editorconfig.org/. Other text editors support this out of the box or through a plugin.
>
>The gist is that a file in the root of a project defines that project's code style, and can even be set up per-language in case of conflicting languages in one project. It's invaluable for distributed projects where people are working in different projects with different code styles.

PR #345, opened 2018-03-14
Add EditorConfig support

pantor:
This adds support for EditorConfig files (#323). It currently does not support all properties (full list [here](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties)), e.g. charset or end_of_line are missing.


cassidyjames:
@pantor if this is merged, could you open issues for each unsupported property? It'd be great to track that and increase the coverage eventually.

plugins/editorconfig/editorconfig.vala
```diff
@@ -0,0 +1,72 @@
...
+                        view.indent_width = int.parse (val);
+                        break;
+                    case "tab_width":
+                        view.tab_width = int.parse (val);
+                        break;
+                    case "end_of_line":
+                        break;
+                    case "charset":
+                        break;
+                    case "trim_trailing_whitespace":
```
cassidyjames: Adding this would depend on enabling the "Strip trailing whitespace" plugin based on the config. Is that a) something that is possible, and b) something we'd be comfortable doing?

cassidyjames: @danrabbit mentioned that the strip whitespace plugin is something we should probably pull into the core app anyway, so this may be a moot point.

Regardless, it's out of the scope of this PR.

pantor:
The Headerbar will now show the local settings of the current document, the popover however still shows the global settings. If you change them, the current document will use the global settings.

I needed to change the position of the `plugins.hook_document (doc);` in `src/MainWindow.vala`. I hope this doesn't break anything.

pantor:
Oh, and in Code there is some confusion between tab_width and and indent_width. While this is very apparent with this PR, it is probably another issue.

PR #347, opened 2018-03-15
Test with Houston CI

danrabbit:


codygarver:
I don't know that 16.04 is still the same docker container we want to pull in based on the conversation we had in Houston the other day, ask @btkostner

btkostner:
Houston will pull and build all the needed docker containers during the ci process. The pull here is just to hopefully speed up the process. If you really want, you could remove that line and all would be well.

PR #349, opened 2018-03-15
Don't try to be too clever with Pastebin API

### LINKED ISSUE #288: Pastebin-Extension "This page has been removed"
using the pastebin extension to share text succeeds but the paste is unavailable.

davidmhewitt:
Fixes #288 

The Pastebin API has obviously changed since this was written so none of the assumptions made were valid anymore.

I've replaced the not so nice parsing with a check to see if what's returned is a URI. If it is, that's our paste URI and present it as a link to the user. If it's not a link, it's an API error message, display that as text to the user.

P.S. There's a horrible mix of code styles and tabs and spaces in these files. I don't intend to fix any of that with this PR.


Shnatsel:
Why is this a Code plugin and not a Contractor item?

PR #352, opened 2018-03-15
Bring developer brand colors to icons

micahilbery:


fabiozaramella:
Nice :100:

tintou:
I don't know how i feel about it, we're loosing the alignment with Solarized and that purple is really flashy

cassidyjames:
I think I'm gonna close this. The purple here is a bit much and the current icons match the UI pretty well. Feel free to re-open if I'm off base!

danrabbit:
I think I agree with Cassidy. I like the idea of the tie-in with our developer brand, but I also really like the WYSIWYG nature of using the solarized light and dark colors

PR #357, opened 2018-03-16
Add open folder in welcome view

### LINKED ISSUE #398: New file from clipboard doesn't appear to work
This seems to have broken at some point in Juno. It's possible some Mutter or GTK changes for Wayland broke it.

pantor:
Fixes #398 

This PR replaces the new-from-clipboard action with an open-folder action in the welcome view. As Code shifts to a full code editor, the new-from-clipboard action for small text edits is probably needed much less.

I think it would be good to rename folder to project.


src/Widgets/WelcomeView.vala
```diff
@@ -34,7 +34,7 @@ public class Code.WelcomeView : Granite.Widgets.Welcome {
         vexpand = true;
         append ("document-new", _("New file"), _("Create a new empty file."));
         append ("document-open", _("Open file"), _("Open a saved file."));
-        append ("edit-paste", _("New file from clipboard"), _("Create a new file from the contents of your clipboard."));
+        append ("folder-saved-search", _("Open folder"), _("Open a saved folder."));
```
cassidyjames: If we do this, we probably want to better explain what this does. Something like, "Add a project folder to the sidebar" or something?

PR #359, opened 2018-03-16
On change tab, select file in sidebar if visible

### LINKED ISSUE #6: Can not open the same file after closing it
If you close a file (when using both filemanager or folder manager plugins) the file stays highlighted in the sidebar - so you can't open it again without clicking another file or collapsing and expanding the folder in the sidebar.

Launchpad Details: [#LP1382558](https://bugs.launchpad.net/bugs/1382558) Tomo Moto - 2014-10-17 13:28:26 +0000

davidmhewitt:
Fixes #6 

When changing tab, select the item belonging to the new tab in the sidebar if it is visible (i.e. expanded) in the folder manager scroll view.

Otherwise, deselect the currently selected item in the folder manager, so it can be clicked again if necessary.


davidmhewitt:
Marking as in progress as this can be optimized by not searching irrelevant paths for the file.

PR #375, opened 2018-03-21
Prevent strip trailing plugin from jumping document to top on save

davidmhewitt:
The current way of doing this was just to do a regex search and replace on the entire buffer, removing any trailing spaces.

On replacing the text in the buffer with the new stripped text, GtkTextView drops all text in the buffer (causing it to scroll back to the top), then puts the new text in.

I did look at some workarounds that restored the original scroll position after the text replacement, but they were a bit hacky and you visually saw the jump. So, I'm proposing going back to doing this line by line (still with less horrible C pointer magic), so we can use the proper TextBuffer editing methods and avoid the scrolling.


cassidyjames:
At a first glance UX-wise this seems to do the thing. Should still get someone to review the code, though.

plugins/strip-trailing-save/strip-trailing-save.vala
```diff
@@ -73,14 +73,32 @@ public class Scratch.Plugins.StripTrailSave: Peas.ExtensionBase, Peas.Activatabl
...
         var text = buffer.text;
 
-        try {
-            var regex = new Regex ("[ \t]+$", RegexCompileFlags.MULTILINE);
-            text = regex.replace (text, -1, 0, "");
-        } catch (RegexError e) {
-            warning ("Error while replacing trailing whitespace: %s", e.message);
+        string[] lines = Regex.split_simple ("""[\r\n]""", text);
+        if (lines.length != buffer.get_line_count ()) {
+            warning ("Mismatch between line counts when stripping trailing spaces, not continuing");
```
donadigo: I think we established that we would use `critical` for warnings like that?

plugins/strip-trailing-save/strip-trailing-save.vala
```diff
@@ -73,14 +73,32 @@ public class Scratch.Plugins.StripTrailSave: Peas.ExtensionBase, Peas.Activatabl
...
+                if (regex.match (line, 0, out info)) {
+                    buffer.get_iter_at_line (out start_delete, line_no);
+                    start_delete.forward_to_line_end ();
+                    end_delete = start_delete;
+                    end_delete.backward_chars (info.fetch (0).length);
+
+                    buffer.@delete (ref start_delete, ref end_delete);
+                }
+            } catch (RegexError e) {
+                warning ("Error while replacing trailing whitespace: %s", e.message);
```
donadigo: Same here.

plugins/strip-trailing-save/strip-trailing-save.vala
```diff
@@ -73,14 +73,32 @@ public class Scratch.Plugins.StripTrailSave: Peas.ExtensionBase, Peas.Activatabl
...
+        if (lines.length != buffer.get_line_count ()) {
+            warning ("Mismatch between line counts when stripping trailing spaces, not continuing");
+            return;
         }
 
-        buffer.text = text;
+        MatchInfo info;
+        int line_no = 0;
+        TextIter start_delete, end_delete;
+        foreach (var line in lines) {
```
donadigo: If we need the line number here, I guess you could just use the standard `for (int line_no = 0; line_no < lines.length; line_no++)`?

PR #376, opened 2018-03-22
Don't add temporary GLib file

tintou:
This temporary file made the file view flicker on file modification


danrabbit:
CI is broken, I can confirm this builds

PR #377, opened 2018-03-23
Move the outline to the right allowing it to be document-dependant

tintou:


danrabbit:
screenshot from 2018-03-22 18 18 43 2x

Hrm this seems really weird to me

I think one problem is that it kind of feels like we're not sure where the sidebar should be

There's also a bit of an oddity here in how the web preview pane is working. Maybe this is just an issue that's exposed now because we didn't have other plugins on the right side before, but it's weird that when you click the web preview button, it just adds a tab and doesn't actually show you the web preview.

Maybe @cassidyjames has some more feedback, but I don't really like how this feels :/

tintou:
If the question is "should the file tabbar come in-between", remember that not all files are using this plugins and that it makes it unusable when the pane appear/disappear when changing pages as the tabs are then resizing…

tintou:
capture d ecran du 2018-03-23 01 30 17 2x
A more real screenshot

danrabbit:
One thing we could do is make the sidebar hide and close manual and just make the tabs insensitive or have placeholder when they're not useful. This would stop the UI from jumping around between tabs.

tintou:
I'm still persuaded that this belongs to document context and thus has nothing to do outside of it :man_shrugging:

tintou:
capture d ecran du 2018-03-23 01 42 03 2x
With multicolumns

danrabbit:
Yeah with that context of multiple files open that makes sense that the symbols are per-view. I'm also really aware though that here there's 3 times as much chrome as in the current design and no way to hide it without disabling the symbols plugin.

tintou:
I've worked with this a few days and I'm really satisfied because I can finally use this plugin. I quite never used it before because It was hidden by the filemanager pane that I use more often and It's a real pain to have to switch between tabs to see one or the other.

tintou:
Ofc, we need a way to hide the pane in some cases but those settings will really be useful once we get real Project management and not file management like it's done right now

cassidyjames:
Hmm… I agree with @danrabbit that it's a LOT more chrome. I really like the current design where I can switch between contexts. However, I agree that I end up using the Symbols view a lot less because I switch between files using the sidebar a lot. 

I do like the idea that the folder sidebar belongs to the window, and the symbols sidebar belongs to the document. If we implement manual sidebar showing/hiding, then I could be convinced that document-level sidebars belong on the right.

cassidyjames:
@tintou any chance to bring this up to date and address @jeremypw's issue?

jeremypw:
@tintou Is this abandoned?  The design team does not seem to have squashed it entirely.  I agree with you that the symbol table would be a lot more usable if you did not have to keep switching tabs in the sidebar.  It has some parallels with the mini-map, which also shows on the right.

jeremypw:
The web-preview plugin is now removed so that is one less obstacle.

May be we could style the symbol table differently from the sidebar - more like the document?

danrabbit:
@jeremypw yeah now that the split and web views are gone, could be worth revisiting this.

I think I agree that there's some parallels with the minimap. You might be right that styling more like the document could help it feel more like a scroll indicator

jeremypw:
@tintou I would suggest closing this PR and opening an issue where the relevant design can be discussed.

PR #394, opened 2018-04-14
Accept colon in line number

### LINKED ISSUE #390: Accept colon as line:character separater in Line Number entry
For the go-to-line functionality, it would be handy to accept the colon separator. This is how every other text editor and language I've seen represents line:character, including vala-lint.

cassidyjames:
Fixes #390


cassidyjames:
We discussed the reason we're using the entry text directly in Slack, but I'll address the other concern.

PR #425, opened 2018-05-16
Copy current line when no selection

davidmhewitt:
To match the Ctrl+X cuts a line when there's no selection functionality.


jeremypw:
Needs UX decision as to what the default behaviour should be and how (if) to expose any setting to the user.

cassidyjames:
As much as I hate adding more fiddly knobs… would it make sense to put these under a "Smart copy/paste" toggle in the settings? To me it makes sense to have on by default, because it's a low-risk mistake if you hit the wrong key (you can just undo). But I can understand some resistance to not doing what you'd expect.

jeremypw:
Just to note: you cannot undo pressing Ctrl+C instead of Ctrl+V - the contents of the clipboard are overwritten and undo does not reverse this.

A refinement might be to have smart copy/cut work only when there is nothing pending in the clipboard but that might be considered to be too 'smart' and leads to unpredictable behaviour.

jeremypw:
@cassidyjames :  Can you suggest a string that would meet with approval?

cassidyjames:
@jeremypw I'm working on that now. :smile:

cassidyjames:
@jeremypw @davidmhewitt and @elementary/ux feel free to check #611. If we're going to add and support this feature, I think something like that is the way to do it. I first tried with a dim label (like we've been doing in System Settings lately), but the smaller design of the dialog made it feel really awkward and lopsided. Input/suggestions welcome, though.

PR #429, opened 2018-05-17
Several improvements to bracket completion logic

### LINKED ISSUE #308: Brackets completion over-agressively removes characters
Looks like fallout from #31. If I backspace a quote that's next to another closing bracket, the plugin also erroneously removes the unrelated bracket.
gifout

### LINKED ISSUE #309: Quote completion should skip over matching closing quotes
One awesome feature of Atom that keeps the bracket completion out of my way is that it is smart about when to draw a closing bracket/quote character. For example, if you manually type `"foo"`, it knows that the last `"` you typed is the same as the one it autocompleted for you, so it just passes over it without duplicating it. 

This also means you don't have to do arrow key gymnastics to write more naturally, even with bracket/quote completion on.

### LINKED ISSUE #344: Brackets completion interferes with compose key
When I'm trying to type my name on a copyright header, it ends up as `For'e'` instead of `Foré` because bracket completion doesn't know about the compose key :p

davidmhewitt:
Fixes #344 
Fixes #309 
Fixes #308 

Most of these fixes were pretty interwoven, so it's all ended up in one diff.

### DIFF
--- a/plugins/brackets-completion/brackets-completion.vala
+++ b/plugins/brackets-completion/brackets-completion.vala
@@ -67,14 +67,16 @@ public class Scratch.Plugins.BracketsCompletion : Peas.ExtensionBase,  Peas.Acti
         current_buffer = doc.source_view.buffer;
 
         if (current_source_view != null) {
-            current_source_view.key_press_event.disconnect (on_key_down);
+            current_source_view.deselected.disconnect (on_deselect);
+            current_source_view.selection_changed.disconnect (on_selection);
             current_source_view.event_after.disconnect (on_event_after);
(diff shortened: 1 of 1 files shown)


davidmhewitt:
Marking as WIP since I seem to have introduced a crash somewhere

davidmhewitt:
Seems like the crashing issues I was having are a conflict with the highlight word selection plugin. Not sure what's up yet, will have to investigate further.

davidmhewitt:
I believe the crashes will be resolved after merging #430 , so that should be reviewed and merged first. Marking as blocked for now.

danrabbit:
I can confirm that this fixes #308, #309, and #344 :)

PR #433, opened 2018-05-19
Draw trailing spaces regardless of selection

### LINKED ISSUE #399: Always draw trailing spaces
It looks like it's something we can do in GTK Sourceview, and it might be handy since linting complains about trailing spaces, and if they're intentional (like with breaks in markdown), you probably want to see them anyway.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/57394181-always-draw-trailing-spaces?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #399


jeremypw:
Another problem is that, especially with a dark theme, this PR causes selected trailing spaces to be drawn more faintly than leading spaces.

davidmhewitt:
@jeremypw Yeah, that was intentional, since I think the "never" draw spaces option should probably "never" draw spaces.

I'll mark this as needs design for now so we can decide if it should indeed be always even if the "never" option is chosen.

cassidyjames:
Yeah, "never" should never draw the spaces.

PR #434, opened 2018-05-19
Build style switch buttons dynamically from gsettings

### LINKED ISSUE #395: Split style scheme GSettings into light and dark
Right now we have one style scheme GSetting that we manually change every time you use the light or dark buttons in the style switcher. Instead, we could have a light GSetting and dark GSetting, and just toggle between them with the buttons.

- [ ] Change the UI to just say, "Light" and "Dark" instead of anything about "Solarized" OR programmatically grab the name of the relevant style scheme
- [ ] Create the new GSettings
- [ ] Change the scheme based on the GSetting
- [ ] Make sure we still load the correct one at runtime

This makes it a little easier for downstreams to set their own default light and dark color schemes if they want.

davidmhewitt:
Fixes #395 

Adds a key to gsettings that contains a list of style schemes to be shown on the buttons in the menu. The buttons are then built by loading colours and titles from the style scheme itself.

Allows downstreams to override which styles they want to display a selection of easier.


src/Widgets/HeaderBar.vala
```diff
@@ -93,35 +118,39 @@ namespace Scratch.Widgets {
...
+                                background-color: %s;
+                            }
+                        """.printf (style_data.background);
+                        css_provider.load_from_data (background_css);
+
+                        button_context.add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+                    }
+
+                    style_buttons.add (button);
+                }
```
jeremypw: What happens if style_id not found in sssm.scheme_ids?  The button still gets added to grid it seems.

jeremypw:
Last commit lays out buttons evenly if > 3 styles.  Also applies actual style to the background and text of the button.  Focuses the current style button on opening menu.

Styling details for example only. To be specifies by UI team.

jeremypw:
@davidmhewitt Converting to draft as has conflicts, unaddressed issues and no recent activity.  Please feel free to close or reopen for review after fixing conflicts as desired.

PR #446, opened 2018-05-23
Sort selected lines with F5 shortcut

### LINKED ISSUE #443: Add line sort shortcut
One of the features I love in atom is the ability to highlight a block of text, press `F5`, and have each line be sorted alphabetically. While this is less useful for desktop programming, it's super helpful for CSS files.

davidmhewitt:
Fixes #443


cassidyjames:
@davidmhewitt I think for discoverability, we should also put this in the context menu for highlighted text.

jeremypw:
Presumably this will convert DOS line endings to UNIX line endings for the sorted lines.  Does this matter?  I am not sure whether Code is supposed to preserve DOS line endings - if not it would be better to convert all of them.

jeremypw:
Note some editors make the type of line ending a document property and force all line endings to be the same on loading.

davidmhewitt:
@jeremypw Thanks for point that out. I've now changed this to include the line endings that were already on the end of each line in the split. This should account for all cases (even files that have mixed line endings).

davidmhewitt:
@jeremypw Ok, I think **all** the cases are handled now! Thanks again for testing this.

jeremypw:
@davidmhewitt  Unfortunately my main machine is not operational at the moment so I am sending it away for fixing.  I only have a very old and slow 32 bit machine to work with so I cannot properly test again for a while.  I do not think I selected a blank line but it is only an obscure corner case so I will approve this PR and any problems arising can be fixed later.

PR #450, opened 2018-05-24
Use builder blocks font in source map

### LINKED ISSUE #444: Use Builder Blocks font for minimap
GNOME Builder uses a font made to make the minimap much more useful: "Builder Blocks". It's designed to make code in the minimap at 1px size to be more recognizable based on its structure instead of trying to draw actual characters. The difference is very noticable.

## elementary Code with the defaults:
code-minimap

## GNOME Builder with Builder Blocks:
builder-minimap

Builder Blocks is avaialable in the GNOME Builder GitLab: https://gitlab.gnome.org/GNOME/gnome-builder/tree/master/data/fonts

davidmhewitt:
Depends on #463

Fixes #444 

I think we need to double check the license of this font first. But this works technically.


cassidyjames:
The entire repo appears to be GPLv3, but we can clarify with @chergert to double-check.

cassidyjames:
Whoops, wrong button. Sorry!

davidmhewitt:
Travis is failing because we're blocked by #451

chergert:
Yeah for the time being its GPLv3+. However, if we can get it into GtkSourceView, then we can relicense it as LGPLv2.1+. The thing that was holding that back previously was something weird about fonts on Windows.

data/meson.build
```diff
@@ -11,6 +11,10 @@ foreach i : icon_sizes
     )
 endforeach
 
+install_data([
```
danrabbit: Is it possible to add this as a gresource instead of installing it to the filesystem?

davidmhewitt:
@danrabbit AFAICT, no. It would seem that Pango wants to load the font from the filesystem and I wasn't able to find any resources on the interwebs about loading fonts from GResource in this fashion.

chergert:
Correct, except it's not Pango that is at fault, but `FcConfigAppFontAddFile()` which is problematic.

Issue #133, opened 2017-06-28
Show Git status in sidebar


cassidyjames:
From Launchpad [bug 1522155](https://bugs.launchpad.net/scratch/+bug/1522155):
>It would be helpful for a plugin to show Git status (new and changed files) in the projects sidebar when working inside of a Git project. Atom and Sublime both do this and it's immensely helpful when working in a large project. Screenshot of Atom attached.
>
>(Filed separately from [bug 1179250](https://bugs.launchpad.net/scratch/+bug/1179250) because that one seemed to focus on performing version control actions from within Scratch)
>
>![screenshot from 2015-12-02 14-51-04](https://user-images.githubusercontent.com/611168/27648761-cfae00dc-5bec-11e7-841d-727f5d7f1cfc.png)

PR #452, opened 2018-05-25
Show git status in sidebar

davidmhewitt:
Depends on #461 for packaging and a new release of granite with the `markup` property on `SourceList.Item`


cassidyjames:
@davidmhewitt would doing some sort of a dim label next to the title work? Basically just a lighter weight instead of wrapped in parens, I guess.

davidmhewitt:
@cassidyjames Possibly, although the granite sourcelist is super inflexible for that kind of thing. I guess that might be where I have to dedicate my efforts soon.

There are quite a lot of issues in code blocked by issues/inflexibilities of the sourcelist.

davidmhewitt:
This branch is now showing a `(*)` or a `(+)` next to modified or new files as seen by git. This obviously isn't ideal either, since it makes a bit of a mess of renaming them.

But, I feel like that's pretty easily solved by colour instead of text, just got to figure out the best way to apply that to the SourceList.

cassidyjames:
@davidmhewitt I've also added a packaging branch, so hopefully everything's set up right and CI will pass now :crossed_fingers:

davidmhewitt:
It is, yes. I get the impression that some inconsistent things currently happen when you have both added and modified files in the tree, which I need to fix.

However, some UX input into what should happen in that case would be good.

For example, consider the following tree:

- src
  - Widgets
    - WidgetX.vala (modified)
    - WidgetY.vala (added)

Should the little icons on the `src` folder and the `Widgets` folder denote modified, added, or some other state?

cassidyjames:
@davidmhewitt I think modified takes precedence.

davidmhewitt:
As far as I'm concerned, this is no longer in progress and ready to review.

cassidyjames:
This builds and looks great with Granite built from source. I'm not completely sure if that's been released or not, though. Checking into it.

cassidyjames:
Ah, no, that was committed a few days after the latest release. @codygarver can we get a new release of Granite when it's convenient?

davidmhewitt:
New granite version released, no longer blocked.

cassidyjames:
Fixes #133

PR #455, opened 2018-06-01
Spelling: "fifo" → "FIFO"

comradekingu:
As per https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics)


src/Application.vala
```diff
@@ -175,7 +175,7 @@ namespace Scratch {
                                 reason = _("It is a mountable location.");
                                 break;
                             case FileType.SPECIAL:
-                                reason = _("It is a \"special\" file such as a socket,\n fifo, block device, or character device.");
+                                reason = _("It is a \"special\" file such as a socket,\n FIFO, block, or character device.");
```
danrabbit: I think "block device" is important and different from just "block"

comradekingu: Aren't they all devices except for the socket?

PR #476, opened 2018-07-09
Move contractor to the folder manager

### LINKED ISSUE #474: code wont tile with a file open.
With a file open the editor wont tile to half screen, nor by dragging or using shortcuts. It works fine with no file opened.

using elementary os alpha live.

edit:
my screen resolution is 1366x768

danrabbit:
Add contractor actions to the menu in the folder manager and remove the toolbar item

Should fix #474


danrabbit:
Added for folders!

I think I'm okay with losing contractor actions if you're not using the folder manager, but maybe @elementary/ux has other opinions. My line of thinking is that all of the file/folder management functions should be consolidated here. we don't offer other file/folder management actions like "Rename" or "Move To Trash" in the context menu either

jeremypw:
Maybe a compromise would be for the sidebar to always show the current documents so that it never closes (unless all documents are closed).

PR #483, opened 2018-07-21
Spelling: Questionmark following question, cancel

comradekingu:


jeremypw:
In this case, because the description says "Please wait" the meaning of a plain "Cancel" button is not 100% clear, although I agree that it is contrary to elementary HIG to modify the label in this way.  Maybe the solution would be to alter the description to a question e.g. "Cancel loading?".

jeremypw:
"Maybe it is corrupt ..." is not necessarily a question and I do not think it was meant as one here as no answer is expected.  It should be rephrased as "It may be corrupt or ..."

jeremypw:
Flagged UX team for their opinion on a change in wording from "Please wait ..." when the button just reads "Cancel".

src/Services/Document.vala
```diff
@@ -267,7 +267,7 @@ namespace Scratch.Services {
                     var title = _("Loading File \"%s\" Is Taking a Long Time").printf (get_basename ());
                     var description = _("Please wait while Code is loading the file.");
                     var alert_view = new Granite.Widgets.AlertView (title, description, "dialog-information");
-                    alert_view.show_action (_("Cancel Load"));
+                    alert_view.show_action (_("Cancel"));
```
cassidyjames: I think we avoid "Cancel" on its own because in this context it is not always clear if it is to cancel the dialog or the action. Pairing it with the noun of what it is canceling is more explicit.

src/Services/Document.vala
```diff
@@ -677,7 +677,7 @@ namespace Scratch.Services {
         // Show an error view which says "Hey, I cannot read that file!"
         private void show_default_load_error_view () {
             var title = _("File \"%s\" Cannot Be Read").printf (get_basename ());
-            var description = _("Maybe it is corrupt or you do not have the necessary permissions to read it.");
+            var description = _("It may be corrupt, or you lack the permission to read it.");
```
cassidyjames: I think I'd drop the comma and be a little less technical:

```suggestion
            var description = _("It may be corrupt or you do not have permission to read it.");
```

comradekingu: Corrupt was already technical. "Damaged" is not. Functionally the second part is not part of the first, so as to avoid "corrupt or → something relating to how it is corrupt", the comma serves a purpose.

src/Services/Document.vala
comradekingu: That would be "Close". "Stop" would be even more explicit.

cassidyjames: Buttons are title case https://elementary.io/docs/human-interface-guidelines#capitalization

cassidyjames: "Cancel Loading" or "Stop Loading" could probably work?

src/Services/Document.vala
```diff
@@ -677,7 +677,7 @@ namespace Scratch.Services {
         // Show an error view which says "Hey, I cannot read that file!"
         private void show_default_load_error_view () {
             var title = _("File \"%s\" Cannot Be Read").printf (get_basename ());
-            var description = _("It may be corrupt, or you lack the permission to read it.");
+            var description = _("It may be damaged, or you lack the permission to read it.");
```
cassidyjames: As long as we are talking about files, being "corrupt" is the common and correct term. Nobody says a file is "damaged." What I was pointing out was the seemingly-stilted and technical language of "you lack the permission." Instead, I suggested, "you lack permission." "The permission" is a specific technical thing that is unimportant in this context, because "you lack permission" to do something is both a common phrase and is easily understood by users who understand file permissions.

danrabbit:
Closing due to lack of movement in favor of #622

PR #494, opened 2018-08-15
Inline toolbar tabs based off selected style

### LINKED ISSUE #493: Use inline toolbar style on tab bar
Since elementary/stylesheet#379 was merged. This will require:

- [ ] Defining `@define-color tab_base_color $FOO` in application CSS where `$FOO` is the background color of the current syntax scheme. 
- [ ] Changing that variable when the style scheme changes. We _could_ try to introspect into the style scheme to get that color, _or_ we could just define three versions based on the style scheme switcher, since we know what scheme we're going to use there. Just depends on how complex we want to get.
- [ ] Adding the `.inline-toolbar` style class to the Dynamic Notebook widget.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62180436-use-inline-toolbar-style-on-tab-bar?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #493

### DIFF
--- a/src/Widgets/DocumentView.vala
+++ b/src/Widgets/DocumentView.vala
@@ -124,7 +124,7 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
         }
 
         // Fallback to a non inline toolbar if something went wrong above
-        style_context.remove_class ("inline-toolbar");
+        style_context.remove_class (Gtk.STYLE_CLASS_INLINE_TOOLBAR);
     }
 
     private string unsaved_file_path_builder () {
--- a/src/Widgets/DocumentView.vala
+++ b/src/Widgets/DocumentView.vala
@@ -115,7 +115,7 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
 
             var define = "@define-color tab_base_color %s;".printf (color);
             style_provider.load_from_data (define);
-            style_context.add_class ("inline-toolbar");
+            style_context.add_class (Gtk.STYLE_CLASS_INLINE_TOOLBAR);
             return;
         }
 
--- a/src/Widgets/DocumentView.vala
+++ b/src/Widgets/DocumentView.vala
@@ -114,9 +114,13 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
             }
 
             var define = "@define-color tab_base_color %s;".printf (color);
-            style_provider.load_from_data (define);
             style_context.add_class (Gtk.STYLE_CLASS_INLINE_TOOLBAR);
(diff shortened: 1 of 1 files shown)


src/Widgets/DocumentView.vala
```diff
@@ -84,9 +86,43 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
...
+            // Default gtksourceview background color is white
+            var color = "#FFFFFF";
+            if (text_color_data != null) {
+                // If the current style has a background color, use that
+                color = text_color_data.background;
+            }
+
+            var define = "@define-color tab_base_color %s;".printf (color);
+            style_provider.load_from_data (define);
+            style_context.add_class ("inline-toolbar");
```
danrabbit: This should be `Gtk.STYLE_CLASS_INLINE_TOOLBAR`

src/Widgets/DocumentView.vala
```diff
@@ -84,9 +86,43 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
...
 
+    private void update_inline_tab_colors () {
+        var sssm = Gtk.SourceStyleSchemeManager.get_default ();
+        var style_context = get_style_context ();
+
+        if (settings.style_scheme in sssm.scheme_ids) {
+            var theme = sssm.get_scheme (settings.style_scheme);
+            var text_color_data = theme.get_style ("text");
+
+            // Default gtksourceview background color is white
```
cassidyjames: I know the default color that inline notebooks pick up is white. Do we still want/need to define this here?

src/Widgets/DocumentView.vala
```diff
@@ -84,9 +86,47 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
...
+            try {
+                style_provider.load_from_data (define);
+                return;
+            } catch (Error e) {
+                critical ("Unable to set inline tab styling, going back to classic notebook tabs");
+            }
+        }
+
+        // Fallback to a non inline toolbar if something went wrong above
+        style_context.remove_class ("inline-toolbar");
```
danrabbit: This one too ;D

src/Widgets/DocumentView.vala
cassidyjames: @davidmhewitt makes sense, thanks for the explanation!

PR #512, opened 2018-09-27
Add Accels for keyboard shortcuts

cassidyjames:
WIP, help wanted :stuck_out_tongue:


jeremypw:
You could add an accel tooltip to the modelbutton.

cassidyjames:
@jeremypw UI-wise I'd prefer the inline shortcut label like we do in Terminal

jeremypw:
@cassidyjames How's this? I expect some tweaking of margins etc will be required, but it basically works.

cassidyjames:
I think we'd want to retain the menu styling, whereas this adds some padding and doesn't look like other menus.

image

image

jeremypw:
OK, so the main differences to be resolved are:
* spacing of menuitems
* Font of menuite, text
* Select on hover

jeremypw:
Appearance similar now:

# PR
accel_app_menu

# MASTER
master_app_menu

jeremypw:
Should be able to use Granite.AccelLabel to simplify this?

jeremypw:
@cassidyjames  Is this PR still needed?  There are what look like Granite.AccelLabels in that menu now:

Screenshot from 2020-03-13 19 06 14

cassidyjames:
Yeah this can be closed. Thanks for following up!

Issue #509, opened 2018-09-21
2nd time booting up Code after starting a new document doesn't show Welcome Screen


D0J0P:
And all tabs are closed as well, so I got confused when I first opened up Code and nothing showed up. I had to click on the gear icon, and select "Add New View" to open a new tab. 
![screenshot from 2018-09-21 14 35 27](https://user-images.githubusercontent.com/19495660/45899784-d24b2f00-bdab-11e8-9b69-61fbb2554f9c.png)

PR #527, opened 2018-10-13
Saving improvements

### LINKED ISSUE #381: confusing behaviour when using the "save as" on new document
Steps to reproduce:
1. create new document
2. use the "save as" option -> the tab changes to filename
3. further edit the file
4. closing the tab

By closing the tab the edits in point 3 are discarded. Code doesn't warn you.
I expected the editor to save my file like it does when opening files on the fly without manually saving it.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/56538641-confusing-behaviour-when-using-the-save-as-on-new-document?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #524: New unsaved document not restored
To reproduce:
1. Create new document and add some text
2. Close the app
3. Open it back up - no document, just welcome screen

You are not asked to save the file before exit and it's not shown when opening back Code. That's not expected behavior and can lead to data loss.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/64648782-new-unsaved-document-not-restored?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #381 
Fixes #524 

This fixes some pretty nasty data loss issues, but will need testing for regressions. One case is when closing Code, data would sometimes be lost because we used async, non-blocking methods to save the file on close which often wouldn't complete before the app closed.

The second fix is around the handling of saving of new/temporary files. Because the signal handlers for `changed` and `focus_out` were connected in the `open` method, which isn't called for new files, that auto saving wasn't working too well.

Also, I've made it so that new/temporary files are restored into tabs when re-opening Code.


cassidyjames:
@davidmhewitt I'm not sure if this is a regression or just other weird behavior, but it looks like when closing the window everything auto saves as expected, but when closing tabs I get an unsaved changes dialog. Which just seems odd.

cassidyjames:
That said, this does appear to fix data loss issues. I will continue running it in the meantime to watch for regressions.

davidmhewitt:
Ah, I had noticed that one time. I put it down to the fact that autosave happens on a repeating timer, so I think that only happens if you close a tab quickly after editing a file.

Wasn't sure if that would be one of the cases where there would be data loss in the past or if it's a minor regression I've introduced. I'll see if I can fix it.

ChristopherNugent:
For rapidly closing tabs after changes, saving changes on a tab close would seem to be the simplest solution.

This may be an odd question, but is there a formal definition of when files should be saved?

davidmhewitt:
@donadigo It looks like it's already possible for that issue to happen without this branch (see https://github.com/elementary/code/issues/509).

I think I'd like to investigate that separately to this branch so we can keep this branch focused on making saving more reliable and then look at why Code doesn't reload properly afterwards.

davidmhewitt:
@tintou has suggested I look at using `Application.hold` and `Application.release` instead of a MainLoop. Marking as in progress until I get chance to test that.

jeremypw:
Merging latest branch has fixed issue with a blank notebook on start up after closing with empty new document.

PR #529, opened 2018-10-14
Current view fix

### LINKED ISSUE #7: Save file always saves the right-side view
How to reproduce:

1. Open Scratch
2. Open a new Document
3. Add a new View
4. Type "this should be saved" in the left view, and "this shouldn't be saved" in the right view.
5. Click on the left view.
6. Click the save button and select a file
7. Look at the file and it contains "this shouldn't be saved".

What should happen:

The content of the selected view should be saved and not always the right one.

Launchpad Details: [#LP1392031](https://bugs.launchpad.net/bugs/1392031) Raphael Isemann - 2014-11-12 19:46:24 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44841592-save-file-always-saves-the-right-side-view?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

markodp:
This fixes some actions like save and undo, that don't work reliably anymore after creating 2nd code view.

The issue is SplitView's get_current_view method that isn't properly detecting focused view (anymore?). Some actions used different way of getting focused view, that works correctly - and that is what i reused.

fixes #7


src/MainWindow.vala
```diff
@@ -458,15 +458,7 @@ namespace Scratch {
...
-            Scratch.Widgets.DocumentView? view = null;
-
-            view = split_view.get_current_view ();
-
-            if (view == null && !split_view.is_empty ()) {
-                view = (split_view.get_child1 () ?? split_view.get_child2 ()) as Scratch.Widgets.DocumentView;
-            }
-
-            return view;
+            return split_view.get_focus_child () as Scratch.Widgets.DocumentView;
```
jeremypw: It is recommended to use a cast rather than the ` ... as ...` format.

PR #545, opened 2018-10-21
Accept C-PgUp, C-PgDown for moving between tabs

jmc-88:
It bugs me that I can't cycle between tabs with C-PgUp and C-PgDown, so here's a fix.


jeremypw:
It is recommended you make a new branch from your fork of master and make you changes there.  You can then keep your master branch in sync with elementary:master and make other branches for other fixes.

jeremypw:
This conflicts with the existing behaviour for Ctrl-PageUp and Ctrl-PageDown, which is to move to the beginning/end of the current line.

src/MainWindow.vala
```diff
@@ -981,6 +989,18 @@ namespace Scratch {
 
             project_pane.visible = !project_pane.visible;
         }
+
+        private void action_focus_next_document () {
+            Scratch.Widgets.DocumentView? view =
```
jeremypw: You can just use `var view = ` here as the type is indicated by the rhs.

src/MainWindow.vala
```diff
@@ -981,6 +989,18 @@ namespace Scratch {
...
+                split_view.get_focus_child () as Scratch.Widgets.DocumentView;
+            view.next_document ();
```
jeremypw: You have not dealt with the possibility of a null here.

src/MainWindow.vala
```diff
@@ -981,6 +989,18 @@ namespace Scratch {
...
+        }
+
+        private void action_focus_previous_document () {
+            Scratch.Widgets.DocumentView? view =
+                split_view.get_focus_child () as Scratch.Widgets.DocumentView;
+            view.previous_document ();
+        }
```
jeremypw: See comments above for related function.

jeremypw:
It is best to raise this kind of thing as an issue before coding to see if there is support/design team approval.

jmc-88:
> This conflicts with the existing behaviour for `<Ctrl><PageUp>` and , which is to move to the beginning/end of the current line. Code uses `<Ctrl><Alt><PageUp>` and `<Ctrl><Alt><PageUp>` to switch tabs (like gedit).

I would've swapped the two, but good to know it's already supported. Given this, I suppose this change is not needed. Thanks anyway for having a look at it.

jeremypw:
Hmm, having said that, I cannot get <Control><Alt><PageUp>/<Control><Alt><PageDown> to work now :-( - not sure what changed.  So there is work to be done to make behaviour of Code, Terminal, Files etc consistent in this area.

Issue #568, opened 2018-11-02
Sort open project folders


cassidyjames:
I often end up with a ton of project folders open, listed in the sidebar in whatever order they happened to be open. It could be useful to toggle an alphabetic sort or something to keep them more tidy.

![image](https://user-images.githubusercontent.com/611168/47933208-85be3d80-de99-11e8-8929-1d9c7c408274.png)

If we add an inline toolbar (#554), this could be added as a toggle at the end of the toolbar.

igordsm:
Isn't this solved by the current "Alphabetize" option? If not, could you clarify the difference?

jeremypw:
The current "Alphabetize" option does sort the projects but only on a one-off basis - subsequent additions are added at the bottom again.  I think a toggle option is requested so that the projects are always kept in order.

PR #569, opened 2018-11-02
Add inline toolbar to sidebar

### LINKED ISSUE #554: Move "Open folder" action to inline toolbar
Now that the project sidebar is in the core, would it make sense to add an inline toolbar at the bottom of the sidebar with an add/remove icon for adding/removing project folders? 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65345623-move-open-folder-action-to-inline-toolbar?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #656: Expand/collapse all projects in sidebar
I think this would make sense to live in an inline toolbar (#554) in the sidebar. If you're working with a bunch of projects, it would be handy to be able to collapse them all down at once to help you start organized. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/71423129-expand-collapse-all-projects-in-sidebar?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Adds an inline toolbar to the sidebar with an open folder button. 

Fixes #554 
Fixes #656

- [ ] Probably move toolbar to the project pane so it animates along with the stack switch
- [x] ~~Symbolic icons~~ https://github.com/elementary/stylesheet/pull/563
- [x] ~~Better icon for "Remove folder"?~~ _No "remove folder" action for now_


jeremypw:
The simple folder icon does not really indicate its purpose - need to see tooltip.  Any reason not to use the usual + and - icons?

cassidyjames:
@jeremypw my reasoning was to use a semantic icon for explicitly opening a folder, instead of a generic "add" icon—especially since the contents of the sidebar are mixed (and not just folders). 

I also worry about a "minus" icon being construed as "delete" a folder, instead of just removing it from the project view. I think for removing in that case, a context menu still might be best since the concept of which project folder is focused is also harder to see at a glance, and it might appear to be acting on a selected file instead.

jeremypw:
I do not think the minus icon is likely to be construed as actual deletion in the context of a sidebar as it does not mean this anywhere else (I think).  I agree that special "folder-add" and "folder-remove" icons, with tooltips, would be preferable.

roymckenzie:
Was a bit curious about this. It seems it duplicates functionality in the `HeaderBar`? Is this to help de-clutter the `HeaderBar`?

cassidyjames:
@roymckenzie the idea was to declutter the header bar a bit, to more clearly associate the folder/project actions with the sidebar, and to support further actions like in #568. I'm not sure if it's more discoverable or not, but I do like being able to add more folder management options there quite a bit—and those actions would feel awkward on their own.

roymckenzie:
@cassidyjames OK cool. That makes sense. I think an "Add File" button might be cool, too. And a "Collapse All Folders" button. My only complaint is that the toolbar looks so floaty and lonely down there with just the one button.

cassidyjames:
@roymckenzie yeah with just one icon it's kind of silly. But I imagine both an add/remove on the left, then perhaps a gear with a submenu for things like sorting/collapsing folders on the right.

PR #582, opened 2018-11-10
Fix empty files creation for previous opened files

### LINKED ISSUE #273: Wrong behaviour when trying to open nonexistant file
Steps to reproduce:

1. Open a file in Code
2. Close Code
3. Delete the file
4. Open Code
5. It tries to open the file but fails.

The errors:

1. **Don't show the error message (or make it more sensible)**

error

Either say something like "The file FOO doesn't exist anymore", which summarizes the error better than the current error, or just **fail silently (close the tab)** -- this would be preferred I think.

2. **The file gets created again!**

When Code doesn't find the file and shows the error, it creates the file again. It probably shouldn't?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/51482280-wrong-behaviour-when-trying-to-open-nonexistant-file?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #550: Zombie files created for restored tabs on startup
If a user edits a file in code, exits code, then deletes that file, an empty file with the same name will be created the next time code opens. 

I think the underlying reason for this is that code creates an empty file when told to open with a filename that does not exist, which may also the underlying cause of [issue 544](https://github.com/elementary/code/issues/544).

Code needs to be able to distinguish being told to _create a new file_ from _restoring a tab for a missing file_. If a tab is going to be restored for a file that no longer exists, code should neither create that file nor open that tab. 

I found this very confusing at first, not knowing why my deleted files were coming back, and without their ~souls~ content to boot, until I noticed the empty tabs still open in code.

~braaaaaiins~
coooooooode

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65175027-zombie-files-created-for-restored-tabs-on-startup?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mohelm97:
Fixes #550 
Fixes #273 

When Code restores your last session and there's a file deleted, it creates a zero byte file,
If the file isn't exits anymore don't open it and refresh the view if there's no files opened (To restore the welcome screen)

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -382,14 +382,16 @@ namespace Scratch {
 
             if (uris_view1.length > 0) {
                 var view = add_view ();
-                if(!load_files_for_view (view, uris_view1, focused_document1))
-                    split_view.remove_view(view);
+                if (!load_files_for_view (view, uris_view1, focused_document1)) {
+                    split_view.remove_view (view);
+                }
             }
 
             if (uris_view2.length > 0) {
                 var view = add_view ();
-                if(!load_files_for_view (view, uris_view2, focused_document2))
-                    split_view.remove_view(view);
+                if (!load_files_for_view (view, uris_view2, focused_document2)) {
+                    split_view.remove_view (view);
+                }
             }
         }
 
@@ -404,9 +406,9 @@ namespace Scratch {
                         file = File.new_for_commandline_arg (uri);
                     }
                     /* Leave it to doc to handle problematic files properly 
-                       But for files that doesn't exites we need to make sure that doc won't create a new file
+                       But for files that do not exist we need to make sure that doc won't create a new file
                     */
-                    if(file.query_exists()){
+                    if (file.query_exists ()) {
                         anyfile_loaded = true;
                         var doc = new Scratch.Services.Document (actions, file);
                         if (!doc.is_file_temporary) {
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -381,16 +381,19 @@ namespace Scratch {
 
             if (uris_view1.length > 0) {
                 var view = add_view ();
-                load_files_for_view (view, uris_view1, focused_document1);
+                if(!load_files_for_view (view, uris_view1, focused_document1))
+                    split_view.remove_view(view);
             }
 
             if (uris_view2.length > 0) {
                 var view = add_view ();
-                load_files_for_view (view, uris_view2, focused_document2);
+                if(!load_files_for_view (view, uris_view2, focused_document2))
+                    split_view.remove_view(view);
             }
         }
 
-        private void load_files_for_view (Scratch.Widgets.DocumentView view, string[] uris, string focused_document) {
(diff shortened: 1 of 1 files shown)


snowparrot:
Great! It solves some problems with Git too!

src/MainWindow.vala
```diff
@@ -382,16 +382,19 @@ namespace Scratch {
...
-                load_files_for_view (view, uris_view1, focused_document1);
+                if(!load_files_for_view (view, uris_view1, focused_document1))
+                    split_view.remove_view(view);
             }
 
             if (uris_view2.length > 0) {
                 var view = add_view ();
-                load_files_for_view (view, uris_view2, focused_document2);
+                if(!load_files_for_view (view, uris_view2, focused_document2))
+                    split_view.remove_view(view);
```
jeremypw: Elementary code style: Missing space before bracket.

src/MainWindow.vala
```diff
@@ -400,13 +403,19 @@ namespace Scratch {
                     } else {
                         file = File.new_for_commandline_arg (uri);
                     }
-                    /* Leave it to doc to handle problematic files properly */
-                    var doc = new Scratch.Services.Document (actions, file);
-                    if (!doc.is_file_temporary) {
-                        open_document (doc, view, file.get_uri () == focused_document);
+                    /* Leave it to doc to handle problematic files properly 
+                       But for files that doesn't exites we need to make sure that doc won't create a new file
```
jeremypw: Spelling/grammar: `doesn't exites`  =>  `do not exist`

jeremypw:
This is an improvement on the existing behaviour.  It would be even better if the user could be informed why the file has not been restored and given the option to recreate it - preferably in the Welcome view.  However, that can be left to another PR.

jeremypw:
I have updated the description to link this PR with a related issue that it addresses.  Merging this PR will then automatically close that issue.

src/MainWindow.vala
```diff
@@ -382,16 +382,19 @@ namespace Scratch {
 
             if (uris_view1.length > 0) {
                 var view = add_view ();
-                load_files_for_view (view, uris_view1, focused_document1);
+                if(!load_files_for_view (view, uris_view1, focused_document1))
```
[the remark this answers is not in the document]
Mohelm97: @jeremypw Thanks a lot for your guidance, everything is edited according to the Elementary code style.

Mohelm97:
@jeremypw I did all the changes, Happy to hear your review again ;)

PR #591, opened 2018-11-18
Outline: Modernize the C backend and make it more efficient

### LINKED ISSUE #583: Code doesn't start if ctags fails
This can happen when trying to open a file that for some reason appears to have the mime type of a C file but isn't actually a C file, which I've encountered before.

tintou:
Fixes #583

### DIFF
--- a/po/extra/ja.po
+++ b/po/extra/ja.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: scratch\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2018-10-18 14:13-0700\n"
-"PO-Revision-Date: 2018-11-04 00:03+0000\n"
+"PO-Revision-Date: 2018-11-21 00:03+0000\n"
 "Last-Translator: Ryo Nakano <ryonakaknock3@gmail.com>\n"
 "Language-Team: Japanese <https://weblate.elementary.io/projects/code/extra/"
 "ja/>\n"
@@ -79,7 +79,7 @@ msgstr ""
 
 #: data/io.elementary.code.appdata.xml.in:38
 msgid "Additional features include:"
-msgstr "その他の含まれる機能:"
+msgstr "その他の含まれる機能："
 
 #: data/io.elementary.code.appdata.xml.in:42
 msgid "syntax highlighting with gtksourceview-3"
--- a/po/ja.po
+++ b/po/ja.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: scratch\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2018-10-18 14:12-0700\n"
-"PO-Revision-Date: 2018-11-04 00:03+0000\n"
+"PO-Revision-Date: 2018-11-21 00:03+0000\n"
 "Last-Translator: Ryo Nakano <ryonakaknock3@gmail.com>\n"
 "Language-Team: Japanese <https://weblate.elementary.io/projects/code/code/ja/"
 ">\n"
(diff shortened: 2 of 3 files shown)


plugins/outline/C/CtagsSymbolOutline.vala
```diff
@@ -197,23 +234,9 @@ public class Code.Plugins.CtagsSymbolOutline : Object, Code.Plugins.SymbolOutlin
...
-            parent = fields.substring (index + 6, int.max (fields.index_of (" ", index + 7) - index, -1));
-        }
-        if ((index = fields.index_of ("struct:")) > -1) {
-            parent = fields.substring (index + 7, int.max (fields.index_of (" ", index + 7) - index, -1));
-        }
-        if ((index = fields.index_of ("enum:")) > -1) {
-            parent = fields.substring (index + 5, int.max (fields.index_of (" ", index + 7) - index, -1));
-        }
-    }
+    /*void parse_fields (string fields, out int line, out string? parent, out string? parent_type) {
```
davidmhewitt: Can remove this commented section now?

plugins/outline/C/CtagsSymbolOutline.vala
```diff
@@ -233,14 +256,24 @@ public class Code.Plugins.CtagsSymbolOutline : Object, Code.Plugins.SymbolOutlin
     }
 
     public static string[] get_supported_types () {
```
davidmhewitt: This method doesn't seem to be used anywhere. Can probably remove it?

plugins/outline/C/CtagsSymbolOutline.vala
```diff
@@ -233,14 +256,24 @@ public class Code.Plugins.CtagsSymbolOutline : Object, Code.Plugins.SymbolOutlin
...
+            var subprocess = new GLib.Subprocess (
+                GLib.SubprocessFlags.STDOUT_PIPE|GLib.SubprocessFlags.STDERR_SILENCE,
+                "ctags", "--list-languages"
+            );
+
+            var datainput = new GLib.DataInputStream (subprocess.get_stdout_pipe ());
+            string lang = null;
+            while ((lang = datainput.read_line ()) != null) {
+                langs += lang;
+                warning (lang);
```
davidmhewitt: If this method does still need to be here, please turn this into a debug message instead of a warning.

plugins/outline/C/CtagsSymbolOutline.vala
```diff
@@ -76,9 +79,23 @@ public class Code.Plugins.CtagsSymbolOutline : Object, Code.Plugins.SymbolOutlin
                 // 1 => filename
```
davidmhewitt: I think I'd like some bounds checking here? There's a lot of array access at index `3` and `4` when it's possible (but unlikely) that `ctags` has output a line without these fields?

plugins/outline/C/CtagsSymbolOutline.vala
```diff
@@ -123,57 +140,77 @@ public class Code.Plugins.CtagsSymbolOutline : Object, Code.Plugins.SymbolOutlin
...
+                                case "enum":
+                                    parent.icon = new ThemedIcon ("lang-enum");
+                                    break;
+                            }
                             iter.remove ();
+                            continue;
                         }
                     }
+                    // anonymous enum
+                    if (i.parent.substring (0, 6) == "__anon") {
```
davidmhewitt: Another small thing, can we go with `has_prefix` here?

plugins/outline/C/CtagsSymbolOutline.vala
tintou: I've added a check that ignores all unexpected short lines

Issue #473, opened 2018-07-08
Code doesn't respect "Privacy → History" setting


uncertainquark:
Unlike Scratch, Code always remembers the last open files even if the "Privacy > History" setting is off. Moreover, the option to turn off remembering last open files, which Scratch had, is missing.
![screenshot from 2018-07-08 21-00-28](https://user-images.githubusercontent.com/22585732/42421385-6d39c26e-82f2-11e8-902b-956fa5abb09e.png)

cassidyjames:
I've clarified the scope of this issue a bit; apps should respect the system-wide setting instead of implementing their own.

PR #610, opened 2018-12-19
[Request]Implement a "Restore tabs" option (default true)

### LINKED ISSUE #605: [request] Add option to open code in welcome screen rather than last opened files.
codewelcome

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67599915-request-add-option-to-open-code-in-welcome-screen-rather-than-last-opened-files?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #605 

Adds another option to the Preferences dialog that turns of the default behaviour to open previously open documents when the application is launched and to show the welcome screen instead

### DIFF
--- a/data/io.elementary.code.gschema.xml
+++ b/data/io.elementary.code.gschema.xml
@@ -74,6 +74,11 @@
       <summary>Wrap long lines</summary>
       <description>Wrap long lines.</description>
     </key>
+    <key name="restore-tabs" type="b">
+      <default>true</default>
+      <summary>Whether to restore tabs when app is opened</summary>
+      <description>Whether to restore previously opened tabs or to show the welcome screen when the application is opened</description>
+    </key>
     <key name="show-right-margin" type="b">
       <default>false</default>
       <summary>Show the right margin</summary>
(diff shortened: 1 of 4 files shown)


mustafaREIMAHG:
restore-tabs

cassidyjames:
@jeremypw as mentioned in #473, Code should instead follow the system-wide history setting.

jeremypw:
I am not sure these are the same thing.  The history setting is about overall privacy whereas restoring tabs is a per application choice?  Files has its own setting (although not exposed in UI)

cassidyjames:
@jeremypw right, but apps should not have per-app choices because that goes against the design of the OS and introduces inconsistencies. Apps in elementary OS save their state when closed, and as much as possible closing should be similar in experience to minimizing. Introducing per-app settings for this is passing the design of the OS off to users.

To alleviate the privacy concerns and centralize settings, we have the system-wide history setting that apps should respect.

jeremypw:
OK, I'll close this then and do it the other way.  I'll also have to modify Files to comply.

PR #611, opened 2018-12-19
Try an info icon for smart cut/copy

cassidyjames:
An example of something we could do for #425. Suggestions welcome.

screenshot from 2018-12-19 14 44 06


jeremypw:
Looks good to me :-)

PR #612, opened 2018-12-21
Honor global history setting

### LINKED ISSUE #473: Code doesn't respect "Privacy → History" setting

jeremypw:
Fixes #473 

When history setting is turned off, Code does not save current documents (and clears those settings if it is open at the time).

When history setting is off, Code does not restore previously open documents when it is launched.


jeremypw:
Will revert clear settings when history turned on following comments to similar Files PR.

jeremypw:
The trend is to replace Granite Settings class with GLib.Settings so rather than introduce a new Granite based class I will rework this to use GLib.Settings.

PR #613, opened 2018-12-21
Fixes to the new window desktop action

### LINKED ISSUE #564: Code Open new Window Bug
When i right click code editor and select the option "New Window", code starts up with two windows and an error.
Here is a video, that shows the bug: https://youtu.be/r1mkAIooydE

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65597885-code-open-new-window-bug?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #564 

Don't create the window before we check whether a window already exists or not and ignore null arguments.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -121,11 +121,7 @@ namespace Scratch {
                 File[] files = new File[unclaimed_args];
                 files.length = 0;
 
-                foreach (string arg in args[1:unclaimed_args + 1]) {
-                    if (arg == null) {
-                        continue;
-                    }
-
+                foreach (unowned string arg in args[1:unclaimed_args]) {
                     // We set a message, that later is informed to the user
                     // in a dialog if something noteworthy happens.
                     string msg = "";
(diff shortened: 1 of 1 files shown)


src/Application.vala
```diff
@@ -122,6 +122,10 @@ namespace Scratch {
                 files.length = 0;
 
                 foreach (string arg in args[1:unclaimed_args + 1]) {
```
tintou:
```suggestion
                foreach (unowned string arg in args[1:unclaimed_args]) {
```
Wouldn't that solve the `NULL` issue (as the `args` string array is NULL-terminated)

davidmhewitt: Yes, it would. I hadn't noticed there was an `unclaimed_args + 1` there so hadn't questioned where there was a null in the array.

PR #632, opened 2019-01-28
Implement terminal zoom with keyboard.

### LINKED ISSUE #304: Text zooming should also affect terminal plug
Probably best to be consistent.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/53673659-text-zooming-should-also-affect-terminal-plug?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #304 

Action accelerators could not be used (afaict) as they would clash with the same ones for document zooming.  So the hotkeys were handled by a key press handler in the terminal plug.  Zooming functions were called directly rather than via actions but actions could be added if the terminal plug gains a toolbar in future.

This PR does not save and restore the  terminal zoom value, but a terminal plug setting could be added in a future PR.


jeremypw:
It is assumed that the terminal font size should be independent from the document.

Issue #614, opened 2018-12-31
Vim emulation does not parse keystroke "a" 


mcclurgm:
The vim emulation plugin does not appear to account for the vim `a` (insert after cursor position) command. This is often very useful, giving users more control over where the cursor goes with fewer keystrokes.

I looked at the code in [vim-emulation.vala](https://github.com/elementary/code/blob/master/plugins/vim-emulation/vim-emulation.vala) and it appears the `a` keycode is not included in the switch statements. The current code includes cases for keys `i` and capital `A` and `I` as ways of entering insert mode. A similar case could be included to handle this keystroke.

clobrano:
As a Neovim user, I'd like to look into this.

mcclurgm:
I also have an PR that implements this, as well as some other functionality. It's been waiting in review: #633. Not sure what we should do about the duplicate implementations then.

clobrano:
Oh sorry, I didn't noticed. Yours look more complete, mine can be discarded

mcclurgm:
Perhaps, but your implementation cleans the action string, which is an improvement over mine. If you do want to discard yours, then we should implement it in mine too.

I should have commented here about my PR, so sorry for that.

clobrano:
On the other hand, I believe that merging this one, would just require a small rebase for your's to be valid as well.

PR #633, opened 2019-02-01
Enhance vim emulation commands

mcclurgm:
Started as a fix for my issue #614.
It implements 2 new commands, lower-case "a" to insert after cursor and caret "^" to move to beginning of the indented line.
It also updates the behavior of other commands to use paragraph ends instead of display line ends, like vim does, to be more predictable.

It's a work in progress at the moment, I expect to add more stuff going forward. All my changes should be independent so you can only merge the ones you want.


jeremypw:
Probably best reviewed by a Vim (plugin) user. Code style looks OK.

jeremypw:
@mcclurgm Need to fix lint error (trailing whitespace).

mcclurgm:
Sorry I haven't been keeping up on this. I'm working on starting to break it up into a collection of smaller, focused PRs as you requested. I've also realized issues with some of the changes in this PR (for example, moving up or down resets to the end of the line), so I think it's probably safe to discard it and I'll reimplement the changes with fixed behavior.

jeremypw:
@mcclurgm OK, thanks.  I'll close for now.

PR #636, opened 2019-02-14
meson: Drop superfluous and erroneous install_dir usage

### LINKED ISSUE #635: Meson build failed for armhf
sudo meson build --prefix=/usr

**Output:**
Meson encountered an error in file src/meson.build, line 53, column 0:
Custom_install_dir must be a string

ricotz:
Fixes #635


ricotz:
Ok, I assumed `install: true` would implicitly set the appropriate `install_dir` values

PR #637, opened 2019-02-15
Spawn terminal in current working dir by default

### LINKED ISSUE #58: Terminal extention open in current directory.
I think it makes sense for the terminal extension to open the terminal in the current working directory. If not by default than with an option or a extra click.

jmc-88:
Fixes #58.


plugins/terminal/terminal.vala
```diff
@@ -215,7 +215,10 @@ public class Scratch.Plugins.Terminal : Peas.ExtensionBase,  Peas.Activatable {
         });
 
         try {
-            string last_opened_path = settings.last_opened_path == "" ? "~/" : settings.last_opened_path;
+            string last_opened_path =
+                settings.last_opened_path == ""
+                ? GLib.Environment.get_current_dir ()
+                : settings.last_opened_path;
```
jeremypw: If you need to spread a "single line" if/else clause over 4 lines, it is probably clearer to use a multi-line form:
```
string last_opened_path = settings.last_opened_path;
if (last_opened_path == "") {
    last_opened_path = GLib.Environment.get_current_dir ();
}

jmc-88: Done.

Slightly off-topic, but since it came up:

- Is there a coding style that is required to follow for official elementary projects? This would reduce review time and improve consistency of the code base.
- Better yet, is there a linter/formatter for Vala that would reformat the source to conform to such a style guide and remove inconsistencies at the root? See gofmt for a good example.

jmc-88:
> The problem with this is that it only has effect on the first use of the plugin.

That's a good point. Actually, this seems to be initialized when the plugin is loaded, and then never again. I guess the plugin should be reworked a little to accommodate this use case.

Given this, let me know if you wish to proceed with the review in its current form anyway, or if I should drop it and attack it from another angle.

plugins/terminal/terminal.vala
jeremypw: The elementary code style can be found here `https://elementary.io/docs/code/reference#reference`

There is the `https://github.com/elementary/vala-lint` project (which you need to build yourself). This points out code style problems but does not autocorrect them.

jeremypw:
I have pushed some commits to allow the terminal to follow the current document directory (subject to a setting) or otherwise use the environment current directory (updated).  I am not entirely sure what the desired behaviour is so I'll flag @elementary/ux to comment.

jeremypw:
There is an argument that the terminal should open at the root of the current project directory (and not follow the document path).

Geany has an option to follow or not follow the current document path.  It also has a context menu option to set the terminal path to the current document's folder.

jeremypw:
Latest commit implements the possibility for terminal to follow the current project root folder, unless the "follow-current-folder" setting is true.  This required making some previously `internal` FolderManager methods `public`.

jeremypw:
Need to improve efficiency for large projects.

jeremypw:
Apart from an appreciable delay (a few seconds) in initially opening a large repository (e.g. LibreOffice) in the FolderManagerView (unrelated to this PR) during which the view is frozen, this PR works well with a large repo.  Note that the terminal only switches to the project root directory if the folder containing the document is expanded in the FolderManagerView.  This is an inevitable consequence of the way the FMV works (only loads children when folder is expanded).

jeremypw:
It might be better to for the terminal to only show the root path when inside a git folder? At the moment this PR works for any folder opened as a root in the FolderManagerView.

jeremypw:
@jmc-88 Thinking now is to retire the terminal plugin in favour of a more focused ability to build/install so it is probably not worth working on this any more.

jmc-88:
> @jmc-88 Thinking now is to retire the terminal plugin in favour of a more focused ability to build/install so it is probably not worth working on this any more.

Got it, I'll close this then.

PR #640, opened 2019-02-16
Utils: Use FileChooserNative

danrabbit:
Fixes #639

PR #643, opened 2019-02-26
Utils: Remove filechooser util

### LINKED ISSUE #639: Use FileChooserNative
Because Portals and future.

- [x] https://github.com/elementary/code/blob/e0f239f11ed928907e0661f6a39727447803a4eb/src/Utils.vala#L25
- [x] https://github.com/elementary/code/blob/4cc4b6c7c07391175e7ad426d1ef8b9978bd3a2c/src/MainWindow.vala#L761

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70105250-use-filechoosernative?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Supersedes #640 
Fixes #639 

This util is only used twice so this seems like an unnecessary over abstraction. It's simpler and clearer to just make two dialogs.

Fwiw, this solution adds 2 lines as opposed to #640 which adds 8

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -741,7 +741,7 @@ namespace Scratch {
 
         private void action_open () {
             // Show a GtkFileChooserDialog
-            var filech = Utils.new_file_chooser_dialog (Gtk.FileChooserAction.OPEN, _("Open some files"), this, true);
+            var filech = Utils.new_file_chooser_dialog (Gtk.FileChooserAction.OPEN, _("Open some files"), this);
             var response = filech.run ();
             filech.destroy (); // Close now so it does not stay open during lengthy or failed loading
 
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -740,13 +740,32 @@ namespace Scratch {
         }
 
         private void action_open () {
-            // Show a GtkFileChooserDialog
-            var filech = Utils.new_file_chooser_dialog (Gtk.FileChooserAction.OPEN, _("Open some files"), this);
-            var response = filech.run ();
-            filech.destroy (); // Close now so it does not stay open during lengthy or failed loading
+            var all_files_filter = new Gtk.FileFilter ();
+            all_files_filter.set_filter_name (_("All files"));
+            all_files_filter.add_pattern ("*");
+
+            var text_files_filter = new Gtk.FileFilter ();
+            text_files_filter.set_filter_name (_("Text files"));
+            text_files_filter.add_mime_type ("text/*");
+
+            var file_chooser = new Gtk.FileChooserNative (
+                _("Open some files"),
+                this,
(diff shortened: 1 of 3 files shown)


src/Services/Document.vala
```diff
@@ -506,18 +506,34 @@ namespace Scratch.Services {
...
+            file_chooser.set_current_folder_uri (Utils.last_path ?? GLib.Environment.get_home_dir ());
 
             var success = false;
             var current_file = file.get_path ();
             var is_current_file_temporary = this.is_file_temporary;
 
-            if (filech.run () == Gtk.ResponseType.ACCEPT) {
-                this.file = File.new_for_uri (filech.get_file ().get_uri ());
+            if (file_chooser.run () == Gtk.ResponseType.ACCEPT) {
+                this.file = File.new_for_uri (file_chooser.get_file ().get_uri ());
```
jeremypw: I think file_chooser.get_uri () is better for the portal filechooser.  I am not sure why filech.get_file ().get_uri () was originally used here.

src/MainWindow.vala
```diff
@@ -740,13 +740,32 @@ namespace Scratch {
...
+            var file_chooser = new Gtk.FileChooserNative (
+                _("Open some files"),
+                this,
+                Gtk.FileChooserAction.OPEN,
+                _("Open"),
+                _("Cancel")
+            );
+            file_chooser.add_filter (all_files_filter);
+            file_chooser.add_filter (text_files_filter);
+            file_chooser.filter = text_files_filter;
```
jeremypw: It seems that the first added filter ends up as the "set" filter regardless of anything else.  Not sure if this is a result of using FileChooserNative.  Anyway this is sufficient:

```
            file_chooser.add_filter (text_files_filter);           
            file_chooser.add_filter (all_files_filter);
```

src/Services/Document.vala
```diff
@@ -506,18 +506,34 @@ namespace Scratch.Services {
...
+
+            var file_chooser = new Gtk.FileChooserNative (
+                _("Save File"),
+                null,
+                Gtk.FileChooserAction.SAVE,
+                _("Save"),
+                _("Cancel")
+            );
+            file_chooser.add_filter (all_files_filter);
+            file_chooser.add_filter (text_files_filter);
```
jeremypw: As mentioned above, need to add the text filter first to make it the default.

danrabbit: I think it was intentional that the all files filter be default for saving

jeremypw: OK

PR #650, opened 2019-03-10
Only open empty file after name is set

### LINKED ISSUE #213: Creating new file immediately opens it
When creating a new file in the sidebar, Scratch shouldn't immediately open it in the editor with the "unnamed" name, but allow for renaming it, and after the renaming is done, the file should be loaded in the editor.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49059093-creating-new-file-immediately-opens-it?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

roymckenzie:
Hello! I have a fix in for #213.

It involves limiting when the `FileView` is being refreshed and when it is selecting a specific `Item`/`FolderItem`.

The new functionality will allow you to create a new Empty File and rename it and automatically select it for editing only after the name has been set.


src/FolderManager/FileView.vala
```diff
@@ -51,7 +51,7 @@ namespace Scratch.FolderManager {
                 return;
             }
 
-            if (item is FileItem) {
+            if (item is FileItem && !(item as FileItem).just_created) {
```
donadigo: You should set `view.ignore_next_select` to true in the FolderItem instead to avoid selecting it here.

src/FolderManager/Item.vala
```diff
@@ -42,6 +44,10 @@ namespace Scratch.FolderManager {
 
         protected void rename (string new_name) {
             file.rename (new_name);
+            if (this is FileItem) {
+                view.select(file.file.get_parent ().get_path () + "/" + new_name);
```
donadigo: Should just use `file.file.get_path ()`.

src/FolderManager/Item.vala
```diff
@@ -25,6 +25,8 @@ namespace Scratch.FolderManager {
     internal abstract class Item: Granite.Widgets.SourceList.ExpandableItem, Granite.Widgets.SourceListSortable {
         public File file { get; construct; }
 
+        public bool just_created = false;
```
donadigo: This isn't needed since `FileView` already has the `ignore_next_select` property.

src/FolderManager/Item.vala
roymckenzie: `file.file.get_path ()` returns the path of the file including the file name. I just want the file name's directory so I can append the new name to it.

src/FolderManager/Item.vala
roymckenzie: Fixed this in latest PR. Much cleaner and more manageable to use `ignore_next_select`. Thanks!

src/FolderManager/FileView.vala
roymckenzie: Done. Thank you!

roymckenzie:
@donadigo I believe I resolved all the items you mentioned. There was [one comment you made](https://github.com/elementary/code/pull/650#discussion_r264483092) in particular that I'd like you to [look at](https://github.com/elementary/code/pull/650#discussion_r264520931). Just wanted to confirm that the way I get the path for the new file is the best way to do it. Let me know if you know of a better way. Thank you!

src/FolderManager/Item.vala
donadigo: I didn't test this yet, but if you renamed the file, doesn't that guarantee that you already have the new name in the path?

Also, if you would want to do this in the more correct way it would be:
`Path.build_filename (file.file.get_parent ().get_path (), new_name)`

roymckenzie: I thought that would be the case, but renaming the file causes `FileView` to detect and create a new `Item` object as the `FileView` doesn't know that the file has been renamed, but instead that a new file has appeared in the directory.

I will use the Path builder and update the code. Thanks for that; I didn't know about that! I thought my way was a little sloppy, but this is much cleaner.

src/FolderManager/File.vala
```diff
@@ -139,7 +139,7 @@ namespace Scratch.FolderManager {
                     while ((file_info = enumerator.next_file ()) != null) {
                         var child = file.get_child (file_info.get_name ());
                         var file = new File (child.get_path ());
-                        _children.add (new File (child.get_path ()));
+                        _children.add ( file );
```
donadigo:
```suggestion
                        _children.add (file);
```

src/FolderManager/File.vala
```diff
@@ -176,6 +176,11 @@ namespace Scratch.FolderManager {
         }
 
         public void rename (string name) {
```
donadigo:
```suggestion
         public File rename (string name) {
```

src/FolderManager/File.vala
```diff
@@ -176,6 +176,11 @@ namespace Scratch.FolderManager {
...
+            // set_display_name throws a warning if the file already exists
+            if (this.name == name) {
+                return;
+            }
+
             try {
                 file.set_display_name (name);
```
donadigo:
```suggestion
                 return file.set_display_name (name);
```

src/FolderManager/File.vala
```diff
@@ -176,6 +176,11 @@ namespace Scratch.FolderManager {
...
             } catch (GLib.Error error) {
```
donadigo: You will need to add `return file` here.

src/FolderManager/Item.vala
```diff
@@ -42,6 +42,11 @@ namespace Scratch.FolderManager {
 
         protected void rename (string new_name) {
             file.rename (new_name);
```
donadigo:
```suggestion
             var new_file = file.rename (new_name);
```

src/FolderManager/Item.vala
```diff
@@ -42,6 +42,11 @@ namespace Scratch.FolderManager {
...
+            if (this is FileItem) {
+                var path = Path.build_filename(file.file.get_parent ().get_path (), new_name);
+                view.select(path);
```
donadigo:
```suggestion
                view.select (path);
```

cassidyjames:
@roymckenzie any chance to take another look at this? It would be a great fix to get in!

roymckenzie:
@cassidyjames I switched to Debian Buster and need to get my development environment setup again. I will try to do that this week. If someone else wants to jump in, feel free!

cassidyjames:
@roymckenzie it looks like elementary contributors don't have write access to your branch, maybe (I can't commit those code style fixes here). Is that something you can enable in the fork's repo settings?

roymckenzie:
@cassidyjames I haven't gone through all the comments that @donadigo made so I am hesitant to just commit his suggestions to resolve this PR. I think interested parties should be able to just clone this branch and submit their own pull request, correct?

cassidyjames:
@roymckenzie yeah, but I think there's a setting on the fork to allow elementary contributors (anyone with write access to the upstream repo) to write to this branch, so we can commit some of the minor stuff like whitespace fixes, and someone could come along and more easily continue this work here. But yeah, they can also open a separate PR.

roymckenzie:
@cassidyjames I looked in the settings for the fork and couldn't see anything that would allow that. I can merge the style suggestions.

jeremypw:
I am in the process of implementing donadigo's suggestions. The code is quite confusing as it does not distinguish clearly between Scratch.FolderManager.File and GLib.File.  But it would require another PR to fix this.

src/FolderManager/File.vala
jeremypw: I think you mean `public GLib.File rename (string name) {`

jeremypw:
@cassidyjames I just tried it again and it seems to be doing what is expected - that is the new file entry in the sidebar enters "edit" mode and stays there until "Enter" or "Esc" is pressed or the sidebar loses focus.  However, there is a bug when the sidebar is scrolled before the new file is created - sometimes the edit widget does not appear on top of the item being edited.  It may appear right at the top of the sidebar.  I am not sure this is an issue with this PR or with the sidebar itself.  I will investigate further.

jeremypw:
I notice Files also has trouble renaming a bookmark after scrolling down.  However in Files case, the sidebar scrolls back to the top and the edit-widget is then off the bottom, requiring re-scrolling to see it.

jeremypw:
The scroll issue seems to happen mainly with the following sequence:
* Open Code
* Expand a root folder with lots of subfolders
* Scroll down to see a subfolder that was out of window.
* Do *not* expand that subfolder but use "New File" option on it.

Result: The subfolder expands and shows "New File" item but the edit entry appears at the top of the sidebar.

jeremypw:
Latest commit seems to fix the problem.

jeremypw:
I have found this hangs under some circumstance (not tested master yet). Investigating.

jeremypw:
Also hangs in master.

jeremypw:
Happened when expanding a folder in the sidebar which contained about 1000 files so I guess this is a severe performance issue associated with sorting the entries rather than an infinite loop.

jeremypw:
@donadigo I cannot reproduce your issue when creating new folders and files in existing folders.  Are you able to submit a video of the effect?  A problem I have found is if you try to create a new file in a newly created folder.  The new folder does not have an expander and does not autoexpand when you create the new file.

jeremypw:
I am beginning to wonder whether it may be better to enter the name of the new file *before* creating it. Should result in fewer signals flying around and potential races.

jeremypw:
There is a serious difficulty in editing the name before creating the file - there is no "editing-canceled" signal emitted by SourceList :-( , only the signal "edited" - which is not emitted on cancellation.  The "editing-canceled" signal emitted by the cell renderer is not publicly accessible.   This means there is no way of detected when to remove the dummy item if editing is cancelled.

jeremypw:
An alternative fix for this has now be merged.  Thanks for the contribution, though.

PR #651, opened 2019-03-11
User can see all file types and open in default application. 

### LINKED ISSUE #449: Non-text files not displayed in folder manager
Files such as images or fonts aren't displayed in the folder tree, which I find slightly confusing. I realise this is intentional to stop you trying to open binary files in a text editor, but there are a couple of context menu items like "delete" and "open in" that would still be useful and prevent me having to open my file manager instead.

I guess there just needs to be a design decision as to whether or not to display them and how exactly they should be handled.

roymckenzie:
User can view all file types (files were previously limited to mime text files and folders) and open non-text files in OS default application. Resolves #449.


roymckenzie:
Thanks for the quick review @cassidyjames! I have addressed those two issues. Take a look when you have a moment.

PR #657, opened 2019-03-13
User can see close focused project folder, alphabetize folders, and collapse folders.

roymckenzie:
@cassidyjames I've updated this branch with some of the ideas we messaged about. Specifically:

1. The ability to close an active / selected Project Folder.
2. The ability to collapse all Project Folders.
3. The ability to order Project Folders alphabetically.

Let me know what you think.

Screenshot from 2019-03-13 03 36 05


tintou:
Wait, shouldn't the project folder always be alphabetized?

roymckenzie:
@tintou they didn't appear to be alphabetized by default. If that is something that should be done by default (probably), I can modify the `add_folder` method in `FileView` to do that and we can just remove it from the `Gtk.Menu` in the inline toolbar.

cassidyjames:
@tintou @roymckenzie to clarify, this is about sorting the projects themselves in alphabetical order, not the contents of the projects/folders. In this case, no, they should not always be alphabetized because they're in the order they were added; you know the last-added project will always be at the bottom. However, if you've amassed a large amount of projects you're jumping between a lot, it would be nice to sort them in alphabetical order.

roymckenzie:
Looking at the icon for the remove project folder action, I don't like the icon that much. It is hard to see and does not seem intuitive. 

I am currently using the `folder-move-symbolic` icon for that. 

A deeper question I have is: Does it even make since to have that button / functionality located there? The user can right-click on the project folder in the `FileView` to close a specific project folder. Actions on the newly proposed inline toolbar imply more general actions related to all project folders than one specific one in my opinion.

cassidyjames:
@roymckenzie that's fair. It's also not clear which specific project is "selected", so removing one could be problematic imho.

roymckenzie:
@cassidyjames I removed the "Close Active Folder" functionality from the inline toolbar (and this branch). Let me know what you think.

src/Widgets/Pane.vala
```diff
@@ -39,11 +39,33 @@ public class Code.Pane : Gtk.Grid {
...
-        add_button.tooltip_text = _("Add Project Folder…");
+        var add_folder_button = new Gtk.ToolButton (new Gtk.Image.from_icon_name ("folder-open-symbolic", Gtk.IconSize.BUTTON), null);
+        add_folder_button.action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER;
+        add_folder_button.tooltip_text = _("Add Project Folder…");
 
-        toolbar.add (add_button);
+        var project_more_button = new Gtk.MenuToolButton (null, null);
+        project_more_button.tooltip_text = _("More…");
+        
+        var collapse_all_menu_item = new Gtk.MenuItem.with_label (_("Collapse Project Folders"));
```
cassidyjames: With the other suggestions, this could be more concise:

```suggestion
        var collapse_all_menu_item = new Gtk.MenuItem.with_label (_("Collapse All"));
```

src/Widgets/Pane.vala
```diff
@@ -39,11 +39,33 @@ public class Code.Pane : Gtk.Grid {
...
+        collapse_all_menu_item.action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_COLLAPSE_ALL_FOLDERS;
+
+        var order_projects_menu_item = new Gtk.MenuItem.with_label (_("Alphabetize Project Folders"));
```
cassidyjames: With the other suggestions, this could be more concise:

```suggestion
        var order_projects_menu_item = new Gtk.MenuItem.with_label (_("Alphabetize"));
```

src/Widgets/Pane.vala
```diff
@@ -39,11 +39,33 @@ public class Code.Pane : Gtk.Grid {
...
-        var add_button = new Gtk.ToolButton (new Gtk.Image.from_icon_name ("folder-open-symbolic", Gtk.IconSize.BUTTON), null);
-        add_button.action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER;
-        add_button.tooltip_text = _("Add Project Folder…");
+        var add_folder_button = new Gtk.ToolButton (new Gtk.Image.from_icon_name ("folder-open-symbolic", Gtk.IconSize.BUTTON), null);
+        add_folder_button.action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER;
+        add_folder_button.tooltip_text = _("Add Project Folder…");
 
-        toolbar.add (add_button);
+        var project_more_button = new Gtk.MenuToolButton (null, null);
+        project_more_button.tooltip_text = _("More…");
```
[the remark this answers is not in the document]
roymckenzie: I like the `application-menu-symbolic`. I'll push it here in a bit.

jeremypw:
@roymckenzie  Will you be updating this PR? There seems to be an outstanding change request. If so, you might consider incorporating the functionality of #665 (if feasible).

cassidyjames:
Barring that symbolic issue, I'm personally very happy with this branch and have been using it for a while. I forgot this wasn't in master! :sweat_smile:

cassidyjames:
Ah, this should probably also remove the Open Folder button from the HeaderBar since it could be confusing that there are two places to do it. But I'd want to get thoughts from @elementary/ux.

cassidyjames:
@roymckenzie it turns out the icon issue is actually an issue in the stylesheet, so I've reported it there.

src/Widgets/Pane.vala
```diff
@@ -39,11 +39,32 @@ public class Code.Pane : Gtk.Grid {
...
-        add_button.action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER;
-        add_button.tooltip_text = _("Add Project Folder…");
+        var add_folder_button = new Gtk.ToolButton (new Gtk.Image.from_icon_name ("folder-open-symbolic", Gtk.IconSize.BUTTON), null);
+        add_folder_button.action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER;
+        add_folder_button.tooltip_text = _("Add Project Folder…");
 
-        toolbar.add (add_button);
+        var project_more_button = new Gtk.MenuToolButton (null, null);
+        project_more_button.tooltip_text = _("Manage project folders…");
+
```
jeremypw: Lint: Unexpected whitespace at end of line.

src/Widgets/Pane.vala
```diff
@@ -39,11 +39,32 @@ public class Code.Pane : Gtk.Grid {
         toolbar.get_style_context ().add_class (Gtk.STYLE_CLASS_INLINE_TOOLBAR);
         toolbar.icon_size = Gtk.IconSize.SMALL_TOOLBAR;
 
-        var add_button = new Gtk.ToolButton (new Gtk.Image.from_icon_name ("folder-open-symbolic", Gtk.IconSize.BUTTON), null);
-        add_button.action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER;
-        add_button.tooltip_text = _("Add Project Folder…");
+        var add_folder_button = new Gtk.ToolButton (new Gtk.Image.from_icon_name ("folder-open-symbolic", Gtk.IconSize.BUTTON), null);
```
jeremypw: Lint: Long line

jeremypw: Shouldn't we use either a "folder-add" or "list-add" symbolic icon?  The latter is more consistent with other elementary sidebar toolbars.  It is arguably implicit that we are adding folders since they are the root children of the sidebar but a tooltip could be added to confirm this.  I agree that in the context of the headerbar "folder-open" makes more sense but the toolbar is in the context of the sidebar.

src/Widgets/Pane.vala
```diff
@@ -39,11 +39,32 @@ public class Code.Pane : Gtk.Grid {
...
+        add_folder_button.action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER;
+        add_folder_button.tooltip_text = _("Add Project Folder…");
 
-        toolbar.add (add_button);
+        var project_more_button = new Gtk.MenuToolButton (null, null);
```
jeremypw: I would expect a "open-menu" symbolic icon rather than the default down arrow which I associate with downloading or a combobox. A "hamburger" type icon would be nice but I am not sure that one is available.

src/Widgets/Pane.vala
cassidyjames: There is no `folder-add` icon; the closest equivalents would be either `folder-new` (semantically incorrect because we're not creating a folder), or `folder-open`. I think the later makes sense here. `list-add` could work but it's more ambiguous, whereas using a folder icon makes it more clear at a glance that it's the button to open a folder.

src/Widgets/Pane.vala
cassidyjames: I was just using the default icon which I believe is used in other similar "dropdown" contexts, but I'd be fine to change to something else. I wanted to avoid using a settings-type icon because this is not necessarily settings, it's actions. `view-more-symbolic` might work. Thoughts from @elementary/ux?

cassidyjames:
>Did you still want the headerbar "add folder" button removed here?

@jeremypw I think so, because having it in both places doesn't make a ton of sense to me (and just muddies sidebar actions with global actions). I did remove it from the HeaderBar in a commit above, but I'd be open to @elementary/ux feedback as well.

src/Widgets/Pane.vala
jeremypw: I meant the folder icon with a plus sign (I thought it was named "folder-add" but it is named "folder-new").  The user does not know the name of the icon so visually it could mean "add folder".  However, I am happy to defer to the UX team.

jeremypw:
@cassidyjames I have removed the headerbar "New folder" button and used the same icon in the inline toolbar as users will be familiar with its meaning in this context.

cassidyjames:
@jeremypw we should **not** use the non-semantic icon, the open-folder icon is more correct.

src/Widgets/Pane.vala
```diff
@@ -39,11 +39,33 @@ public class Code.Pane : Gtk.Grid {
         toolbar.get_style_context ().add_class (Gtk.STYLE_CLASS_INLINE_TOOLBAR);
         toolbar.icon_size = Gtk.IconSize.SMALL_TOOLBAR;
 
-        var add_button = new Gtk.ToolButton (new Gtk.Image.from_icon_name ("folder-open-symbolic", Gtk.IconSize.BUTTON), null);
-        add_button.action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER;
-        add_button.tooltip_text = _("Add Project Folder…");
+        var add_folder_button = new Gtk.ToolButton (new Gtk.Image.from_icon_name ("folder-open-symbolic", Gtk.IconSize.BUTTON), null);
```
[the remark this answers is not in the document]
cassidyjames: @danrabbit mentioned this is due to the stylesheet, and not something within Code.

jeremypw:
@cassidyjames  OK. I now notice that this PR is merging into another older PR of yours anyway - I am confused as to which to work on.  I probably should not have pulled in the master branch - there is now an enormous diff for some reason :-(

cassidyjames:
@jeremypw ah I didn't realize that. I say we merge this into my original branch since I'm on board with the changes so far and then we can use that as the actual PR. Sorry about the mess!

PR #661, opened 2019-04-12
Add equal as zoom key

### LINKED ISSUE #627: Zoom in on Ctrl+Equal as well
Most browsers and interfaces that use Ctrl+/- to zoom also respond to `Ctrl`+`=`, since at least on US layouts the `+` is on the same key as `=`.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/69005286-zoom-in-on-ctrl-equal-as-well?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
~Intended to fix #627. However, Code just segfaults instead… :expressionless:~

Fixes #627

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -157,6 +157,7 @@ namespace Scratch {
             action_accelerators.set (ACTION_ZOOM_DEFAULT, "<Control>0");
             action_accelerators.set (ACTION_ZOOM_DEFAULT, "<Control>KP_0");
             action_accelerators.set (ACTION_ZOOM_IN, "<Control>plus");
+            action_accelerators.set (ACTION_ZOOM_IN, "<Control>equal");
             action_accelerators.set (ACTION_ZOOM_IN, "<Control>KP_Add");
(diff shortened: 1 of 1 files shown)


davidmhewitt:
@cassidyjames Fixed that for you :wink:

cassidyjames:
@davidmhewitt thanks! Why did adding the one accel require null terminating? Should it have been null terminated all along or is there something specific about adding this one?

davidmhewitt:
@cassidyjames `set_accels_for_action` expects a null terminated array and `MultiMap.to_array ()` doesn't give you a null terminated array, so the bigger question is why did it work at all before.

But, every time we use this pattern, we should be appending null to the array.

cassidyjames:
@davidmhewitt that's what I was wondering. Thanks for the help and the fix!

PR #665, opened 2019-04-14
Add close all folders except this context item

davidmhewitt:
Adds a context menu item to projects in the sidebar that allows you to close all folders except the one that was right clicked on.

Useful for when you're tidying up your open projects but want to keep the one you're currently working on.


src/FolderManager/ProjectFolderItem.vala
```diff
@@ -78,6 +79,9 @@ namespace Scratch.FolderManager {
             var close_item = new Gtk.MenuItem.with_label (_("Close Folder"));
             close_item.activate.connect (() => { closed (); });
 
+            var close_all_except_item = new Gtk.MenuItem.with_label (_("Close All Folders Except This"));
```
cassidyjames:
```suggestion
            var close_all_except_item = new Gtk.MenuItem.with_label (_("Collapse Other Folders"));
```

jeremypw:
I have fixed the conflict with master and updated text according to Cassidy's recommendation.

PR #669, opened 2019-04-23
Enable Brackets completion and auto-indent by default

### LINKED ISSUE #667: Enable bracket completion ootb
There have been a ton of fixes (#429), and I've been using this for quite a while now and it feels real good. I think we can enable it by default considering we're focusing on code over prose. For reference, the most popular code editors all have bracket completion by default, and it has become a standard and expected feature:

- Atom
- VS Code
- Sublime Text

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/72991886-enable-bracket-completion-ootb?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Fixes the typo from ff9abc3d53edf758d3d28692c87a50eeef49e614 and enables auto-indent since it compliments bracket completion and is also what most other editors do by default. Fixes #667.


davidmhewitt:
I'd been testing out the brackets completion plugin again the other day when I noticed it was proposed to be enabled by default again.

It's kind of a crappy experience when automatic indentation isn't turned on too, which it isn't by default. The extra inserted bracket just gets in the way on a new line.

So I guess we need to think about that.

cassidyjames:
@davidmhewitt that's true, I always enable that on my system and again, it's what VS Code, Atom, and Sublime (and probably others!) all do.

davidmhewitt:
Without automatic indentation, the bracket completion plugin is more of a hindrance to me than it is not having it turned on at all. But with both, the experience is pretty great.

So I think it's better to default both of them to on at the same time or modify the brackets completion plugin somehow to be less annoying around line breaks without automatic indentation.

cassidyjames:
@davidmhewitt if there are arguments against toggling both on by default, then I do think the bracket-completion plugin should put the closing bracket on the same level of indentation a line below if the user inserts a line break.

So (assuming `|` is the typing cursor) you'd go from:

```
    foo |
```

```
    foo (|)
```

```
    foo (
|
    )
```

I personally think that feels strange and we should just enable auto-indent, but I think it's technically the closest to being smart about brackets while also not auto-indenting that we could do.

tintou:
I agree that with auto-indent it is better, I'm still not using the bracket completer but it's a personal choice, so yeah feel free to merge it :)

PR #675, opened 2019-05-21
Git branch submenu is only sensitive if there is more than 1 branch.

### LINKED ISSUE #660: Change branch menu weirdly empty when only one branch
If I have a project with only a `master` branch, the Change Branch option and menu still shows but opens a weird empty menu. Instead, we should either make that item insensitive (seems simplest), or change it to a "Branch" item with all available branches and the selected one marked with a check.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/72303206-change-branch-menu-weirdly-empty-when-only-one-branch?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

evan-buss:
Fixes #660 
image

Git branch submenu is set insensitive if there is only a single branch instead of showing an empty menu.

### DIFF
--- a/src/FolderManager/ProjectFolderItem.vala
+++ b/src/FolderManager/ProjectFolderItem.vala
@@ -246,13 +246,19 @@ namespace Scratch.FolderManager {
                     string? branch_name = null;
                     try {
                         branch_name = branch.get_name ();
-                        if (branch_name != null && branch_name != cur_branch.get_name ()) {
+                        if (branch_name != null) {
                             var ref_name = ref_branch.get_name ();
                             if (ref_name != null) {
-                                var branch_item = new Gtk.MenuItem.with_label (branch_name);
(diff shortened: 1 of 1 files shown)


cassidyjames:
This works as a minimum fix, but I think I'd still like to see the other proposed fix (change it to a "Branch" item with all available branches and the selected one marked with a check) if possible. Marking it as insensitive doesn't tell the user _why_ they can't use the menu.

evan-buss:
I have implemented the other proposed fix if you would like to check it out. 

image

I agree that it is much clearer than having an insensitive option.

PR #676, opened 2019-05-28
Now temporary files will not be displayed in the sidebar

FlightBlaze:
### Now Elementary Code's temporary files will not be displayed in sidebar.   

Before
After  
  
No more trash in file tree

### DIFF
--- a/src/FolderManager/File.vala
+++ b/src/FolderManager/File.vala
@@ -91,7 +91,6 @@ namespace Scratch.FolderManager {
             }
         }
         
-        // checks if we're dealing with a temporary file
         public bool is_temporary {
             get {
                return path.has_suffix ("~");
--- a/src/FolderManager/File.vala
+++ b/src/FolderManager/File.vala
@@ -94,7 +94,7 @@ namespace Scratch.FolderManager {
         // checks if we're dealing with a temporary file
         public bool is_temporary {
             get {
-               return path.has_suffix("~");
+               return path.has_suffix ("~");
             }
          }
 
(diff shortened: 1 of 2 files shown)


src/FolderManager/FolderItem.vala
```diff
@@ -264,7 +264,7 @@ namespace Scratch.FolderManager {
                         if (!exists) {
                             if (file.is_valid_directory) {
                                 item = new FolderItem (file, view);
-                            } else {
+                            } else if(!file.is_temporary) {
```
```suggestion
                            } else if (!file.is_temporary) {
```

src/FolderManager/File.vala
```diff
@@ -90,6 +90,13 @@ namespace Scratch.FolderManager {
                 return false;
             }
         }
+        
+        // checks if we're dealing with a temporary file
+        public bool is_temporary {
+            get {
+               return path.has_suffix("~");
```
tintou:
```suggestion
               return path.has_suffix ("~");
```

src/FolderManager/File.vala
```diff
@@ -90,6 +90,13 @@ namespace Scratch.FolderManager {
                 return false;
             }
         }
+        
+        // checks if we're dealing with a temporary file
```
tintou: Remove this line, the variable name is self-describing it already :slightly_smiling_face:

FlightBlaze: Let it be! Comments is not temporary files

FlightBlaze:
> A few codestyle changes and it's ready to go

Changed. Now what? )

danrabbit:
Looks good. Thanks for your pull request!

PR #684, opened 2019-06-26
Spelling: EditorConfig → editor config

comradekingu:
"EditorConfig" resulted in a lot of direct/verbatim translations and bad coverage.


danrabbit:
I think that is working as intended since EditorConfig is a specific format/brand and not a class of formats

PR #674, opened 2019-05-20
Added "Open In" menu to the top-level project folder

evan-buss:
Top level folders now have an "open in" menu to open the project directory in other applications.

Closes: #671  

Screenshot

The existing "open in" menu code contained within the `FolderItem` component was refactored into a new function so that it can be accessed from the `ProjectFolderItem` since both "open in" menus are essentially the same.

PR #686, opened 2019-07-02
Add "Open In" menu to the top-level project folder

evan-buss:
Resubmission of  #674 

Resubmitted this pull request because my original forked branch was deleted. The latest requested changes have been implemented.

### DIFF
--- a/src/FolderManager/FolderItem.vala
+++ b/src/FolderManager/FolderItem.vala
@@ -99,7 +99,7 @@ namespace Scratch.FolderManager {
             delete_item.activate.connect (trash);
 
             var menu = new Gtk.Menu ();
-            menu.append (create_submenu_for_openin (info, file_type));
+            menu.append (create_submenu_for_open_in (info, file_type));
             menu.append (contractor_item);
             menu.append (new Gtk.SeparatorMenuItem ());
             menu.append (create_submenu_for_new ());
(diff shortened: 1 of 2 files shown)


src/FolderManager/FolderItem.vala
```diff
@@ -144,6 +110,50 @@ namespace Scratch.FolderManager {
...
+                    item_app.add (menuitem_grid);
+
+                    item_app.activate.connect (() => {
+                        launch_app_with_file (app_info, file.file);
+                    });
+                    open_in_menu.add (item_app);
+                }
+            }
+
+            open_in_menu.add (new Gtk.SeparatorMenuItem ());
```
donadigo: Wouldn't the `other_menuitem` be packed only if there apps added to the menu already? If yes, then move this line up right under `foreach` and check how many children the menu has:
```
if (open_in_menu.get_children ().length () > 0) {
    open_in_menu.add (new Gtk.SeparatorMenuItem ());
}
```

evan-buss: @donadigo It would look off if the separator was present without any menu items above it, I agree. On a fresh install, the "inode/directory" mime-type is handled by Files and Terminal without having to add any additional apps. I have added the suggested change anyway, because it is better to be safe.

Issue #649, opened 2019-03-08
Feature request: add undo or confirmation to restore file action


mcclurgm:
The "restore this file" action seems dangerous as it's currently designed. The way it destructively reverts the file you're editing seems like it could easily lead to data loss, especially since it can be easily mistaken for an undo. Speaking from my personal experience, I've lost some work because I mistakenly clicked this button, with no way to get it back.

The biggest issue for me is that its icon looks too close to the "undo" icon. When I'm looking to undo changes, I've learned that the leftward-pointing arrow is where I should go to accomplish that. The revert icon isn't the same, but the document background is easy to miss or ignore, especially since there is no on-screen undo functionality. The main issues for me are: I expect to be able to redo at any time, often doing an undo/redo cycle to see what changes I've made. This isn't possible in Code, so if I hunt for an undo button (as users who aren't big on keyboard shortcuts would do), I can't get back to what I want. Second, the number of changes it undoes is unexpectedly huge. (I'm actually not sure what state it restores to, so that may be useful to document somewhere.) I expect an undo button to go back a few seconds of work, but this can change massive parts of a file.

If I could suggest some behaviors to make this less destructive:
- Could you add a confirmation dialog when a user clicks the button?
- You could also include a "repeat changes" button that appears once you've clicked it, or a banner at the top of the window with an option to put the changes back in.

I'm not sure how to design this really, or if it's something you want to work on, but these suggestions could help mitigate what could happen if someone mistakenly clicks the button.

PR #687, opened 2019-07-03
Add confirmation dialog to Restore action

ChristopherNugent:
Requested in #649 and marked as confirmed. Text will probably need to be tweaked, but thats a quick change.

image

Please let me know if you need any changes to this.


src/Dialogs/RestoreConfirmationDialog.vala
```diff
@@ -0,0 +1,41 @@
...
+* General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+namespace Scratch.Dialogs {
+    public class RestoreConfirmationDialog : Granite.MessageDialog {
```
danrabbit: Instead of adding `namespace` here, you can do `public class Scratch.Dialogs.RestoreConfirmationDialog : Granite.MessageDialog { }` and save yourself an indentation level :)

src/Dialogs/RestoreConfirmationDialog.vala
```diff
@@ -0,0 +1,41 @@
...
+                buttons: Gtk.ButtonsType.NONE,
+                transient_for: parent
+            );
+        }
+
+        construct {
+            image_icon = new ThemedIcon ("dialog-warning");
+            primary_text = _("Restoring a file will undo all changes made since opening it");
+
+            secondary_text = _("Are you sure you want to restore this file?");
```
danrabbit: Do you think you could switch your primary and secondary text? I think that would be more in line with how other dialogs are formatted.

Also, please make dialog primary text Title Case (https://titlecaseconverter.com/ makes that easy)

ChristopherNugent:
The dialog currently appears like so:
image

PR #690, opened 2019-07-05
Use ApplicationCommandLine to resolve relative paths

### LINKED ISSUE #688: Incorrect relative path
When I use `cd` to navigate in my directories, and then using `io.elementary.code ./foo.txt`, Code always opens ~/foo.txt (the one in my home directory) instead of the foo.txt in my current directory.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/76480422-incorrect-relative-path?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #688 

We had some overly complicated leftover ways of trying to resolve relative paths from the command line and then passing them to the already running "remote" instance.

If we use the `ApplicationCommandLine` object, we can use that to resolve the relative paths for us and create the `File` object.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -31,7 +31,6 @@ namespace Scratch {
         public string default_font { get; set; }
         private static string _app_cmd_name;
         private static string _data_home_folder_unsaved;
-        private static string _cwd;
         private static bool print_version = false;
         private static bool create_new_tab = false;
         private static bool create_new_window = false;
@@ -111,9 +110,6 @@ namespace Scratch {
                 Utils.action_from_group (MainWindow.ACTION_NEW_TAB, window.actions).activate (null);
             }
 
-            // Set Current Directory
-            Environment.set_current_dir (_cwd);
-
             int args_length = args.length;
             // Open all files given as arguments
             if (args_length > 1) {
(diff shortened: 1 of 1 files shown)


jeremypw:
Fixes the linked problem.  Might affect some scripts that use the -c flag?  Is there any valid use for this flag - i.e. to set the environment current directory first?

davidmhewitt:
It looks like this option was added as a workaround to the same issue 5 years ago in https://github.com/elementary/code/commit/391773ec15b3e09c8c94dacee3b1ee4fded83cec .

It was used internally for the launching instance to pass its CWD to the already running instance and has since been broken. I guess it's possible that people have used it in scripts etc since then but it wasn't documented in the man page and I find it unlikely that people would be scripting a GUI editor in this way, so :man_shrugging: 

By switching to `command_line.create_file_for_arg`, the working directory from the environment of the running instance becomes irrelevant as it always works relative to the CWD of the launching instance, so this flag would no longer have any effect, hence removing it. I guess someone just needs to make a decision as to whether we should remove it or not.

It's a confusing option as IMO, I would expect a code editor launched on the command line to respect the CWD of the window I was typing the command into, not have a separate CWD argument.

jeremypw:
OK, I'll go ahead and approve in that case.

PR #691, opened 2019-07-05
Fix null file warning dialog with new window from dock.

### LINKED ISSUE #680: New Window from Plank dock gives error message
* Open Code (with a document)
* Right click Plank icon and choose "New Window"
* An error dialog appears - "FIle (null) cannot be opened - it is an unknown file type"

The new window is created but is below the existing one and cannot be seen.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/74975479-new-window-from-plank-dock-gives-error-message?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #680

Deal with null args after parsing

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -116,9 +116,12 @@ namespace Scratch {
 
             int args_length = args.length;
             // Open all files given as arguments
-            if (args_length > 1) {
+            if (args_length > 1) { /* First arg is program name */
                 File[] files = {};
                 foreach (unowned string arg in args[1:args_length]) {
+                    if (arg == null) { /* Recognised options changed to null */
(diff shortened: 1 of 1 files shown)


davidmhewitt:
We have `args[1:args_length]`, should it not be `args_length - 1`? Otherwise we go past the end of the array which might be where the null is coming from.

Issue #653, opened 2019-03-12
Deprioritize git-ignored files


cassidyjames:
Something Atom (and I'm sure other VCS-aware editors) does is deprioritizes git-ignored files by lowering their opacity in the sidebar. We could do something similar, or possibly even mark them as disabled in the SourceList.

cassidyjames:
This could also supersede #652, as those files should probably be gitignored in your project anyway.

roymckenzie:
@cassidyjames I did notice the `.git` directory is now visible, too. In VS Code the `.git` directory is hidden, even in directories that haven't been initialized with git. I created a `.git` directory manually and it doesn't appear that VS Code checks to see if it is valid. It may just check for the existence of the folder and hides it if it exists. Could implement a similar rule in Code fairly quickly I think.

cassidyjames:
@roymckenzie is that the case for all hidden (.dot) files in a folder, or just .git?

roymckenzie:
Just the .git folder is hidden in VS Code. Other .dot files and directories are visible.

cassidyjames:
Fixed in #705

PR #705, opened 2019-08-31
Deprioritize git-ignored files

seanballais:
Currently working on addressing #653. The image below shows my current progress.

PR Output

~I still have not worked on supporting glob patterns inside `.gitignore` files. It'll be up as soon as I am able to work on this again.~

~Glob patterns inside `.gitignore` files are now supported. However, skipping certain files from being ignored (as done via `!`) is not yet fully supported. Deprioritizing files inside an ignored folder is not yet done.~

I have added the feature that deprioritizes files. Changes to `.gitignore` automatically gets mirrored in the sidebar now. ~It still needs additional testing with folders and multiple project folders.~

~I also need an opinion from the UI team with regards what should the colour be for the text of the deprioritized files. At the moment, I have set the text colour to `#BFBFBF`.~ Instead of changing the text colour of the deprioritized files to `#BFBFBF`, I modified their opacity instead and set it to 50%. The decision to change the opacity instead of the text colour is based on the property of `dim-label` in the default elementaryOS stylesheet. Choosing `dim-label` as the basis for the aforementioned decision is due to @lainsce's suggestion of using `Gtk.STYLE_CLASS_DIM_LABEL`, which is the GTK enum constant of the `dim-label` CSS class. I originally set the opacity from 75%, as specified by `dim-label`, but reduced it to 50% since the text didn't look deprioritized enough.


eyelash:
Have you considered using [`Ggit.Repository.path_is_ignored`](https://valadoc.org/libgit2-glib-1.0/Ggit.Repository.path_is_ignored.html) for this?

lainsce:
> I also need an opinion from the UI team with regards what should the colour be for the text of the deprioritized files. At the moment, I have set the text colour to #BFBFBF.

Instead of doing that, style the `Gtk.Label` with the `Gtk.STYLE_CLASS_DIM_LABEL` class, and remove it when the file is no longer in the gitignore file.
This ensures readability and contrast.

seanballais:
> Have you considered using `Ggit.Repository.path_is_ignored` for this?

@eyelash, I haven't. I'll have a look at it once I am able to work on this again. Thanks!

> Instead of doing that, style the Gtk.Label with the Gtk.STYLE_CLASS_DIM_LABEL class, and remove it when the file is no longer in the gitignore file. This ensures readability and contrast.
> 
> I can provide a code snippet if it helps you with this.

@lainsce, oh. I never knew about that class. I'm not sure if there is a `Gtk.Label` inside `SourceList.Item`, but I'll make use of that knowledge about that class. Providing me a code snippet on it would be helpful. Thanks! :smiley:

cassidyjames:
I'd also consider marking the item as insensitive instead of applying a class on the label itself, though that does probably imply that you wouldn't be able to edit that file. I guess I'm not sure if we'd want that or not.

seanballais:
I think it'd be best if we go for applying the class to the label instead of marking it as insensitive.

On another note, @cassidyjames, are we going to deprioritize files only if the project folder is a local git repository or are we still going to deprioritize if a `.gitignore` file exists?

seanballais:
In my latest commit, I used `Ggit.Repository.path_is_ignored ()` in checking if a file or folder is git-ignored, as suggested by @eyelash. It works perfectly so far. I used it in the function `is_file_gitignored ()` in `ProjectFolderItem`.

When catching any error `path_is_ignored ()` throws, I would log the error message. However, I am unsure what log level to use. I noticed in the codebase that `warning ()` is sometimes used, and sometimes `critical ()`. But, I can only speculate the reasons why a `warning ()` would be used instead of `critical ()` and vice-versa. In the `try-catch` block I have in `is_file_gitignored ()`, when logging an error message, I used a log level of `WARNING` since I don't feel that any error that would arise would deserve a log level of `ERROR` nor `CRITICAL` and I followed the log level that a similar `try-catch` block, the first such block in `do_git_update ()` in `ProjectFolderItem`, used. However, I have a feeling that the log level should probably be `ERROR`. So, I'm not really confident that the log level should be `WARNING`.

What is the recommended log level to use in the case I presented? Is there a convention specific for elementaryOS codebases that details which log level to use?

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -218,15 +235,49 @@ namespace Scratch.FolderManager {
                 }
 
                 item.name = item.file.name;
-                item.markup = null;
                 item.activatable = null;
 
+                if (is_file_gitignored (item)) {
+                    item.markup = Markup.printf_escaped ("<span fgalpha='50&#37;'>%s</span>", item.name);
```
cassidyjames: It feels weird to have this repeated as well as in `deprioritize_gitignored_files ()`, could we just re-use that method?

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -218,15 +235,49 @@ namespace Scratch.FolderManager {
...
 
+        private void update_git_deprioritized_files () {
+            deprioritize_gitignored_files (this);
+        }
+
+        private void deprioritize_gitignored_files (Item top_level_item) {
+            foreach (var child in top_level_item.children) {
+                var item = child as Item;
+                if (is_file_gitignored (item)) {
+                    item.markup = Markup.printf_escaped ("<span fgalpha='50&#37;'>%s</span>", item.name);
```
cassidyjames: 50% gets pretty inaccessible, so let's use 75% and italic:

```suggestion
                    item.markup = Markup.printf_escaped ("<span fgalpha='75&#37;'><i>%s</i></span>", item.name);
```

PR #709, opened 2019-09-17
SearchBar: Fix shift+enter

### LINKED ISSUE #708: SearchBar: Checking for SHIFT_MASK doesn't work
Found when trying to implement a similar find bar in Ephemeral.  It looks like the Shift_L and Shift_R keys are fired as their own individual key press events, so this is never reachable.

https://github.com/elementary/code/blob/fb8af720bd7fb02516b54d78b2036648eac8ea34/src/Widgets/SearchBar.vala#L396-L402

cassidyjames:
It looks like this fixes #708 

https://github.com/midori-browser/core/blob/34e938e2abf239df658bec91b6d8e708f5177f7a/core/urlbar.vala#L89

### DIFF
--- a/src/Widgets/SearchBar.vala
+++ b/src/Widgets/SearchBar.vala
@@ -394,7 +394,7 @@ namespace Scratch.Widgets {
             }
 
             string key = Gdk.keyval_name (event.keyval);
-            if ((event.state & Gdk.ModifierType.SHIFT_MASK) != 0) {
+            if (Gdk.ModifierType.SHIFT_MASK in event.state) {
                 key = "<Shift>" + key;
             }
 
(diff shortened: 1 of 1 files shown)


davidmhewitt:
I think `Gdk.ModifierType.SHIFT_MASK in event.state` is equivalent and more readable :slightly_smiling_face:

src/Widgets/SearchBar.vala
```diff
@@ -394,7 +394,7 @@ namespace Scratch.Widgets {
             }
 
             string key = Gdk.keyval_name (event.keyval);
-            if (event.state == Gdk.ModifierType.SHIFT_MASK) {
+            if ((event.state & Gdk.ModifierType.SHIFT_MASK) != 0) {
```
cassidyjames:
```suggestion
            if (Gdk.ModifierType.SHIFT_MASK in event.state) {
```

cassidyjames:
@davidmhewitt d'oh that makes so much more sense. Thanks. Addressed.

PR #712, opened 2019-10-03
added "keystroke a: insert after current position" action

### LINKED ISSUE #614: Vim emulation does not parse keystroke "a"

clobrano:
Support for keystroke "a", which VIM maps into "insert after current position", was missing from the `handle_key_press` function.

Fixed adding the case using the `Gdk.Key.i` as example.

closes #614


plugins/vim-emulation/vim-emulation.vala
```diff
@@ -80,6 +80,18 @@ public class Scratch.Plugins.VimEmulation : Peas.ExtensionBase,  Peas.Activatabl
...
+                    return false;
+                } else {
+                    // clean action string
+                    action = "";
+                }
+
+                mode = Mode.INSERT;
+                view.move_cursor (Gtk.MovementStep.VISUAL_POSITIONS, 1, false);
+                debug ("Vim Emulation: INSERT Mode!");
+                return true;
```
jeremypw: Why is this not implemented further down (where handling of `Gdk.Key.A` is implemented?

clobrano: It seemed to me that the cases where ordered alphabetically, so I applied the patch for `Gdk.Key.a` at the top (also `Gdk.Key.i` and `Gdk.Key.I` are not close)

jeremypw: But `Gdk.Key.A` is processed in a different switch statement.  The first one is related to setting the correct mode - not actions.

jeremypw:
I am not a Vim user so could I confirm that the expected action is to start inserting text after the character that is on the right of the cursor?  The effect of pressing `<Esc>a` is to move the cursor one character to the right (and then enter INSERT mode).

clobrano:
Uhm, I was double checking the documentation to confirm, but while the usual use case is indeed "start inserting text after 1 character", it is possible to prefix the command with a number N, then the effect is to start inserting text after N characters.
If @mcclurgm's implementation already has this feature fully implemented, then better keep his PR (patched as [discussed](https://github.com/elementary/code/issues/614#issuecomment-538688594)) and just reject this.

jeremypw:
As far as I can tell, @mcclurgm 's implementation of the `a` command is the same as yours.  That PR has the disadvantage of changing several things at once and has not passed CI.  I am inclined to approve yours and then the other PR can be update and preferably split into smaller PRs.

clobrano:
I see, could you give me some more time to see if a complete implementation is already doable?

clobrano:
It is possible :)

mcclurgm:
Great! I'll remove this change from my PR then. 
@jeremypw would you suggest that I split it up into individual command changes or should I batch them by subject somehow? I don't know if it would be better for you to review a lot of tiny changes or a couple larger ones.

jeremypw:
@mcclurgm It is better to split into the smallest PRs that make sense to merge alone.  Small focussed "bitesize" PRs are likely to get reviewed and merged much more quickly.

mcclurgm:
Alright, I'll start working on that. Thanks for the suggestion!

jeremypw:
@mcclurgm Are you making any progress on splitting this?

mcclurgm:
Not really sorry, I got a little stuck and then school started again. I'll try to get back on it. I did get one simple PR #737.

mcclurgm:
@clobrano are you sure that adding a number before `a` inserts after N characters? According to the [docs](https://vimhelp.org/insert.txt.html#a) I found, it inserts the text after one character, N times. I tried this in vim itself and that's what it did.

clobrano:
I think you're right, I don't know I was convinced it worked differently. However, this makes things way harder than expected, using a number code should

- take into a different buffer any sequence typed after `a` until `<esc>` is pressed
- copied the buffer N times
- go back to normal mode

mcclurgm:
I think that's correct. I think this may be doable using the `action` string, which is used to implement the `.` command. 

It looks like this sort of behavior isn't accounted for in the code at all though. It would need some design to figure out how to make it reuse the buffer N times though, I'm not sure how that would fit in with the current code. Maybe we could make it check for a number when you press escape?

clobrano:
I should look into this again, I don't even have Elementary installed anymore, sorry.

> Maybe we could make it check for a number when you press escape?
I can't follow, why checking for a number?

mcclurgm:
Oh--sorry. When I said check for a number, I was thinking of the `number` variable. When we start inserting it should contain the number of times to do that insert, so when we leave insert mode it should tell us what to do. Unless there's something I'm missing. Does that make sense?

clobrano:
Yes, you are right, basically replying the last buffer for `N-1` times, since the first sequence is already written.
I'll look into the `action` string, then, but I can't be sure when, sorry.

mcclurgm:
Maybe it makes sense to merge this as is, and then deal with the multiple-insert issue later once all the different insert options are in the codebase?

clobrano:
It might sounds lazy, but I agree. At least the basic feature would be in and it's quite used

Issue #131, opened 2017-06-28
Drop "Draw Spaces: Never" preference


cassidyjames:
Now that we have the awesome automagic show-on-highlight default, I don't see a lot of value in carrying a preference to toggle to other largely inferior modes.

tintou:
I'm completely against it, I use the "always" setting and I'm happy with it, people should always be aware that they are using space or tabs.

danrabbit:
I personally use the new highlight setting, but I can see how this might be something more personal. Something to ask maybe is if anyone is using the "never" option. If they aren't at least we could remove that and just have "always" and "smart"

tintou:
"smart" being "on selection", nothing that smart ;)

cassidyjames:
Since we can't come to a consensus for dropping this entirely, I'm repurposing for dropping the "Never" option.

PR #722, opened 2019-10-27
Drop "Never" from Draw Spaces

jorgemanzo:
Removed the `Never` option from the Draw Spaces combobox. 

Question: should the `/data/io.elementary.code.gschema.xml` file also be modified to remove the option?

As always feedback is greatly appreciated!

fixes: #131


jorgemanzo:
Okay, I gave it a shot! I've removed the option from the combobox. But that setting also is mentioned in the schema, so does something extra need to be done in case someone already saved this setting?

cassidyjames:
This seems to work; if someone had "never" selected before, the combobox is empty unless/until they choose one of the two supported options.

Design-wise, since we've moved from a ternary option to a binary one, we could also revisit how it's displayed. My thought is to just expose it as "Visible whitespace" with a switch: on is "Always," and off is "For Selection". Thoughts @danrabbit?

cassidyjames:
@jorgemanzo to answer your earlier question, I don't think we want to remove it from the schema; I'm not sure what happens if someone has that schema selected and then we remove it. We might want to mention that it's deprecated and not exposed in the UI though.

PR #723, opened 2019-10-28
Desktop action: “New Document” → “New File”

cassidyjames:
Since it's not a general-purpose document, but specifically a file for code. 

Edit: After the Twitter discussion, yeah, we don't need to duplicate the name of the app here. Instead, Applications Menu search should show the name of the app with the results.


cassidyjames:
If we do something smart and more prominently show the name of the associated app in the Applications Menu search, then just “New File” would also be fine.

PR #737, opened 2019-12-08
Vim emulation: startup in command mode

mcclurgm:
Following the behavior of vim and other vim-emulators (like gnome-builder, atom-vim-mode, etc), this changes the plugin to start up in command mode. This should make its behavior more consistent for users.


jeremypw:
What worries me about this is that there is no visual clue as to what mode the plugin is in (as far as I can see). So, with this PR, at start up the view appears unresponsive (unless you happen to hit a command key.  How closely do we want to emulate Vim? Running Vim in a terminal behaves and appears quite different (e.g. hitting `:` switches to command mode and the cursor moves to the bottom of the screen)

mcclurgm:
That's a fair question. I haven't used vim mode in a good while, but when I was used to vim it's completely muscle memory to move around the document and then enter insert mode when I open a document. When it starts up in insert mode I wound up entering a bunch of random text in the document pretty much every time, which isn't ideal. So that's my pitch for why this could make sense. I'm not sure if that or appearing unresponsive is worse though, tbh I never thought about that issue when I wrote this.

jeremypw:
We need input from an actual user of Vim (and, preferably, this plug).  Otherwise we can just merge and wait for any issues to arise (not sure how widely used this plug is).

jeremypw:
On reflection, I think it would be OK to merge this and address the issue of indicating which mode the plugin is in (and if it is even loaded) separately as this exists in master anyway.

jeremypw (approved):

PR #742, opened 2019-12-14
Add Tooltip Explanation

### LINKED ISSUE #598: No explanation for green and yellow lights
If you open the directory view for dir integrated with git,
you can see changes of a file with green or yellow circle on the right side.

However it is not clear what is the exact meaning of the circle.

Possible solution: Offer a tooltip with an explaination.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/66869870-no-explanation-for-green-and-yellow-lights?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jorgemanzo:
Adds Tooltip explanations for modified and newly added files in the `SourceList`. Tooltip includes the filename, plus `, Modified` or `, New` depending on if the file is modified or new. Feedback on this is greatly appreciated!

fixes #598

Screenshot from 2019-12-13 23-44-55


jorgemanzo:
Okay, I made the strings ready for translation, and I also switched the tooltips to be formatted strings instead. Should the _Modified_ and the _New_ tooltips be lowercase?

jeremypw:
No, tooltips like labels should be in title-case.  Is this ready for review now?

jeremypw:
Constructing translated strings like this can be problematic, e.g. in RTL environments. I am not sure we need to repeat the filename in the tooltip as it is in the item (although possibly truncated). I think I will ask for a design and i18 opinion.

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -185,6 +185,7 @@ namespace Scratch.FolderManager {
                 find_items (this, path, ref modified_items);
                 foreach (var modified_item in modified_items) {
                     modified_item.activatable = modified_icon;
+                    modified_item.tooltip = _("%s, Modified").printf (modified_item.name);
```
danrabbit: It would probably be better to do something like:
```suggestion
                    modified_item.tooltip = "%s, %s".printf (modified_item.name, _("Modified"));
```

So that you're not including formatting in the translatable string.

And in fact, I would then do some more fancy pango magic to mimic accel tooltips like this:
```
    modified_item.tooltip_markup = """%s\n<span weight="600" size="smaller" alpha="75%">%s</span>""".printf (modified_item.name, _("Modified"));
```

ryonakano: It's common to use an ideographic comma "、" instead of a comma "," in Japanese, so I would like to *translate* this comma into an ideographic comma too. Thus, I personally would not like this suggestion to be committed.

jeremypw: Are there languages where it would be more natural for "Modified" or "New" to precede the filename?  That too would be another reason not to take the formatting out of the translatable string.

jeremypw: @danrabbit There is a problem with using markup in the tooltip - `modified_item` is not a widget, it is a subclass of Granite.Widgets.SourceList.ExpandableItem, which is an object.  Granite exposes the property `tooltip` but not `tooltip-markup`.

jeremypw:
There is also a problem that the comma is in fact translatable - it is a different character in some languages.  So I think `_("%s, Modified").printf (modified_item.name)` is the best available solution.

PR #743, opened 2019-12-17
FolderManager: Rename before create new file

### LINKED ISSUE #213: Creating new file immediately opens it

jeremypw:
Fixes #213 

An alternative to #650, which should be less prone to race conditions.  The filename to be created is entered *before* any file is created using a temporary renaming item.  To match Files behaviour, if <Esc> is pressed (or a blank name entered) then a file with the default name is created rather than cancelling the file creation.

It proved necessary to poll the SourceList to determine when editing finishes because cancelling the editing does not generate any signal and we need to remove the temporary item in either case.

If this were fixed in Granite then a simplification here would be possible.

### DIFF
--- a/src/FolderManager/FolderItem.vala
+++ b/src/FolderManager/FolderItem.vala
@@ -366,6 +366,12 @@ namespace Scratch.FolderManager {
 
         construct {
             edited.connect (rename);
+
+            if (is_folder) {
+                icon = GLib.ContentType.get_icon ("inode/directory");
+            } else {
+                icon = GLib.ContentType.get_icon ("text");
+            }
         }
 
         protected new void rename (string new_name) {
--- a/src/FolderManager/FolderItem.vala
+++ b/src/FolderManager/FolderItem.vala
@@ -327,7 +327,7 @@ namespace Scratch.FolderManager {
                 return;
             }
 
-            var name = is_folder ? _("untitled folder") : _("new file");
+            unowned string name = is_folder ? _("untitled folder") : _("new file");
             var new_file = file.file.get_child (name);
             var n = 1;
 
@@ -356,7 +356,7 @@ namespace Scratch.FolderManager {
                         var new_name = rename_item.name;
                         remove (rename_item);
                         var path = file.file.get_path () + "/" + new_name;
-                        GLib.File gfile = GLib.File.new_for_path (path);
+                        var gfile = GLib.File.new_for_path (path);
                         try {
                             if (is_folder) {
                                 gfile.make_directory ();
@@ -392,7 +392,7 @@ namespace Scratch.FolderManager {
 
         construct {
             editable = true;
-            edited.connect (rename);
+            edited.connect (on_edited);
 
             if (is_folder) {
                 icon = GLib.ContentType.get_icon ("inode/directory");
@@ -401,7 +401,7 @@ namespace Scratch.FolderManager {
             }
         }
 
-        protected new void rename (string new_name) {
+        private void on_edited (string new_name) {
             if (new_name != "") {
                 name = new_name;
             }
--- a/src/FolderManager/FolderItem.vala
+++ b/src/FolderManager/FolderItem.vala
@@ -355,8 +355,7 @@ namespace Scratch.FolderManager {
                     } else {
                         var new_name = rename_item.name;
                         remove (rename_item);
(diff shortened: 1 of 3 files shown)


src/FolderManager/FolderItem.vala
```diff
@@ -328,67 +320,90 @@ namespace Scratch.FolderManager {
...
+        private void on_add_new (bool is_folder) {
             if (!file.is_executable) {
                 // This is necessary to avoid infinite loop below
                 warning ("Unable to open parent folder");
                 return;
             }
 
-            var new_folder = file.file.get_child (_("untitled folder"));
-
+            var name = is_folder ? _("untitled folder") : _("new file");
```
```suggestion
            unowned string name = is_folder ? _("untitled folder") : _("new file");
```

src/FolderManager/FolderItem.vala
```diff
@@ -328,67 +320,90 @@ namespace Scratch.FolderManager {
...
+
+                /* Need to poll view editing as no signal is generated when canceled (Granite bug) */
+                Timeout.add (200, () => {
+                    if (view.editing) {
+                        return Source.CONTINUE;
+                    } else {
+                        var new_name = rename_item.name;
+                        remove (rename_item);
+                        var path = file.file.get_path () + "/" + new_name;
+                        GLib.File gfile = GLib.File.new_for_path (path);
```
```suggestion
                        var gfile = file.file.get_child_for_display_name (rename_item.name);
                        remove (rename_item);
```

src/FolderManager/FolderItem.vala
```diff
@@ -328,67 +320,90 @@ namespace Scratch.FolderManager {
...
+        }
 
-                if (!children_loaded) {
-                    clear ();
-                    add_children ();
-                    children_loaded = true;
-                }
-            } catch (Error e) {
-                warning (e.message);
+        protected new void rename (string new_name) {
```
```suggestion
        protected new void rename (string new_name) {
```
this should really be avoided, I really don't like the use of the `new` keyword.

src/FolderManager/FolderItem.vala
jeremypw: I did not think that there would be any string copying in this context (assigning from an expression not a variable or property)?  But I'll follow your advice.

src/FolderManager/FolderItem.vala
jeremypw: No difference in you suggested line?  I'll see if I can find an alternative.

jeremypw:
@tintou Thanks for the review :-)   Do you think this a workable approach compared to the alternative PR (which I found difficult to debug)?

src/FolderManager/FolderItem.vala
jeremypw: On further inspection, new is not needed - I was confused between FolderManager.Item (which was the initial base class I used) and Granite.Widgets.SourceList.Item.

src/FolderManager/FolderItem.vala
jeremypw: OK

tintou: have you seen this, it looks safer and more efficient to what you're doing :)

src/FolderManager/FolderItem.vala
tintou: `var name = "foo"` results to `name = g_strdup ("foo");` in C when `unowned string name = "foo"` results in `name = "foo"` so it's a tiny gain here ;)

tintou:
Yeah it looks really better, hopefully once Gtk4 will be out Granite Source View won't be a thing anymore :slightly_smiling_face:

src/FolderManager/FolderItem.vala
jeremypw: OOps - I just saw the change to var.  I'll use your improved version.

jeremypw:
Missed change made - sorry about that.

PR #753, opened 2020-01-15
Pane: Use actionbar and adjust tooltips

danrabbit:
* Use actionbar to save us a few lines with `pack_end`
* Use `view-more-symbolic` for menubutton icon to be consistent with the icon elsewhere
* No ellipses for a tooltip that doesn't spawn a new window or view
* "Open" project folder instead of "Add" since we're not creating a new folder
* Tooltips are sentence case

### DIFF
--- a/src/Widgets/Pane.vala
+++ b/src/Widgets/Pane.vala
@@ -36,16 +36,12 @@ public class Code.Pane : Gtk.Grid {
         stack_switcher.stack = stack;
         stack_switcher.homogeneous = true;
 
-        var toolbar = new Gtk.Toolbar ();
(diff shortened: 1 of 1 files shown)


cassidyjames:
@danrabbit I know “Add” was intentional because you're adding a project to Code/the sidebar. VS Code says “Add folder to workspace” whereas Atom says “Add Project Folder”.

That said, if “Open” is more clear/consistent and doesn't seem to confuse people, I'm also fine with that.

danrabbit:
Yeah as far as I know we've only ever used "Add" when we're creating a new object and "Open" when we're importing an object

PR #754, opened 2020-01-15
Add accelerators to menu items

danrabbit:
It looks like the majority of menuitems come directly from Gtk.SourceView so we're not gonna get accels there unless we completely override that menu


danrabbit:
Granite tries to grab the first shortcut according to `Gtk.Application.get_accels_for_action ()`: https://github.com/elementary/granite/blob/master/lib/Widgets/AccelLabel.vala#L103

it's possible that Gtk.Application puts shortcuts in some kind of order and it doesn't matter what order we insert them but I haven't tested that

PR #764, opened 2020-02-10
Remember and restore state of cyclic search button

### LINKED ISSUE #758: Remember cyclic search setting
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

---

Code doesn't remember the cyclic search setting upon closing and reopening. It would be nice if it did. This is probably not difficult to implement (probably you could just add a `gsetting` or something).

jeremypw:
Fixes #758


src/Widgets/SearchBar.vala
```diff
@@ -30,7 +30,7 @@ namespace Scratch.Widgets {
          * "Down", it will go at the start of the file to search for the content
          * of the search entry.
          **/
-        private Gtk.ToggleButton tool_cycle_search;
+        public Gtk.ToggleButton tool_cycle_search {get; construct;}
```
danrabbit: this should probably be `private set` since it's not assigned in the constructor

```suggestion
        public Gtk.ToggleButton tool_cycle_search { get; private set; }
```

jeremypw: It does get assigned in the constructor - see line 83?

Issue #663, opened 2019-04-13
Code deb package still contains references to Scratch


matzipan:
```
sudo apt show io.elementary.code

Description: text editor that works for you
 Scratch is the text editor that works for you. It auto-saves your files,
 meaning they're always up-to-date. Plus it remembers your tabs so you never
 lose your spot, even in between sessions.
 .
 Make it yours. Scratch is written from the ground up to be extensible. Keep
 things super lightweight and simple, or install extensions to turn Scratch
 into a full-blown IDE; it's your choice. And with a handful of useful
 preferences, you can tweak the behavior and interface to your liking.
 .
 It's elementary. Scratch is made to be the perfect text editor for elementary,
 meaning it closely follows the high standards of design, speed, and
 consistency. It's sexy, but not distracting.
 .
 Works with your language. Whether you're crafting code in Vala, scripting with
 PHP, or marking things up in HTML, Scratch has you covered. Experience full
 syntax highlighting with nearly all programming, scripting, and markup
 languages.
```

PR #779, opened 2020-03-12
Scratch to Code in debian/control

Timecraft:
Kind of a random thing, but I was playing around with `apt show` and ended up doing `apt show io.elementary.code` and saw that the description still reads "Scratch."
Since elementary Scratch has been renamed to elementary Code, I believe it would be a good idea to reflect this in the debian packaging description. Let me know if I am wrong about this or need to change anything!

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -36,21 +36,21 @@ Depends: exuberant-ctags,
          ${misc:Depends},
          ${shlibs:Depends}
 Description: text editor that works for you
- Scratch is the text editor that works for you. It auto-saves your files,
+ Code is the text editor that works for you. It auto-saves your files,
  meaning they're always up-to-date. Plus it remembers your tabs so you never
  lose your spot, even in between sessions.
  .
- Make it yours. Scratch is written from the ground up to be extensible. Keep
- things super lightweight and simple, or install extensions to turn Scratch
+ Make it yours. Code is written from the ground up to be extensible. Keep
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -36,21 +36,21 @@ Depends: exuberant-ctags,
...
  preferences, you can tweak the behavior and interface to your liking.
  .
- It's elementary. Scratch is made to be the perfect text editor for elementary,
+ It's elementary. Code is made to be the perfect text editor for elementary,
  meaning it closely follows the high standards of design, speed, and
  consistency. It's sexy, but not distracting.
  .
  Works with your language. Whether you're crafting code in Vala, scripting with
- PHP, or marking things up in HTML, Scratch has you covered. Experience full
+ PHP, or marking things up in HTML, code has you covered. Experience full
```
Zyten, line 53: ~code~ Code

Zyten:
This should fix #663

debian/control
Timecraft: Are you sure? I didn't change that line, so it was "code" originally, and I believe it is talking about "code" as in actual code, not the product/app.

Timecraft: Oh, I was lookin at the wrong instance of "code". Let me fix that real quick!

Timecraft: That **should** do it!

tintou:
thanks for this :)

PR #782, opened 2020-03-18
Closes files opened from the side bar when their corresponding project is closed

### LINKED ISSUE #774: Close associated files when closing a project
It feels like it would keep things cleaner to close any open files for a project when closing that project.

igordsm:
Closes #774 : when a project folder is closed all files opened by clicking in the sidebar are closed too. Files opened by using the "open" button in the toolbar (or Ctrl+O) remain even if they belong to a project folder that was closed. 

Keep in mind that this is my first experience with Vala. I am aware that my solution is probably not the best, but I am willing to make it so. 

Thank you for your contributions on this and hope to hear from you soon.

### DIFF
--- a/src/FolderManager/FileItem.vala
+++ b/src/FolderManager/FileItem.vala
@@ -22,9 +22,19 @@ namespace Scratch.FolderManager {
     /**
      * Normal item in the source list, represents a textfile.
      */
+
+
     internal class FileItem : Item {
-        public FileItem (File file, FileView view) {
+        private ProjectFolderItem _root;
+        public ProjectFolderItem project_root {
+            get {
+                return _root;
+            }
+        }
+
+        public FileItem (File file, FileView view, ProjectFolderItem root) {
             Object (file: file, view: view);
+            this._root = root;
         }
 
         public override Gtk.Menu? get_context_menu () {
--- a/src/FolderManager/FileItem.vala
+++ b/src/FolderManager/FileItem.vala
@@ -26,7 +26,7 @@ namespace Scratch.FolderManager {
 
     internal class FileItem : Item {
         private ProjectFolderItem _root;
-        public ProjectFolderItem project_root {
+        public ProjectFolderItem project_root { get; construct; }
             get {
                 return _root;
             }
--- a/src/FolderManager/FileItem.vala
+++ b/src/FolderManager/FileItem.vala
@@ -33,7 +33,7 @@ namespace Scratch.FolderManager {
         }
 
         public FileItem (File file, FileView view, ProjectFolderItem root) {
-            Object (file: file, view: view);
+            Object (file: file, view: view, project_root: root);
             this._root = root;
         }
 
--- a/src/FolderManager/FileView.vala
+++ b/src/FolderManager/FileView.vala
@@ -25,7 +25,8 @@ namespace Scratch.FolderManager {
     internal class FileView : Granite.Widgets.SourceList, Code.PaneSwitcher {
         private FolderManagerSettings settings;
 
-        public signal void select (string file);
+        public signal void select (string file, string project_root);
+        public signal void project_closed (string project_root);
 
         // This is a workaround for SourceList silliness: you cannot remove an item
         // without it automatically selecting another one.
@@ -52,7 +53,8 @@ namespace Scratch.FolderManager {
             }
 
             if (item is FileItem) {
-                select ((item as FileItem).file.path);
+                var it = item as FileItem;
+                select (it.file.path, it.project_root.file.path );
             }
         }
 
@@ -142,6 +144,7 @@ namespace Scratch.FolderManager {
             folder_root.closed.connect (() => {
                 root.remove (folder_root);
                 write_settings ();
+                project_closed (folder_root.file.path);
             });
 
             folder_root.close_all_except.connect (() => {
--- a/src/FolderManager/FileView.vala
+++ b/src/FolderManager/FileView.vala
@@ -53,7 +53,7 @@ namespace Scratch.FolderManager {
             }
 
             if (item is FileItem) {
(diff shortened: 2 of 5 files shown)


src/Services/Document.vala
```diff
@@ -90,6 +90,7 @@ namespace Scratch.Services {
         public string original_content;
         private string last_save_content;
         public bool saved = true;
+        public string project_folder;
```
tintou:
```suggestion
        public string? project_folder;
```

src/MainWindow.vala
```diff
@@ -342,6 +342,23 @@ namespace Scratch {
...
+            folder_manager_view.project_closed.connect((a) => {
+                foreach (Scratch.Widgets.DocumentView? v in split_view.views) {
+                    GLib.List<Services.Document> to_close = new GLib.List<Services.Document>();
+                    foreach (Services.Document doc in v.docs) {
+                        if (doc.project_folder == a) {
+                            to_close.append (doc);
+                        }
+                    }
+
+                    foreach (Services.Document doc in to_close) {
```
tintou:
```suggestion
                    foreach (unowned Services.Document doc in to_close) {
```

src/MainWindow.vala
```diff
@@ -342,6 +342,23 @@ namespace Scratch {
                 }
             });
 
+
+            folder_manager_view.project_closed.connect((a) => {
+                foreach (Scratch.Widgets.DocumentView? v in split_view.views) {
+                    GLib.List<Services.Document> to_close = new GLib.List<Services.Document>();
```
tintou:
```suggestion
                    var to_close = new GLib.List<Services.Document>();
```
We prefer to have the type written once when it's clear on the right hand side what it is :slightly_smiling_face:

src/FolderManager/FileView.vala
```diff
@@ -52,7 +53,8 @@ namespace Scratch.FolderManager {
             }
 
             if (item is FileItem) {
-                select ((item as FileItem).file.path);
+                var it = item as FileItem;
```
tintou:
```suggestion
                unowned FileItem it = (FileItem) item;
```
No need to increase the reference here (so using unowned)
Also we can silence cast it as we already know that item is a FileItem

src/FolderManager/FileItem.vala
```diff
@@ -22,9 +22,19 @@ namespace Scratch.FolderManager {
     /**
      * Normal item in the source list, represents a textfile.
      */
+
+
     internal class FileItem : Item {
-        public FileItem (File file, FileView view) {
+        private ProjectFolderItem _root;
+        public ProjectFolderItem project_root {
```
tintou: In this case, it might be better to do:
```suggestion
        public ProjectFolderItem project_root { get; construct; }
```

src/FolderManager/FileItem.vala
```diff
@@ -22,9 +22,19 @@ namespace Scratch.FolderManager {
...
+            get {
+                return _root;
+            }
+        }
+
+        public FileItem (File file, FileView view, ProjectFolderItem root) {
             Object (file: file, view: view);
```
tintou: That way you can do:
```suggestion
            Object (file: file, view: view, project_root: root);
```

src/FolderManager/FolderItem.vala
```diff
@@ -183,7 +183,7 @@ namespace Scratch.FolderManager {
                 if (child.is_valid_directory) {
                     item = new FolderItem (child, view);
                 } else if (child.is_valid_textfile) {
-                    item = new FileItem (child, view);
+                    item = new FileItem (child, view, get_root_folder());
```
```suggestion
                    item = new FileItem (child, view, get_root_folder ());
```
style fix

src/FolderManager/FolderItem.vala
```diff
@@ -288,7 +288,7 @@ namespace Scratch.FolderManager {
                             if (file.is_valid_directory) {
                                 item = new FolderItem (file, view);
                             } else if (!file.is_temporary) {
-                                item = new FileItem (file, view);
+                                item = new FileItem (file, view, get_root_folder());
```
```suggestion
                                item = new FileItem (file, view, get_root_folder ());
```
style fix

src/MainWindow.vala
```diff
@@ -331,9 +331,9 @@ namespace Scratch {
 
             folder_manager_view = new FolderManager.FileView ();
 
-            folder_manager_view.select.connect ((a) => {
+            folder_manager_view.select.connect ((a, b) => {
```
tintou:
```suggestion
            folder_manager_view.select.connect ((a, project_root) => {
```
The original code here has bad practices, better always use understandable temporary variables

igordsm:
@tintou thank you for your revisions. I have accepted them and fixed the last linting error. CI now passes.

jeremypw:
@igordsm @tintou  The PR #833 has been submitted addressing the same issue, but in a simpler way. Although it has some issues these are easily fixed.  Are there any advantages to your approach?

igordsm:
@jeremypw When I created this PR I wanted to separate files opened from the project bar and files opened via the Open toolbar button. So this would not close a file that belongs to a project that is being closed but was opened via the toolbar. It is much more complex than the other proposal, so it is up for discussion if this additional complexity is worth it. I am not sure, honestly, as this seems like an edge case.

jeremypw:
@igordsm  Yes, it seems unlikely that a user would open a project file manually when it easier to open it from the sidebar. Even then, if you close a project folder you are kind of saying "I don't want to work on the files in this project anymore" so if one stays open it could be unexpected.  So personally I don't think it is worth the extra complexity.

One refinement that could be useful is a context menu option for expandable folders "Close all documents in this folder" but that is for another PR.

igordsm:
Closing this in favour of #833 .

PR #783, opened 2020-03-20
Vim emulation: fix 0 command

mcclurgm:
Part of splitting #633. Based on the implementation of the `I` command and gnome-builder.

Note, this does not work with selection. It simply removes the selection entirely. But as far as I can tell, the plugin does not implement selection in any way, so I don't believe this is a regression.


jeremypw:
This seems to move the cursor down a line - not to the beginning of the line (which I think is what is expected from https://devhints.io/vim).

plugins/vim-emulation/vim-emulation.vala
```diff
@@ -212,7 +212,11 @@ public class Scratch.Plugins.VimEmulation : Peas.ExtensionBase, Peas.Activatable
             case Gdk.Key.Home:
             case Gdk.Key.@0:
                 if (number == "") {
-                    view.move_cursor (Gtk.MovementStep.DISPLAY_LINES, 1, false);
+                    var buffer = view.buffer;
+                    Gtk.TextIter insert;
+                    buffer.get_iter_at_mark (out insert, buffer.get_insert ());
+                    insert.set_line_offset (0);
+                    buffer.place_cursor (insert);
```
jeremypw, line 219: This can simply be `view.move_cursor (Gtk.MovementStep.DISPLAY_LINE_ENDS, -1, false);`

jeremypw (changes requested):
See inline comment

plugins/vim-emulation/vim-emulation.vala
mcclurgm: That seems like it should, but it ignores whitespace, so it moves to after any indentation. `0` is supposed to move the the first character of the line, regardless of indentation. That's one of the issues I was trying to fix with this PR. The whitespace-respecting command is `^`.

jeremypw: Oh, yeah - sorry.  In that case you can fix it with fewer lines with:
```
                    view.move_cursor (Gtk.MovementStep.PARAGRAPH_ENDS, -1, false);
                    view.move_cursor (Gtk.MovementStep.DISPLAY_LINE_ENDS, -1, false);
```

This seems to give the same behaviour as the Vim app.

mcclurgm: Hmm, looks good in my testing. Makes sense with the docs too. That's a nice improvement, thanks!

jeremypw (approved):
Lets go!

jeremypw:
Need someone to fix the flatpak CI before this can be merged.

PR #784, opened 2020-03-25
Vim emulation: move by paragraph instead of display line

mcclurgm:
Part of splitting #633.

Line-based movement commands (eg `j`, `k`, `0`, `A`, `$`) in vi/vim work by newline-ended lines, not visual lines. [Vim docs](https://vimhelp.org/motion.txt.html#linewise). This is an issue when line wrapping is enabled. This PR fixes the behavior in the plugin to move by newline-ended lines, instead of visual lines.

Movement by visual lines works by appending `g` to the front (as in `gj`, [docs](https://vimhelp.org/motion.txt.html#g%5E)), which is a command we could implement. But I think that's a separate PR, since it's much more complex.


jeremypw (dismissed):
Makes sense

jeremypw:
@mcclurgm You need to fix a conflict on this before I review.

mcclurgm:
Done. Sorry for the delay.

jeremypw (changes requested):
In this PR, `j` and `k` commands do not seem to work the same as in `vim`.  In `vim`, in a document with paragraphs containing word wrapped lines and more than one sentence these commands move up and down between the same offset within adjacent paragraphs (if there is a blank line between paragraphs then it moves to the start of that blank line).  In this PR, the commands always move to the beginning or end of the next or previous paragraph.

`0`, `$` and `A` work the same as in `vim`.

mcclurgm:
Oh yeah, looks like `MovementStep.PARAGRAPHS` behaves kind of weirdly. It also skips blank lines, but only some of the time. I guess I'll have to do some more looking at it and probably find some sort of workaround. Do you want me to remove those changes so we can merge the changes that work within a line, which seem to work, and then come back to moving between lines?

jeremypw:
Yes, please do that.

mcclurgm:
Done

jeremypw (approved):

PR #785, opened 2020-03-29
Vim emulation: add "o", "O" commands

mcclurgm:
Implements a basic form of the `o` and `O` commands in vim. It adds a new line before or after the current, navigates to that line, and enters insert mode.

As with #712 (and the other insertion commands), this doesn't properly work with numbers. It won't repeat the insert multiple times, only once.


plugins/vim-emulation/vim-emulation.vala
```diff
@@ -205,6 +205,33 @@ public class Scratch.Plugins.VimEmulation : Peas.ExtensionBase, Peas.Activatable
...
+                    return false;
+                }
+                mode = Mode.INSERT;
+                debug ("Vim Emulation: INSERT Mode!");
+
+                view.move_cursor (Gtk.MovementStep.PARAGRAPH_ENDS, 1, false);
+                var buffer = view.buffer;
+                Gtk.TextIter insert;
+                buffer.get_iter_at_mark (out insert, buffer.get_insert ());
+                buffer.insert (ref insert, "\n", 1);
```
jeremypw, line 219: Can't this be replaced by view.insert_at_cursor ("\n"); ?

plugins/vim-emulation/vim-emulation.vala
```diff
@@ -205,6 +205,33 @@ public class Scratch.Plugins.VimEmulation : Peas.ExtensionBase, Peas.Activatable
...
+                    return false;
+                }
+                mode = Mode.INSERT;
+                debug ("Vim Emulation: INSERT Mode!");
+
+                view.move_cursor (Gtk.MovementStep.PARAGRAPH_ENDS, -1, false);
+                var buffer = view.buffer;
+                Gtk.TextIter insert;
+                buffer.get_iter_at_mark (out insert, buffer.get_insert ());
+                buffer.insert (ref insert, "\n", 1);
```
jeremypw, line 232: Also this.

jeremypw (changes requested):
This works as expected but some simplification is possible.

plugins/vim-emulation/vim-emulation.vala
mcclurgm: Yep, seems to work for me. I don't think I knew about that before. Thanks!

mcclurgm:
Ok, I fixed your requests. There's one big issue remaining here that I just discovered, that the `MovementStep.PARAGRAPH_START` goes to the beginning of the line excluding whitespace. So the shift-O command needs some more work. I'll see if I can fix that.

mcclurgm:
Since the effect is pretty much the same as inserting a newline after the `0` motion, I'm going to wait until we figure that out and then follow that solution here.

plugins/vim-emulation/vim-emulation.vala
```diff
@@ -205,6 +205,27 @@ public class Scratch.Plugins.VimEmulation : Peas.ExtensionBase, Peas.Activatable
...
+                view.insert_at_cursor ("\n");
+                break;
+            case Gdk.Key.O:
+                if (mode == Mode.INSERT) {
+                    return false;
+                }
+                mode = Mode.INSERT;
+                debug ("Vim Emulation: INSERT Mode!");
+
+                view.move_cursor (Gtk.MovementStep.PARAGRAPH_ENDS, -1, false);
```
jeremypw, line 225: I think you can insert `view.move_cursor (Gtk.MovementStep.DISPLAY_LINE_ENDS, -1, false);` here to ignore whitespace.

mcclurgm:
One more thought I just had. This doesn't respect the current line ending setting. I guess I'll have to fix that too…

jeremypw:
Need someone to fix the flatpak CI before this can be merged.

jeremypw (approved):

jeremypw:
I think any issues with non-standard line-endings can be dealt with, if required, later. Code is focused on code development on Linux.

Issue #768, opened 2020-02-19
Use a single gsetting key for window coords/size


danrabbit:
It's wasteful to use 4 keys when we can use 2

PR #794, opened 2020-04-22
Use a single gsetting key for window coords/size (#768)

tmullender:
Lockdown has led me to play with elementary and vala and a stab at one of the Bitesize issues

This may be utter nonsense but it seems to work for me so far, I'm not sure what to do about migrating an existing schema though?

Any feedback appreciated

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -21,7 +21,7 @@
 */
 
 namespace Scratch {
-    public SavedState saved_state;
+    public GLib.Settings saved_state;
     public Settings settings;
     public ServicesSettings services;
     public GLib.Settings privacy_settings;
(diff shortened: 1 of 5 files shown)


src/Application.vala
```diff
@@ -55,7 +55,7 @@ namespace Scratch {
 
             // Init settings
             default_font = new GLib.Settings ("org.gnome.desktop.interface").get_string ("monospace-font-name");
-            saved_state = new SavedState ();
+            saved_state = new GLib.Settings (Constants.PROJECT_NAME + ".saved-state");
```
jeremypw, line 58: As the schema path is hard-coded in the gschema.xml file it may as well be hard-coded here.

jeremypw:
Thanks for you contribution!  This looks pretty much like how its been done in Files and Terminal - were you using those as a model?  I don't recall any particular steps being taken when @danrabbit made similar changes to other apps.

jeremypw:
Seems to work OK - all that happens is that the first time it is used, the previous window size/position info is lost. This does not seem worth worrying about.

tmullender:
Thanks for the feedback.  Yep, I spent a while looking for a way to use granite settings with tuples but then spotted it was deprecated so took a look at what other apps were doing and followed their example.

cassidyjames:
Fixed #768

PR #802, opened 2020-04-28
SourceView: Scroll past end

### LINKED ISSUE #799: Scroll past end of file
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature
Working at the end of a file may be uncomfortable for some users. I prefer the line I'm working on to be in the centre of the screen. Having an option to scroll beyond the last line would make experience much better. Without such feature I often have to resort to adding many empty lines to the end of the file, which just doesn't feel right.

image

**Existing work**

This feature exists in many editors such as Visual Studio Code (on by default), Atom, Notepad++, etc.

Scrolling beyond last line in VSCode:
image

cassidyjames:
Fixes #799.

- Assumes we want at least three lines visible at all times (not configurable)
- Always on (not configurable)
- Throttled for performance (noticeably lags on resize without a timeout)

### DIFF
--- a/src/Widgets/SourceView.vala
+++ b/src/Widgets/SourceView.vala
@@ -147,6 +147,11 @@ namespace Scratch.Widgets {
             });
 
             populate_popup.connect_after (on_context_menu);
+
+            size_allocate.connect ((allocation) => {
+                // TODO: Throttle this for performance
+                bottom_margin = calculate_bottom_margin (allocation.height);
+            });
         }
 
         private bool get_current_line (out Gtk.TextIter start, out Gtk.TextIter end) {
@@ -414,6 +419,15 @@ namespace Scratch.Widgets {
             }
 
             menu.show_all ();
+        }
+
+        private int calculate_bottom_margin (int height_in_px) {
+            const int LINES_TO_KEEP = 3;
+            const double PT_TO_PX = 1.6666; // Normally 1.3̅, but this accounts for line-height
+
+            double px_per_line = Application.instance.get_last_window ().get_current_font_size () * PT_TO_PX;
+
+            return (int) (height_in_px - (LINES_TO_KEEP * px_per_line));
         }
 
         void on_mark_set (Gtk.TextIter loc, Gtk.TextMark mar) {
--- a/src/Widgets/SourceView.vala
+++ b/src/Widgets/SourceView.vala
@@ -29,11 +29,11 @@ namespace Scratch.Widgets {
 
         private string font;
         private uint selection_changed_timer = 0;
+        private uint size_allocate_timer = 0;
         private Gtk.TextIter last_select_start_iter;
         private Gtk.TextIter last_select_end_iter;
 
-        // Pause after end user highlighting to confirm select,in ms
-        private const uint SELECTION_CHANGED_PAUSE = 400;
+        private const uint THROTTLE_MS = 400;
 
         public signal void style_changed (Gtk.SourceStyleScheme style);
         public signal void selection_changed (Gtk.TextIter start_iter, Gtk.TextIter end_iter);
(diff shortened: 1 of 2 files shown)


src/Widgets/SourceView.vala
```diff
@@ -416,6 +421,15 @@ namespace Scratch.Widgets {
             menu.show_all ();
         }
 
+        private int calculate_bottom_margin (int height_in_px) {
+            const int LINES_TO_KEEP = 3;
+            const double PT_TO_PX = 1.6666; // Normally 1.3̅, but this accounts for line-height
+
+            double px_per_line = Application.instance.get_last_window ().get_current_font_size () * PT_TO_PX;
```
cassidyjames, line 428: Should this be cached and only updated on change so we don't have to get it every time?

src/Widgets/SourceView.vala
```diff
@@ -416,6 +427,15 @@ namespace Scratch.Widgets {
             menu.show_all ();
         }
 
+        private int calculate_bottom_margin (int height_in_px) {
```
donadigo, line 430: This function can be static.

src/Widgets/SourceView.vala
```diff
@@ -147,6 +147,17 @@ namespace Scratch.Widgets {
             });
 
             populate_popup.connect_after (on_context_menu);
+
+            size_allocate.connect ((allocation) => {
+                // Throttle for performance
+                if (size_allocate_timer == 0) {
+                    size_allocate_timer = Timeout.add (THROTTLE_MS, () => {
```
donadigo, line 154: I'm not really sure if the timer here is a good solution. Calculating the botom margin shouldn't be expensive at all in this case. Does adding this code in without the timer cause any lags / hiccups?

cassidyjames: @donadigo yes, removing this creates noticable lag on resizing.

src/Widgets/SourceView.vala
cassidyjames:
```suggestion
        private static int calculate_bottom_margin (int height_in_px) {
```

src/Widgets/SourceView.vala
```diff
@@ -416,6 +427,15 @@ namespace Scratch.Widgets {
             menu.show_all ();
         }
 
+        private static int calculate_bottom_margin (int height_in_px) {
+            const int LINES_TO_KEEP = 3;
```
danrabbit, line 431: Hm I wonder if this is low. What about like 10-15 or something?

cassidyjames: I was looking at prior art, and it looks like VS Code does 1, Sublime Text does 1, and Atom does 3. I guess the idea is if people really prefer to write pretty close to the top of the window… :shrug:

danrabbit: Ah okay if there's prior art then ignore me

jeremypw:
I don't quite understand why the bottom margin needs to be recalculated on size-allocate. If it is fixed at exactly 3 lines then it should only only change if the line height changes (?)

davidmhewitt:
The bottom margin needs to be the height of the source view minus 3 lines worth of height, so that 3 lines can always be in view.

jeremypw:
I though Gtk.SourceView handled that automatically - it seemed to when I tried it.  Note: `bottom_margin` is different from `margin_bottom` - it is internal. If the bottom_margin of the SourceView => its allocation - 3 lines that should almost double its height request - unless I am completely confused.

davidmhewitt:
>  I thought the problem was that there was no extra lines at the end of the text

Correct, but more specifically we want to be able to scroll enough so that we can bring at least the last 3 lines of the file to the top of the scrolled area. 

If it helps to think of it in terms of lines, we don't want 3 blank lines on the bottom, we want `(total number of visible lines - 3)` blank lines on the bottom, so that the bottom 3 actual content lines can be the top 3 visible lines.

Maybe try out the behaviour of this branch so you can see what we're trying to get at. 

I don't particularly like having to connect up to the size-allocate here either, but I can't think of a more clever way to achieve the same effect.

However, I think this probably behaves a bit strangely for files that only have a few lines, you're right that it will probably almost double the allocation in this case and leave you with a scrollbar for a document that only has a couple of lines. So perhaps it needs to be a bit more smart.

jeremypw:
>Correct, but more specifically we want to be able to scroll enough so that we can bring at least the last 3 lines of the file to the top of the scrolled area.

Oh, in that case you are right, sorry.

jeremypw:
Geany just has one blank line at the end.

cassidyjames:
@jeremypw yeah the original issue was comparing Code to other popular modern code editors like VS Code, Atom, and Sublime Text.


While exploring those editors, I found that VS Code and Sublime Text both allow you to scroll the bottom line all the way to the top of the view (i.e. they have margin equal to the height in lines minus 1 line). Atom behaves similarly but with 3 lines. So that's the behavior I and others are used to and am emulating here.

I agree that we should follow-up and special case files that are shorter than the SourceView allocation.

PR #805, opened 2020-05-02
Put full document path in tab tooltip

### LINKED ISSUE #804: Full file path is currently not obtainable; would be useful to show in mouse-over pop-up.
Hello! Other text editors (such as Geany) show the full path of the file being edited when the mouse passes over that file's tab; but in Code, all I see when I mouse-over a tab is the exact same filename that's already displayed on the tab. I've been using Code only for a short time, yet have already run into several instances when I was unsure of exactly where the file in a particular tab had been loaded from (particularly if Code had restored the tabs from the previous session). This issue is compounded by the nature of the Save As dialog, which does not default to opening where the original file was located (which would be tremendously useful). I've been through all the preferences, dug around with dconf-editor, and generally tried my best to see if I had something set up wrong, but apparently not. Could you add a preference to show the full-path filename in the mouse-over pop-up? Thanks for your consideration of this issue!

jeremypw:
Fixes #804 

Requires https://github.com/elementary/granite/pull/388

CI will fail until Granite PR is merged and available.

https://github.com/elementary/granite/pull/388 now merged.


jeremypw:
I guess this failing flatpak CI because the wrong Granite version is being used.

Marukesu (approved):

cassidyjames:
This does what it says and seems straightforward. My only question for @elementary/ux is if we want to attempt to clean up the path string at all, or if the default formatting (e.g. `/home/cassidyjames/Projects/elementary/code/meson.build`) is fine.

cassidyjames (approved):
Previous comment can be addressed as a follow-up if needed.

Issue #761, opened 2020-02-01
Terminal Icon Missing in Header when no files open


marbetschar:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug

I'm no longer able to start the Terminal extension from within Code. The header icon is not shown anymore and setting the extension to inactive and reactivate it afterwards does not help.

I think what lead to this issue is that I closed the Terminal in Code using the shortcut `Control + w`, altough I'm not sure i really did that.

### To Reproduce

1. Open Code
2. Go to Settings > Preferences > Extensions
3. Enable `Terminal`
4. Open the Terminal by clicking on the Terminal icon in the header
5. Focus the Terminal, so you're able to input text
6. Press `Control +w` - the Terminal closes.
7. Close Code
8. Open Code
9. The Terminal Icon is gone

### Expected behavior

Pressing `Control + w` in the Code Terminal should do the same thing as the Terminal Icon in the header bar does: hide the Terminal.

marbetschar:
Doh! Nevermind, issue is solved....

Terminal Icon is there, but it is only shown when I open a File in Code. Shouldn't the icon be there all the time?

PR #806, opened 2020-05-02
Removing toggling of terminal button visibility for #761

tmullender:
Meant as a fix for #761 

This makes the terminal button visible once the plugin is enabled irrespective of whether a document is open or not.

This could have an impact on other plugins as the bottom bar is now visible when no documents are selected.  I have not found any impacts on the plugins that are part of this repository, are there plugins available that are not part of this repo?


jeremypw:
I believe it is intended that the terminal plugin will be removed in the future in favour of more focused build/install functionality.

I am not sure why you would need to toggle the terminal when there are no documents open anyway.  In this PR, the bottombar closes when switching to the welcome page so you have to toggle it on again to see it.

I can see there is an inconsistency in that the search icon remains visible when the welcome page is shown although it is insensitive.

Flagging for design team comment.

tmullender:
Thanks Jeremy.  I played with making the button insensitive when switching to the welcome page but it felt wrong to close the terminal when closing a document.  I'll wait for the design team's comments

jeremypw:
@tmullender You will probably need to resolve the conflicts before the UX team will review.

danrabbit (approved):
Sorry for taking so long to review! Fixed conflicts and lgtm. Thanks for this fix!

PR #810, opened 2020-05-05
Add "Set Path from Document" menu option to terminal plug

### LINKED ISSUE #735: Terminal: Set path to current document menu option
I have searched open and closed issues for duplicates.

## Feature
A quick way to cd to the directory containing the current document would be useful.

**Describe the solution you'd like**
Add a menu option "Set path to current document" to the terminal context menu.

**Existing work**
Geany has this feature

**Describe alternatives you've considered**
Smart path following could be an option but is difficult to get right for everyone. Some way of overriding it would still probably be necessary.

jeremypw:
Fixes #735 

I am aware that terminal plug may become redundant in the future, but it is unclear how long that will take. This is a fairly simple feature that may be useful in the meantime.

### DIFF
--- a/plugins/terminal/terminal.vala
+++ b/plugins/terminal/terminal.vala
@@ -29,6 +29,7 @@ public class Scratch.Plugins.Terminal : Peas.ExtensionBase, Peas.Activatable {
     Gtk.Notebook? bottombar = null;
     Scratch.Widgets.HeaderBar? toolbar = null;
     Gtk.ToggleToolButton? tool_button = null;
+    Gtk.MenuItem current_dir_menu_item;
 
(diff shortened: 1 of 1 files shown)


cassidyjames:
I don't think I like the phrasing of this… is it effectively just navigating the Terminal to the same folder as the current file?

jeremypw:
@cassidyjames Yes it sets the terminal path to the folder the current document is in.  The wording was taken from "Geany".

jeremypw:
Needs a suggestion for a better but concise, menu option description.

jeremypw:
Terminal plug will probably be deprecated so no point in keeping this around

PR #815, opened 2020-05-14
Handling detached HEAD when displaying git info

### LINKED ISSUE #801: Impossible to change git branch when in "HEAD detatched" state
If you get into a HEAD detached state in a project, the Branch menu goes away, making it impossible to change branches. I'm not sure how we should handle this state, though.

tmullender:
Potentially fixes #801 at a superficial level

* Using first 8 digits of OId when HEAD is not a branch

* Displaying the branch menu item even when HEAD is not a branch


src/FolderManager/ProjectFolderItem.vala
```diff
@@ -162,10 +162,14 @@ namespace Scratch.FolderManager {
 
             try {
                 var head = git_repo.get_head ();
+                string? display_name = head.get_target ().to_string ();
```
jeremypw, line 165: `head.get_target ()` could return `null` so should check that before dereferencing with `to_string ()`

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -296,22 +300,23 @@ namespace Scratch.FolderManager {
...
                     sensitive = false;
                 }
 
-                if (branches == null || cur_branch == null) {
+                if (branches == null || cur_head == null) {
                     sensitive = false;
                 }
 
                 var change_branch_menu = new Gtk.Menu ();
+                var head_is_branch = cur_head.is_branch ();
```
jeremypw, line 319: `cur_head` could be `null` so need to check that before derefencing.

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -324,7 +329,7 @@ namespace Scratch.FolderManager {
                                 var branch_item = new Gtk.CheckMenuItem.with_label (branch_name);
                                 branch_item.draw_as_radio = true;
 
-                                if (branch_name == cur_branch.get_name ()) {
+                                if (head_is_branch && branch_name == (cur_head as Ggit.Branch?).get_name ()) {
```
jeremypw, line 332: To avoid compile warning should use `((Ggit.Branch?)cur_head)` for the cast (with null check if required).  I know the `as` form is used elsewhere but it will get changed as some point and new code should follow current practice.

jeremypw (dismissed):
Thanks for the contribution!

This works as expected - just a few minor code issues.

Enhancements are possible but this can be iterated in further PRs according to design team input.

jeremypw:
DetachedHeadLabel

jeremypw (approved):
I think this can be merged and any UX issues addressed later

PR #819, opened 2020-05-23
Fixes esc key not hiding searchbar when search entry is empty

sixpounder:
The initial condition for not performing a search in case of empty search entry doesn't take into account the Esc key, which would not be able to hide the search bar.


danrabbit:
Can you provide specific steps to reproduce this issue or link an issue report? As far as I can tell, hitting the escape key already closes the search entry

sixpounder:
No specific steps required in my case: hitting Escape with an empty search entry doesn't hide the search bar.

But if it is just me and you cannot reproduce the issue I guess we can close this PR.

sixpounder:
Turns out it was my setup that, for some reason, had the escape key captured by another process preventing it being sent to Code. I didn't fully investigate it but it's clear that it was me and not a real issue. Closing this.

PR #832, opened 2020-06-24
Create a duplicated file that maintains Syntax Highlighting from the original

### LINKED ISSUE #817: Duplicated file loses Syntax Highlighting
Duplicating a file either by

1. Right clicking on it's tab and selecting `Duplicate` or

2. Just by double clicking on it's tab

creates a new file called `New Document` and the Syntax Highlighting resets to `Plain Text`. People mostly duplicate files as a matchstick version control system so most possibly they intend making some changes by keeping the previous code.

I believe it should keep all the state regarding that file including the extension hence Syntax Highlighting. For instance when `myFile.hs` is duplicated, the duplicate can be renamed like `myFile (copy).hs` at least in one of the above duplication choices. This would also be inline with what Files do when a file is copied pasted into the same directory.

franciol:
Resolves #817. It will create a duplicated file that maintains the highlights of the original. Duplicated file has (copy) on name, before first dot. If there are no dots, (copy) goes to the end of the filename.


src/Widgets/DocumentView.vala
```diff
@@ -134,6 +134,16 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
         return Path.build_filename (((Scratch.Application) GLib.Application.get_default ()).data_home_folder_unsaved, new_text_file);
     }
 
+    private string unsaved_duplicated_file_path_builder (string original_filename) {
```
jeremypw, line 137: The parameter is actually the original basename - the parameter name should reflect this.

src/Widgets/DocumentView.vala
```diff
@@ -134,6 +134,16 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
...
+        string[] parts = original_filename.split (".", 2);
+        if (parts.length == 1) {
+            return _("%s(copy)").printf (parts[0]);
```
jeremypw, line 140: This will just return a basename not a path?  I think you need a single exit point, building a path.   Otherwise the duplicate file will be created in whatever folder Code was run from.

src/Widgets/DocumentView.vala
```diff
@@ -134,6 +134,16 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
...
+        }
+        string new_text_file = _("%s(copy).%s").printf (parts[0], parts[1]);
+
+        return Path.build_filename (Application.instance.data_home_folder_unsaved, new_text_file);
```
jeremypw, line 144: This does not build now - use `((Scratch.Application) GLib.Application.get_default ())`

jeremypw:
Closed in favour of #899

PR #833, opened 2020-06-26
Close Open Documents from folder when closing folder

### LINKED ISSUE #576: Close open files from project when project folder is closed
Currently, when a project folder is closed from the sidebar, all open files from that project remain open in tabs. I would expect those tabs to close as well, since the project is no longer required.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65676584-close-open-files-from-project-when-project-folder-is-closed?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

franciol:
Fixes #576
FileView sends a signal to MainWindow so it can verify if there's an open file from that folder. 
MainWindow then closes all docs open from that folder.


src/MainWindow.vala
```diff
@@ -350,6 +350,22 @@ namespace Scratch {
                 }
             });
 
+            folder_manager_view.close_all_docs_from_path.connect ((a) => {
+                Scratch.Widgets.DocumentView? view = null;
+                if (split_view.is_empty ()) {
+                    view = split_view.add_view ();
```
jeremypw, line 356: Why add a view?  I think you need to just return here.

src/MainWindow.vala
```diff
@@ -350,6 +350,22 @@ namespace Scratch {
...
+                } else {
+                    view = get_current_view ();
```
jeremypw, line 358: This only works if you do not have a split view or if you have a split view and the correct view is focused.  Otherwise, if you close a projectfolder whose documents are open in the unfocused view, they do not get closed.

jeremypw:
@franciol I have noticed that there seems to be another previous open PR addressing the same issue - #782.  Your approach seems simpler though, so if you can fix the outstanding issues and if no disadvantages of your approach come to light, I would prefer it.

src/MainWindow.vala
franciol: Thanks for the feedback. I don't remember why i did this, probably just forgot to remove that part. This seems to be the cause for the second issue you mentioned.

jeremypw:
@tintou Do have any comments for this?

jeremypw:
Hmm, seems to have acquired some lint from somewhere ...

jeremypw:
Seems to have acquired a build - problem - investigating.

jeremypw:
@franciol  This PR needs updating due to changes in master (which merging did not fix).  `split_view` no longer exists.

Are you in a position to return this PR to working order?

jeremypw:
Closed in favour of #912

Issue #417, opened 2018-05-12
[REQUEST] Add Typescript Support


btd1337:
Is It possible to add Typescript language support?

jwinnie:
If #220 is implemented then TypeScript support should be easy

eyelash:
[This](https://github.com/jefferyto/gtksourceview-js-lang) might be useful.

eyelash:
Some basic TypeScript support (syntax highlighting) was [just merged](https://gitlab.gnome.org/GNOME/gtksourceview/merge_requests/63) into GtkSourceView.

chances:
Is this now moot given #837?

PR #837, opened 2020-07-07
port to gtksourceview-4

eyelash:
gtksourceview-4 was released more than 2 years ago and includes many updates to the language definitions.
https://developer.gnome.org/gtksourceview/stable/porting-guide-3-to-4.html
~Word completion seems to be broken somehow, so I mark this pull request as a draft for now.~ I did not install code after compiling it so it would use the plugins that were installed from the old version. After installing the new version, code completion works fine.


eyelash:
This should close #417.

eyelash:
@danrabbit I updated the REAME file. There are still a few more instances of `gtksourceview-3` in the code base that I don't know how to handle. For example in the `codecore.deps` file.

danrabbit:
I'm a little less familiar with how the library portion of code works, but I'd image we should update that too to match

PR #849, opened 2020-07-13
Remove split view

### LINKED ISSUE #22: why add and remove view with one disabled
Why do we have a button in the Menu for "Add new view" and "Remove current view" when only one of them is not disabled at a time?  We should either allow more than just two views or make there only be one menu item to toggle showing the side view (with the description changing).

Is there some reason for this UX decision?  It seems like a waste of space to me.

Launchpad Details: [#LP1390572](https://bugs.launchpad.net/bugs/1390572) SuperScript - 2014-11-07 18:15:36 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44841660-why-add-and-remove-view-with-one-disabled?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #22


cassidyjames:
Now that we can close the sidebar per-window, it makes tiling with the window manager much more friendly. So I'm :+1: with this now.

PR #852, opened 2020-08-02
Disconnect selected signals on tab change

davidmhewitt:
I was seeing some issues where changing tab would sometimes work for a second and then switch back to the original tab.  Let's disconnect the `item_selected` signals in the Folder Manager when we're manually changing the selection so we don't get into some weird loop.


danrabbit:
Thank you for finding the source of this!

PR #868, opened 2020-09-01
Remove Browser Preview plugin

### LINKED ISSUE #867: Consider removing browser preview extension
While this non-default extension can be handy in some limited cases where you're editing a self-contained HTML file, modern web development has long moved away from that paradigm. It feels like these days shipping an entire browser inside of Code is Not a Good Idea, unless it could be aware of all kinds of web frameworks and be far, far smarter than it is today. 

I think the best course would be to remove it unless/until someone wants to develop _and maintain_ something else—but even then, just use your favorite web browser and tile it side-by-side…

danrabbit:
Fixes #867


tintou (approved):

jeremypw:
Rather than removing this it might have been better to convert it to a markdown preview (there are libraries to convert markdown to HTML)

PR #869, opened 2020-09-01
MainWindow: Subclass Hdy.Window

danrabbit:


jeremypw (changes requested):
Round corners sure enough.  
As master is already incompatible with Hera no reason why this cannot be merged.
Just needs conflict resolving.

jeremypw (approved):

PR #870, opened 2020-09-02
Set window title to focused document

### LINKED ISSUE #863: Set window title to currently-focused file
If we're gonna use window titles in Alt+Tab window switching, this will be more important. But it's already weird in Plank when you have two Code windows both just titled "Code".

sebastianlay:
Fixes #863

This should already be it, right?


cassidyjames (approved):
Well that was simple! I wonder why the comment was there already, but no code was. Huh.

Issue #773, opened 2020-03-01
Fix deprecation warnings


worldofpeace:
## Describe the bug
When building with the versions of these relevant packages:
- GTK 3.34.14
- Vte 0.58.3
- libgee 0.20.3

is see these deprecation warnings.

```
../src/Widgets/SourceView.vala:222.13-222.25: warning: Gtk.Widget.override_font has been deprecated since 3.16
../src/Widgets/SourceView.vala:222.13-222.25: warning: Gtk.Widget.override_font has been deprecated since 3.16
../plugins/brackets-completion/brackets-completion.vala:264.19-264.39: warning: Gee.Map.contains is deprecated. Use Map.has_key
../plugins/pastebin/pastebin_dialog.vala:325.31-325.45: warning: Gtk.Dialog.get_action_area has been deprecated since 3.12
../plugins/terminal/terminal.vala:255.17-255.26: warning: Gtk.Menu.popup has been deprecated since 3.22
../plugins/terminal/terminal.vala:262.13-262.31: warning: Vte.Terminal.spawn_sync has been deprecated since 0.48
../plugins/word-completion/completion-provider.vala:165.32-165.84: warning: Gtk.SourceCompletionItem.new has been deprecated since 3.24
```

## Platform Information
OS: NixOS

- [ ] I'm using the latest version from git that I've manually compiled
- [x] I'm using the latest released stable version

jeremypw:
Now only getting the `spawn_sync` deprecation warning with

GTK 3.24.33
VTE 0.68
GEE 0.20.5

on OS7.1

PR #872, opened 2020-09-04
Fix issues with the Pastebin plugin

### LINKED ISSUE #871: Issues with the Pastebin plugin
There are a number of issues with the current state of the Pastebin plugin:

- Upload is not working (since the plugin uses an http url, which returns a 301 redirect)
- The icons on the buttons in the "Other formats" window are missing/broken
- The "Upload" window resizes itself when closing the "Other formats" window
- The share button and the "Upload to Pastebin" menu stay visible after disabling the "Pastebin plugin" in the preferences
(can be fixed by restarting Code)

sebastianlay:
This fixes all of the issues mentioned in #871, removes one warning from #773 and improves the overall aesthetics of this plugin.

Before:
before

After:
after

I also adjusted some of the translatable texts in the code and I am not sure whether there is anything else needed there to make them available for translation.

Fixes #871

### DIFF
--- a/plugins/pastebin/pastebin_dialog.vala
+++ b/plugins/pastebin/pastebin_dialog.vala
@@ -385,7 +385,7 @@ namespace Scratch.Dialogs {
             var cancel_button = (Gtk.Button)format_dialog.add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
             cancel_button.clicked.connect (cancel_button_clicked);
 
-            var select_button = (Gtk.Button)format_dialog.add_button (_("Select format"), Gtk.ResponseType.OK);
+            var select_button = (Gtk.Button)format_dialog.add_button (_("Select Format"), Gtk.ResponseType.OK);
             select_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
             select_button.clicked.connect (select_button_clicked);
 
--- a/plugins/pastebin/pastebin_dialog.vala
+++ b/plugins/pastebin/pastebin_dialog.vala
@@ -228,12 +228,6 @@ namespace Scratch.Dialogs {
             {"n", "zxbasic", "ZXBasic", ""}
         };
 
-        private const string CSS = """
-            list {
-                background-color: transparent;
-            }
-        """;
-
         public Scratch.Services.Document doc { get; construct; }
 
         private Gtk.Button close_button;
@@ -258,14 +252,6 @@ namespace Scratch.Dialogs {
         }
 
         construct {
-            var provider = new Gtk.CssProvider ();
-            try {
-                provider.load_from_data (CSS, CSS.length);
-                Gtk.StyleContext.add_provider_for_screen (Gdk.Screen.get_default (), provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
-            } catch (Error e) {
-                critical (e.message);
-            }
-
             name_entry = new Gtk.Entry ();
 
             var name_entry_l = new Gtk.Label (_("Name:"));
@@ -389,9 +375,12 @@ namespace Scratch.Dialogs {
             select_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
             select_button.clicked.connect (select_button_clicked);
 
-            var content_area = format_dialog.get_content_area ();
-            content_area.add (languages_scrolled);
+            var frame = new Gtk.Frame (null);
+            frame.add (languages_scrolled);
+            frame.margin = 6;
+            frame.margin_top = 0;
 
+            format_dialog.get_content_area ().add (frame);
             format_dialog.show_all ();
         }
 
--- a/plugins/pastebin/pastebin_dialog.vala
+++ b/plugins/pastebin/pastebin_dialog.vala
@@ -373,7 +373,7 @@ namespace Scratch.Dialogs {
 
                 languages_listbox.add (label);
             }
-            
+
             var languages_scrolled = new Gtk.ScrolledWindow (null, null);
             languages_scrolled.hscrollbar_policy = Gtk.PolicyType.NEVER;
             languages_scrolled.height_request = 250;
@@ -401,16 +401,13 @@ namespace Scratch.Dialogs {
 
         private void select_button_clicked () {
             var selection = languages_listbox.get_selected_row ();
-            if (selection != null)
-            {
+            if (selection != null) {
                 var label = (Gtk.Label)selection.get_child ();
                 var lang_name = label.label;
                 var lang_code = "";
 
-                for (var i=0; i < languages.length[0]; i++)
-                {
-                    if (languages[i, 2] == lang_name)
-                    {
+                for (var i=0; i < languages.length[0]; i++) {
+                    if (languages[i, 2] == lang_name) {
                         lang_code = languages[i, 1];
                         format_combo.append (lang_code, lang_name);
                         format_combo.set_active_id (lang_code);
--- a/plugins/pastebin/pastebin_dialog.vala
+++ b/plugins/pastebin/pastebin_dialog.vala
@@ -262,8 +262,9 @@ namespace Scratch.Dialogs {
 
             format_combo = new Gtk.ComboBoxText ();
 
-            var format_button = new Gtk.Button.from_icon_name ("view-more-horizontal-symbolic");
-            format_button.name = _("Choose Different Format");
+            var format_button = new Gtk.Button.from_icon_name ("view-more-horizontal-symbolic") {
+                tooltip_text = _("Choose different format")
+            };
             format_button.clicked.connect (format_button_clicked);
 
             //populate combo box
--- a/plugins/pastebin/pastebin_dialog.vala
+++ b/plugins/pastebin/pastebin_dialog.vala
@@ -335,7 +335,7 @@ namespace Scratch.Dialogs {
 
             close_button = (Gtk.Button)add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
             upload_button = (Gtk.Button)add_button (_("Upload to Pastebin"), Gtk.ResponseType.OK);
-            upload_button.get_style_context ().add_class ("suggested-action");
+            upload_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
 
             read_settings ();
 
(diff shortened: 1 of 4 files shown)


plugins/pastebin/pastebin_dialog.vala
```diff
@@ -228,29 +228,44 @@ namespace Scratch.Dialogs {
...
-        private Gtk.ListStore format_store;
+        private Gtk.Dialog format_dialog;
         private Gtk.Stack stack;
+        private Gtk.ListBox languages_listbox;
 
         public PasteBinDialog (Gtk.Window? parent, Scratch.Services.Document doc) {
             Object (
-                border_width: 5,
                 deletable: false,
+                resizable: false,
```
danrabbit, line 252: What's the reason for making this not resizable?

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -228,29 +228,44 @@ namespace Scratch.Dialogs {
...
         }
 
         construct {
+            var provider = new Gtk.CssProvider ();
+            try {
+                provider.load_from_data (CSS, CSS.length);
+                Gtk.StyleContext.add_provider_for_screen (Gdk.Screen.get_default (), provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+            } catch (Error e) {
+                critical (e.message);
+            }
```
danrabbit, line 267: In general, it's better to add the provider directly to the widgets you need instead of to the whole screen. This can cause unintended consequences for other widgets in the app.

But I think it would be more in keeping with the style of other system components to put this list inside of a Gtk.Frame instead of removing its background

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -261,7 +276,7 @@ namespace Scratch.Dialogs {
 
             format_combo = new Gtk.ComboBoxText ();
 
-            var format_button = new Gtk.Button.with_label (_("Others…"));
+            var format_button = new Gtk.Button.with_label (_("…"));
```
danrabbit, line 279: It definitely looks cleaner this way than with the label, but now it's a little inaccessible. I would suggest instead making this a `Button.with_image_from_icon_name` and using `view-more-horizontal-symbolic` for the icon. Then it will need a tooltip to remain accessible. Maybe "Other formats" or something descriptive like this

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -283,18 +298,18 @@ namespace Scratch.Dialogs {
                 format_combo.set_active_id ("text");
             }
 
-            var expiry_combo_l = new Gtk.Label (_("Expiry time:"));
+            var expiry_combo_l = new Gtk.Label (_("Expiration:"));
```
danrabbit, line 301: Haha yep, this is definitely better

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -283,18 +298,18 @@ namespace Scratch.Dialogs {
...
 
             private_check = new Gtk.CheckButton.with_label (_("Keep this paste private"));
-            private_check.margin_top = 12;
+            private_check.margin_bottom = 15;
 
             var grid = new Gtk.Grid ();
-            grid.column_spacing = 6;
-            grid.row_spacing = 12;
+            grid.column_spacing = 10;
+            grid.row_spacing = 8;
```
danrabbit, line 312: For spacing we always use multiples of 3 (unless you're trying to account for some off-by-one issue). Column spacing should be 6px and row spacing should be 12px typically with double spacing (24px) between sections. So the previous values here were correct

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -315,28 +330,23 @@ namespace Scratch.Dialogs {
             stack.add (grid);
             stack.add (spinner);
 
-            var content_area = get_content_area () as Gtk.Box;
+            var content_area = get_content_area ();
             content_area.add (stack);
```
danrabbit, line 334: If you wanted to simplify this further, we probably don't need to cache `content_area` since we don't use it again
```suggestion
            get_content_area ().add (stack);
```

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -315,28 +330,23 @@ namespace Scratch.Dialogs {
...
-
-            var cancel_button = new Gtk.Button.with_label (_("Close"));
-
-            var action_area = get_action_area () as Gtk.Box;
-            action_area.margin_top = 7;
-            action_area.add (cancel_button);
-            action_area.add (send_button);
+            close_button = (Gtk.Button)add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
+            upload_button = (Gtk.Button)add_button (_("Upload to Pastebin"), Gtk.ResponseType.OK);
+            upload_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
```
danrabbit, line 338: Nice improvement!

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -345,67 +355,68 @@ namespace Scratch.Dialogs {
...
-            format_others_win.title = _("Other formats");
-            format_others_win.set_default_size (250, 300);
-
-            format_others_view = new Gtk.TreeView ();
-            format_others_view.set_headers_visible (false);
+            format_dialog = new Gtk.Dialog ();
+            format_dialog.deletable = false;
+            format_dialog.resizable = false;
+            format_dialog.border_width = 5;
+            format_dialog.title = _("More formats");
```
danrabbit, line 363: Titles should probably be in Title Case
```suggestion
            format_dialog.title = _("More Formats");
```

And I guess the tooltip for the button from the other dialog should match this title. But tooltips are sentence case :)

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -345,67 +355,68 @@ namespace Scratch.Dialogs {
...
+            }
 
-            var format_others_ok = new Gtk.Button.from_icon_name ("dialog-ok", Gtk.IconSize.BUTTON);
-            format_others_ok.clicked.connect (format_others_ok_clicked);
+            var languages_scrolled = new Gtk.ScrolledWindow (null, null);
+            languages_scrolled.hscrollbar_policy = Gtk.PolicyType.NEVER;
+            languages_scrolled.height_request = 250;
+            languages_scrolled.expand = true;
+            languages_scrolled.margin_start = languages_scrolled.margin_end = 10;
+            languages_scrolled.margin_bottom = 10;
```
danrabbit, line 382: While we're doing all this, might as well use latest code style
```suggestion
            var languages_scrolled = new Gtk.ScrolledWindow (null, null) {
                hscrollbar_policy = Gtk.PolicyType.NEVER,
                height_request = 250,
                expand = true,
                margin = 12,
                margin_top = 0
            };
```

And a little trick since we want all sides to have even margin except the top, we can set margin and then override that one side

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -345,67 +355,68 @@ namespace Scratch.Dialogs {
...
-            format_others_buttons.pack_start (format_others_ok);
+            var select_button = (Gtk.Button)format_dialog.add_button (_("Select Format"), Gtk.ResponseType.OK);
+            select_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
+            select_button.clicked.connect (select_button_clicked);
 
-            var format_others_box = new Gtk.Box (Gtk.Orientation.VERTICAL, 10);
-            format_others_box.pack_start (format_others_scroll);
-            format_others_box.pack_start (format_others_buttons);
+            var content_area = format_dialog.get_content_area ();
+            content_area.add (languages_scrolled);
```
danrabbit, line 393: Same as above
```suggestion
            format_dialog.get_content_area ().add (languages_scrolled);
```

danrabbit (changes requested):
I can confirm that this branch works as expected. Really nice job with the cleanup and the fixes!

I have just a few suggestions inline. Thanks for working on this!

plugins/pastebin/pastebin_dialog.vala
sebastianlay: I don't see the benefit here since there is no widget that would scale with the window.

plugins/pastebin/pastebin_dialog.vala
sebastianlay: You are right. Looks better this way, too!

plugins/pastebin/pastebin_dialog.vala
sebastianlay: I replaced it with the icon you suggested and set the "name" property on the button. That should do the trick, right?

plugins/pastebin/pastebin_dialog.vala
sebastianlay: Good to know! I reverted the change and adjusted the other spacings to match that.

plugins/pastebin/pastebin_dialog.vala
sebastianlay: Absolutely.

plugins/pastebin/pastebin_dialog.vala
sebastianlay: Already did that in another commit.

plugins/pastebin/pastebin_dialog.vala
sebastianlay: Thanks!

plugins/pastebin/pastebin_dialog.vala
danrabbit: The name property won't set a tooltip. You'll need to use the property `tooltip_text`.

plugins/pastebin/pastebin_dialog.vala
```diff
@@ -261,7 +262,8 @@ namespace Scratch.Dialogs {
 
             format_combo = new Gtk.ComboBoxText ();
 
-            var format_button = new Gtk.Button.with_label (_("Others…"));
+            var format_button = new Gtk.Button.from_icon_name ("view-more-horizontal-symbolic");
+            format_button.name = _("Choose Different Format");
```
danrabbit, line 266: I believe this is only used internally by Gtk. You'll want to use `tooltip_text`
```suggestion
            var format_button = new Gtk.Button.from_icon_name ("view-more-horizontal-symbolic") {
                tooltip_text = _("Choose different format")
            };
```

also, tooltip labels should use Sentence case. I know it's a lot to remember :sweat_smile:

danrabbit (changes requested):
This looks great! Just the one comment about the tooltip :)

plugins/pastebin/pastebin_dialog.vala
sebastianlay: Always learning.

danrabbit (approved):
Great work! Thank you very much :)

plugins/pastebin/pastebin_dialog.vala
sebastianlay: Okay, but I will leave that for another time.

PR #877, opened 2020-09-07
Add flatpak manifest

Marukesu:
create a flatpak manifest using the gnome Sdk and elementary BaseApp.


danrabbit:
Hey thank you for submitting this PR! Do you think it would be possible to add a Flatpak build step to CI as well so that we ensure this stays functional?

Marukesu:
sure, is this sufficient ?
```yaml
  flatpak:

      runs-on: ubuntu-latest

      container:
       image: elementary/docker:unstable

      steps:
     - uses: actions/checkout@v1
     - name: Install flatpak-builder
       run: |
         apt update
         apt install -y flatpak flatpak-builder
     - name: install flathub remote and runtimes
       run: |
         flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
         flatpak install -y flathub org.gnome.Platform//3.36 org.gnome.Sdk//3.36 io.elementary.BaseApp//juno-19.08
     - name: Build flatpak manifest
       run: flatpak-builder flatpak-build --repo=flatpak-repo --force-clean --ccache flatpak-build io.elementary.code.yml
```

danrabbit:
@Marukesu As long as that builds, works for me :)

Marukesu:
added the job on CI.

io.elementary.code.yml
```diff
@@ -0,0 +1,126 @@
...
+    config-opts:
+      - '--disable-gtk-doc'
+    cleanup:
+      - '/share/gtk-doc' # for some reason, it still install the html folder.
+    sources:
+      - type: archive
+        url: https://downloads.sourceforge.net/gtkspell/gtkspell3-3.0.10.tar.xz
+        sha256: b040f63836b347eb344f5542443dc254621805072f7141d49c067ecb5a375732
+    modules:
+      - name: intltool
```
danrabbit, line 104: Afaict this isn't used anymore so we shouldn't need it

danrabbit (changes requested):

io.elementary.code.yml
Marukesu: without it, gtkspell won't build

4jNsY6fCVqZv:
How do you want to deal with the situation that the GNOME runtime is hosted on Flathub, but Flathub is not a FLOSS compatible solution because proprietary applications are also distributed there as well?

danrabbit (approved):
Nice job. Builds and runs locally and CI works :+1:

PR #882, opened 2020-09-13
Fix window positioning via saved state

### LINKED ISSUE #881: Window positioning is off when restoring from saved state
When closing and reopening Code the window shifts a little to the right and bottom each time.

vjr:
Fix #881


jeremypw (approved):
Fixes the issue - thanks!

PR #883, opened 2020-09-14
Pane → Sidebar

danrabbit:


tintou (approved):
That's an API break but we don't have any guaranty on this

PR #890, opened 2020-10-02
Add Markdown plugin

### LINKED ISSUE #850: WYSIWYG-like keyboard shortcuts in Markdown
When editing a Markdown README file or blog post, it would be handy to be able to use some familiar/standard WYSIWYG or word-processor-like keyboard shortcuts for text formatting:

- Ctrl+B for bold (`**strong**`)
- Ctrl+I for italic (`_emphasis_`)
- Ctrl+K for link (`[text](url)`)

For a great implementation, pressing the key should insert the characters and put the cursor where the user would most likely type next (i.e. between the `**` or `_`, or in the `[]` of a link), while hitting the shortcut with selected text should wrap the text with the formatting characters. This is how GitHub—which users of Code are likely to be familiar with—handles it in their Markdown inputs. 

This might be best implemented in an optional plugin.

igordsm:
This fixes #850 . Add a Markdown plugin with the following shortcuts:

* Ctrl+b for bold
* Ctrl+Shift+I for italic
* Ctrl+K for links

See a demo below. All three works with and without a selection (and place the cursor where it makes sense).

PR-markdown


cassidyjames:
Ah, yeah Ctrl+I is the jump-to-line feature. Hrm. @danrabbit thoughts on that?

plugins/markdown-actions/markdown-actions.vala
```diff
@@ -0,0 +1,124 @@
...
+            current_source.key_press_event.connect (shortcut_handler);
+        } else {
+            current_source.key_press_event.disconnect (shortcut_handler);
+        }
+    }
+
+    private bool shortcut_handler (Gdk.EventKey evt) {
+        var control = (evt.state & Gdk.ModifierType.CONTROL_MASK) != 0;
+        var shift = (evt.state & Gdk.ModifierType.SHIFT_MASK) != 0;
+        if (control && !shift && evt.keyval == Gdk.Key.b) {
```
jeremypw, line 55: You should ideally also test for other possible shortcut modifiers:

`var other_mods = (evt.state & Gtk.accelerator_get_default_mod_mask () & ~Gdk.ModifierType.SHIFT_MASK & ~Gdk.ModifierType.CONTROL_MASK) != 0;`

and check this is false so that you know only Control and/or Shift is being pressed.

You can save some process by returning immediately if `event.is_modifier == 1` is true.

plugins/markdown-actions/markdown-actions.vala
```diff
@@ -0,0 +1,124 @@
...
+        plugins = (Scratch.Services.Interface) object;
+        plugins.hook_document.connect ((doc) => {
+            if (current_source != null) {
+                current_source.key_press_event.disconnect (shortcut_handler);
+            }
+
+            current_source = doc.source_view;
+            configure_shortcuts ();
+
+            current_source.notify["language"].connect (configure_shortcuts);
```
jeremypw, line 39: I am not sure this works with a proxy property - I'll check.  Probably need
`((Gtk.SourceBuffer)(current_source.buffer)).notify["language"].connect ()`

Might be worth keeping an (unowned) reference `current_buffer = (Gtk.SourceBuffer)(current_source.get_buffer ())` anyway rather than keep creating temporary references.

Does this connection need to be disconnected when doc changes?

jeremypw (commented):
See inline comments - review for functionality and regressions to follow.

jeremypw:
I think all the shortcuts should work with ` <Control><Shift>` otherwise its confusing whether or not you need to hold the shift key  down.

jeremypw (changes requested):
The code works as expected and no regressions were found.  Some coding issues are mentioned inline.

plugins/markdown-actions/markdown-actions.vala
```diff
@@ -0,0 +1,124 @@
...
+
+    public Object object { owned get; construct; }
+
+    public void update_state () {}
+
+    public void activate () {
+        plugins = (Scratch.Services.Interface) object;
+        plugins.hook_document.connect ((doc) => {
+            if (current_source != null) {
+                current_source.key_press_event.disconnect (shortcut_handler);
```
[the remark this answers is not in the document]
igordsm: Done.

plugins/markdown-actions/markdown-actions.vala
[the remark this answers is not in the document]
igordsm: Done.

igordsm: Done.

plugins/markdown-actions/markdown-actions.vala
igordsm: Should we change this?

igordsm:
@jeremypw Thank you for the reviews! I have addressed all but one (left a comment in the discussion above) and re-request a review. 

I actually like using `ctrl`+`shift` more, since it feels more natural to apply the markups after selecting text.

plugins/markdown-actions/markdown-actions.vala
jeremypw: No, its OK.  I seem to remember in the past this wouldn't work but it seems with more recent Vala it does.

jeremypw (dismissed):

jeremypw:
Works nicely now - thank!  But you will need to add an entry to the next release in the appdata file.  I'll reinstate the review request to the ux team.

igordsm:
By the way, the Flatpak CI is returning with a very non-specific error (`Failed to download sources: module gtksourceview: Failed to execute child process ?git? (No such file or directory)`) that appears to be unrelated to the last commit. Is it supposed to work?

igordsm:
I have created a standalone repository [https://github.com/igordsm/code-markdown-plugin](https://github.com/igordsm/code-markdown-plugin) that implements this PR and installs to the correct folders in Elementary 6. 

I started a discussion on third party plugins at #908 . I would be happy to hear your @jeremypw @cassidyjames thoughts about that and to make this a third party plugin distributed via AppCenter.

danrabbit (changes requested):
This seems to work as expected except with undo. It looks like for some reason undo history sees either 2 or 3 actions. For example:

1. Create some text `foo`
2. Highlight and Ctrl + Shift+ B for bold
3. Notice that the text is now `**foo**` as expected
4. Press Ctrl + Z for undo
5. Notice that the text is now `**foo` instead of `foo`

with links you'll notice that each bracket and the set of parentheses are treated individually

davidmhewitt:
> It looks like for some reason undo history sees either 2 or 3 actions

@igordsm If I remember correctly, the solution to this is to call `TextBuffer.begin_user_action ()` at the start of an "action" (making something bold), then do all of the insertions and deletions necessary, then call `TextBuffer.end_user_action ()`. The undo manager should then treat that as one action.

igordsm:
@davidmhewitt Thank you. Just done that and it works as expected. @danrabbit undo/redo now work as expected.

code-markdown-undo

danrabbit (approved):
Works as expected! Nice job

PR #892, opened 2020-10-02
Use CSS to change default font, not deprecated code.

### LINKED ISSUE #847: Use CSS to set font style
`Gtk.Widget.override_font` is deprecated

It looks like Dazzle has a method `dzl_pango_font_description_to_css` that would make this easier

jeremypw:
Fixes #847 

Ported utility function from libdazzle to convert Pango.FontDescription to CSS which is applied to the SourceView.

The original code did not handle Pango.FontMask.STYLE properly (used the "font-variant" property instead) so that was corrected.

What reference is required to the original C code?

Tested with a range of fonts chosen from the Preferences dialog.


src/Utils.vala
```diff
@@ -25,4 +25,112 @@ namespace Scratch.Utils {
...
+
+            sb.append ("font-weight: %i;".printf (weight));
+        }
+
+        if (Pango.FontMask.STRETCH in mask) {
+            var stretch = font_descr.get_stretch ();
+
+            switch (stretch) {
+                case Pango.Stretch.NORMAL:
+                    sb.append ("font-stretch: %s;".printf ("normal"));
```
tintou, line 88:
```suggestion
                    sb.append_printf ("font-stretch: %s;", "normal");
```

src/Utils.vala
```diff
@@ -25,4 +25,112 @@ namespace Scratch.Utils {
...
     }
+
+    /* Ported (with corrections and improvements) from libdazzle
+     * (https://gitlab.gnome.org/GNOME/libdazzle/-/blob/master/src/util/dzl-pango.c)
+     */
+    public string pango_font_description_to_css (Pango.FontDescription font_descr) {
+        var sb = new StringBuilder ("");
+        var mask = font_descr.get_set_fields ();
+        if (Pango.FontMask.FAMILY in mask) {
+            var family = font_descr.get_family ();
```
tintou, line 36:
```suggestion
            unowned string family = font_descr.get_family ();
```

danrabbit (approved):
afaict this works as expected. Nice job :)

PR #893, opened 2020-10-02
Fix plugins compile warnings

jeremypw:
Fixes most compile warnings arising from plugins

Something weird is happening with `Vte.Terminal.spawn_async ()` - the Vala compiler is producing code that fails the C compiler for some reason so the deprecated `Vte.Terminal.spawn_sync ()` will be addressed in a later PR.


tintou:
That's https://gitlab.gnome.org/GNOME/vala/-/issues/721

danrabbit (approved):

PR #894, opened 2020-10-11
List Edit Mode for Markdown

igordsm:
Adds new items automatically when editing lists in Markdown. See below.

list-mode

It has some limitations, though.

1. if an item is added in an ordered list, only the items after it are renumbered.
1. there is no support for splitting/merging ordered lists

I am far from being an expert in TextBuffer and TextIter, but I am willing to work on top of this to get it in good shape. 

This was created on top of #890 , so that PR should be considered first. When/if it is merged I can rebase this one so that only new commits are present.


hanaral:
Great work!

igordsm:
Now that #890 has been accepted, this can be reviewed. I have already done a rebase and only the correct commits are present in this PR. Works as shown in the GIF above. Some thoughts about the current implementation:

* unordered lists work quite well. 
* ordered lists are a bit more tricky. I did some work to fix numbering when adding items in the middle of a list, but splitting lists and deleting items do not recount. I am not sure if this is a big deal, but it is worth noting. 
* github's editor does not do number recounting on ordered lists. The rendered simply ignores all numbers but the first one. 

Anyways, this is ready for a round of review for both code and functionality.

danrabbit (approved):
Works as expected! Thanks for working on this. Sorry for taking a while to review :sweat_smile:

PR #910, opened 2020-11-22
Check that characters counted as indentation are whitespace

### LINKED ISSUE #909: Preserve indent plugin counts "actual" characters as indentation
## What Happened

The preserve indent plugin wrongly counts indentation. It counts every character on the current line as indentation, regardless of whether it's whitespace or not. This is fine if you only ever paste into blank lines, but not if you paste into existing lines.

## Expected Behavior

The actual whitespace before the current line should be the indentation of the lines pasted.

## Steps to Reproduce

0. Make sure that the preserve indent plugin is active.
1. Copy a few lines of text.
2. Paste them at the end of a non-empty line.
3. Every line after the first will be indented by the number of characters before the cursor, not the number of whitespace characters at the start.

Here's an example:
Copy 
```
World
And
Other
Things
```
Paste at the end of:
```
Hello|
```
(The `|` character is the insertion point.)
Result:
```
HelloWorld
     And
     Other
     Things
```
### Logs

n/a

## Platform Information

Compiled from git, Ubuntu 20.04.

mcclurgm:
Fixes #909

### DIFF
--- a/plugins/preserve-indent/preserve-indent.vala
+++ b/plugins/preserve-indent/preserve-indent.vala
@@ -66,7 +66,7 @@ public class Scratch.Plugins.PreserveIndent : Peas.ExtensionBase, Peas.Activatab
         int tabwidth = Scratch.settings.get_int ("indent-width");
 
         unichar ch = pos.get_char ();
-        while (pos.get_offset () < iter.get_offset () && ch != '\n') {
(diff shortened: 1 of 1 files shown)


jeremypw (approved):
Works as expected.  Overwrites any tab indentation with spaces but that appears to be a feature of master.

PR #911, opened 2020-11-22
Properly count one-line paste on last line

### LINKED ISSUE #904: Preserve whitespace plugin wrongly indents last line
## What Happened

When I paste content in the last line of a file, Code indents that line. It seems like the length of the indentation is the number of characters in the line before the pasted content. 

## Expected Behavior

Pasting mid-line should not produce any indentation.

## Steps to Reproduce

1. Copy some content.
2. Open a new file. Type a few characters on the last line of the document (including blank lines), and don't create a new line after it.
3. Paste the content anywhere in the line (except the first column, where you won't see the difference).
4. The line will become indented by the number of characters that appeared before the cursor before pasting.

### Logs

No output produced on paste, even with debugging enabled.

## Platform Information

about
Latest release version (3.4.1)

mcclurgm:
Fixes #904 

It seems that the behavior of `TextIter` at the last line of a document is a little odd. So our code for the preserve indentation plugin counts the line number of the start and end of the paste as the same. This seems to not be what we expect, so previously the code continued on to the loop that indents all lines following the first one. Then, once there, the `TextIter` couldn't iterate past the last line, so it silently picked the last line and indented that.

This branch fixes all of that by just double checking whether the paste is contained entirely on the last line of the document. If so, it exits the `increase_indent` just like the other cases where only one line is pasted. It does handle if you paste multiple lines onto the last line of the document, then those are properly indented.

I'm open to suggestions about how to handle the multi-line if statement. I'm not sure if this is the recommended code style and didn't find anything in the official guide.


danrabbit (approved):
AFAICT this works as expected. Nice little fix! Thanks :)

PR #811, opened 2020-05-05
Adding an explicit help option

tmullender:
Needed to avoid the exit(0) in OptionContext.print_help

Fixes #689 

This is my weekly attempt at learning a bit more vala et al.  Shout if this more annoying to you than helpful and I can just keep it to myself.

I'm not sure if I'm supposed to update the translatable strings or not, so I've separated out the commits for some clarity.  There seems to be quite a lot of changes generated, should I be trimming this down to what is required or is this just expected?

PR #913, opened 2020-11-24
Simplify cmdline option handling

### LINKED ISSUE #689: Running the command with the -h option closes the application
I was running Code, then i opened a new Terminal and wrote `io.elementary.code -h` to get the help about the command (i was looking for where i can report an issue), but it closed the Code's window.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/76480617-running-the-command-with-the-h-option-closes-the-application?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #689 

Supersedes #811 

This makes the option parsing much more compliant with the GLib recommended best practices and removes a lot of the manual parsing and error checking of arguments.

We already have error views per-tab (instead of one dialog for each file) when something fails to open, so let's use those.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -40,7 +40,7 @@ namespace Scratch {
             { "new-window", 'n', 0, OptionArg.NONE, null, N_("New Window"), null },
             { "version", 'v', 0, OptionArg.NONE, null, N_("Print version info and exit"), null },
             { "set", 's', 0, OptionArg.STRING, ref _app_cmd_name, N_("Set of plugins"), N_("plugin") },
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
This seems to solve the linked issue and all of the commandline options work as expected here

PR #924, opened 2021-01-11
FormatBar: fix margin in lang popover

danrabbit:
* Fix margins around searchbar
* Remove extra margin in radiobuttons
* Use menuitem element name instead of css style
* Proper GObject style
* Other code style


cassidyjames (approved):
Nice

PR #925, opened 2021-01-13
Update README.md

k1r0s:
make live easier

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,22 +5,30 @@
 
 ## Building, Testing, and Installation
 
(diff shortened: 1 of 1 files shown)


k1r0s:
tested on ARM64 ubuntu mate both in rasp 4, odroid c2

danrabbit:
Hey thanks for your PR, but these directions are intentionally distro agnostic and your directions don't work on elementary OS at all. This isn't a change we want to make

k1r0s:
my bad, im sorry

PR #926, opened 2021-01-17
Expand folder when opening an already open project

davidmhewitt:
Currently if you have a long list of projects in the sidebar and you try and open one that's already open, nothing happens (other than a warning in the terminal). How about we expand that already open folder so you can see where it is easier?


meisenzahl (approved):
LGTM and works as intended

cassidyjames (approved):

PR #930, opened 2021-01-22
HeaderBar: Correctly load color button states

cassidyjames:
We were incorrectly checking for `high-contrast` as a color scheme name in the switch statement, when it should have actually been `classic`; the fact that it was a group of radio buttons meant the mismatch selected the first radio button which just happened to be the right one, but this makes it more explicit in a couple of ways:

- Adds and uses consts for color scheme names so we can't mismatch them anymore
- Adds a hidden `color_button_none` that's never attached
- Defaults to selecting `color_button_none` when the switch statement doesn't have a match so we don't select the wrong color button e.g. if someone manually edits the GSettings


meisenzahl (approved):
LGTM

PR #931, opened 2021-01-22
GSchema: Correct whitespace

cassidyjames:
There were just a couple of spots using inconsistent newlines and four-space indentation instead of two.


meisenzahl (approved):
LGTM

Issue #939, opened 2021-02-06
Vala and Meson compilation controls


hanaral:
## Problem

Right now the official method of making elementary apps is via the native code editor, so there might as well be an easy way to compile elementary apps.

## Proposal

Add cool controls to start a compilation script with meson, maybe even with integrated features specifically for target flatpak runtimes, archs and other useful app-making stuff.

## Prior Art

elementary-ide used to have this:
![image](https://user-images.githubusercontent.com/58219504/107103317-25a36800-6815-11eb-94af-3e796d20c038.png)
Panic's sexy code editor:
![image](https://user-images.githubusercontent.com/58219504/107103342-3c49bf00-6815-11eb-86c3-7a2063f63184.png)
![image](https://user-images.githubusercontent.com/58219504/107103401-6b603080-6815-11eb-9784-9ca3227b3259.png)

PR #940, opened 2021-02-06
PostGtk4 - Feature: Add build and run controls for Flatpak based projects

meisenzahl:
Related to https://github.com/elementary/code/issues/939

It's been quite a long time since I played around with it. After merging `master` the logic is broken in some places.

Bildschirmfoto von 2021-02-06 09 20 34


hanaral:
ngl I think that haveing a small 'section' of the titlebar to have a n almost media-controls-esque play, stop and a button to open a  build settings menu in the sidebar or something. Panic's Nova is something with a smilar design in that it's a mac exclusive but also using native design to macOS, Code should be the same for elemetary OS.
I'm going to post some more screenshots in the main issue related to this

meisenzahl:
@hanaral I separated the run button into it's own *section*.

Bildschirmfoto von 2021-02-06 09 41 47

I have to implement some more logic to add a stop button.

hanaral:
> @hanaral I separated the run button into it's own _section_.

I meant something a bit more like this:
107103342-3c49bf00-6815-11eb-86c3-7a2063f63184

> I have to implement some more logic to add a stop button.

Maybe add dummy buttons for now since I don't see it being optimal to PR as-is (from a UX standpoint)

meisenzahl:
Okay let's keep it simple and focus on building, installing and running a project in this PR.

@hanaral collect features in your issue. So that we can split this up into smaller PRs.

jeremypw:
@meisenzahl Thanks for working on this - it will be useful. Not sure if it is still in progress and should be marked "draft"?
Some comments after an initial try:

 - "Run" button should change appearance (or show/hide) depending on whether a compilable project is loaded/focused
 - Need to be able to choose to build only or build and install
 - Tooltip should indicate what action will be taken on which project
 - "Run" button is not active unless you click on a file in the sidebar even if a suitable ("meson.build") file is loaded and highlighted in the sidebar.
 - The "Run" button is not active when the project is collapsed 
 - Maybe put "Build" and "Install" options in the project folder context menu?
 - Cannot see output unless running Code in a terminal

davidmhewitt:
I'm not entirely sold on the idea of installing projects systemwide from Code at all. In general, I avoid installing projects compiled from source as much as possible.

Given the fact that we're expecting 3rd party applications to be built as flatpaks in the near future, and flatpak has a much more consistent and predictable build process that supports any build system (meson, cmake, GNU make etc), I think that would be a more worthwhile use of time. And installing/running built flatpaks doesn't have so much potential to mess up the system.

jeremypw:
But Code is for development not for installing things *per se*. The default should definitely be to build, not install.  However, for testing some things need to be installed so that should be available.  I agree that installing into a VM or something other than a production machine is probably wise.

jeremypw:
>Given the fact that we're expecting 3rd party applications to be built as flatpaks in the near future, and flatpak has a much more consistent and predictable build process that supports any build system (meson, cmake, GNU make etc), I think that would be a more worthwhile use of time. And installing/running built flatpaks doesn't have so much potential to mess up the system.

Could you expand a bit more how this relates to using Code?

davidmhewitt:
I'm saying that this PR currently adds support for building and installing meson based projects. As you say, to test many of these, they have to be installed which you have to be aware of the consequences of.

Given that most elementary repos now have a Flatpak manifest, and all 3rd party apps that a lot of people probably use Code to develop will be Flatpak based too. It might be more worthwhile having this PR add support for building and running Flatpak packages.

It solves the problem of needing to install the apps (since they're sandboxed) and it has the bonus of being an easier packaging format to support in Code.

meisenzahl:
@jeremypw thanks for your comments! I will address the points.

I will focus in this PR on being able to build and launch a project through the UI. @davidmhewitt has already noted that system-wide install is not ideal, but necessary in many cases. So I'm open to whether we build the project, install system-wide and launch it, or use Flatpak.

meisenzahl:
## Current state

- Emit a signal for every command line output in `ProjectManager`
  - could be used to show it in a view
- Add build button
  - only build the project
- Add stop button
  - stop a process if the project is building or has started

Bildschirmfoto von 2021-02-07 08 08 32

I'm open to suggestions for the build icon.

meisenzahl:
> "Run" button should change appearance (or show/hide) depending on whether a compilable project is loaded/focused

@jeremypw I already tackled this but I could use help for the logic of when a project is activated or a file is selected.

jeremypw:
@davidmhewitt I agree that FlatPak building and installing needs to be implemented in Code as well.  I don't think it needs to be exclusively Flatpak oriented though. There will always be code that is not FlatPak (system level stuff).

To keep PR small the FlatPak features could be a separate PR.

davidmhewitt:
And the system level stuff almost always needs to be started in a specific way:

- wingpanel: got to kill the currently running instance twice so that gnome-session stops restarting it.

- gala: got to start with `gala --replace` to replace the currently running instance and have a way of starting another instance afterwards so you don't get stuck in a session without a window manager.

- daemons: got to kill the currently running instance

And so on...

Code isn't and can't be aware of these nuances for each system application. So personally I think having a run button for anything that isn't a Flatpak application is a recipe for disaster.

So for most system applications, this run button won't work anyway (because there's probably an already running instance), but it'll be silently overwriting the installed version of the packages with (potentially broken) ones in the background. Then the next time you boot the system, you have no window manager. We might understand what's going on there, but to put a big green run button in a user facing application that hides this complexity doesn't sit well with me.

jeremypw:
@davidmhewitt Yeah, I agree that any install capability needs to be fenced off and not just a single click away. Maybe with a custom pre-install command associated with each project.

I would support not implementing system install for now.  A button for initiating a build would still be convenient but without a way of seeing the output is of limited use.

To be honest, I mostly use Geany and do not use the buttons for building/installing - I do everything from the terminal anyway.  So perhaps we do need to think a bit more about who we are aiming this feature at and for what purpose.

Maybe in the future all projects will be built as Flatpaks from day one but I would have thought that most projects are started as native then Flatpak'd later.

Asking for @elementary/ux input.

hanaral:
I totally agree that virtualy everything to do with application, component or runtime development should be restricted to flatpak since Code should be able to function like Xcode in a way but instead specifically for elementary and AppCenter. I also think it would be nice to have web dev support tools too, since that is also another huge component of development in general.

src/MainWindow.vala
```diff
@@ -252,6 +263,12 @@ namespace Scratch {
             // Show/Hide widgets
             show_all ();
 
+            project = new Services.ProjectManager ();
```
jeremypw, line 266: I would call this `project_manager` (cf `folder_manager`).  The project manager is not itself a particular project.

src/Services/ProjectManager.vala
```diff
@@ -0,0 +1,223 @@
...
+ * You should have received a copy of the GNU General Public
+ * License along with this program; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA
+ *
+ * Authored by: Marius Meisenzahl <mariusmeisenzahl@gmail.com>
+ */
+
+public class Scratch.Services.ProjectManager : Object {
+    public string path { get; set; }
```
jeremypw, line 23: I would call this `project_path`.  I try and avoid generic public variable names.

src/Services/ProjectManager.vala
```diff
@@ -0,0 +1,223 @@
...
+    private Pid command_pid;
+
+    private FlatpakManifest? flatpak_manifest () {
+        try {
+            Dir dir = Dir.open (path, 0);
+            string? name = null;
+            while ((name = dir.read_name ()) != null) {
+                string f = Path.build_filename (path, name);
+
+                if (FileUtils.test (f, FileTest.IS_REGULAR) && f.has_suffix (".yml")) {
```
jeremypw, line 38: Do we need to restrict this to `yml` manifests?  What about `json`?

meisenzahl: Good idea 😃️

hanaral:
It's kinda out of scope for this issue, but any chance we can get a more rounded play button?
That way code could use a modified version of it with some kind of 'developer' symbol on it (a hammer as a badge maybe?)

meisenzahl:
Would benefit from https://github.com/elementary/code/issues/991

## Possible follow-up PRs

- Show output of build and run in a `TerminalView`

jeremypw (dismissed):
Some minor preliminary code comments.

Do the build/run button sensitivity check whether the selected project is a FlatPak?
Do the build/run button sensitivities update if the process is stopped externally?

meisenzahl:
@jeremypw thanks for your review. Good suggestions, which I have implemented straight away.

Yes, the build/run button sensitivities update if the process is stopped externally.

hanaral:
I just want to point out a couple issue that may seem small but are actually quite problematic:
- Using the sandboxed version of this branch doesn't, and afaik (currently), has no way of working due to the restrictions the command line 
- I've already opened a discussion for this, but at this point the ball needs to get rolling on whether this IDE remains the default text editor or not since the public beta is almost out.

✨️ Solutions (?) ✨️
1. Maybe try to create a small builtin server on the OS that can run the flatpak-builder and flatpak commands, which Code would ask for over DBus (maybe there is already something that can do this)
2. Potentially using Writer as a base, try replacing the default text editor with a notes app that can sync with DAV notes and also import/export RTF files

I understand that these may warrant their own issues, but this seems to be a big PR anyway

jeremypw:
It is definitely needed to throw some kind of dialog if the build fails, preferably containing the critical message from the output (I see you are capturing that already).

src/Services/ProjectManager.vala
```diff
@@ -0,0 +1,259 @@
...
+                    try {
+                        string content;
+                        FileUtils.get_contents (f, out content);
+
+                        var re_app_id = new Regex ("app-id:\\s*(?P<app_id>[A-Za-z0-9-\\.]+)");
+                        var re_command = new Regex ("command:\\s*(?P<command>[A-Za-z0-9-\\.]+)");
+
+                        var flatpak_manifest = new FlatpakManifest () {
+                            manifest = f,
+                            build_dir = Path.build_filename (project_path, "build")
```
jeremypw, line 55: It would be better to have a separate directory for flatpak builds to avoid deleting the meson build folder which is a nuisance.
```suggestion
                            build_dir = Path.build_filename (project_path, "flatpak-build")
```

src/Services/ProjectManager.vala
```diff
@@ -0,0 +1,259 @@
...
+                        string content;
+                        FileUtils.get_contents (f, out content);
+
+                        var parser = new Json.Parser ();
+                        parser.load_from_data (content, -1);
+                        var object = parser.get_root ().get_object ();
+
+                        return new FlatpakManifest () {
+                            manifest = f,
+                            build_dir = Path.build_filename (project_path, "build"),
```
jeremypw, line 84:
```suggestion
                            build_dir = Path.build_filename (project_path, "flatpak-build"),
```

jeremypw:
OK, I have now got a flatpak building OK (after managing to install the elementary Sdk and Platform). I note that there is no feedback for a successful build either, though.

One small bug I noticed is that the build/run/stop buttons are insensitive at startup even though there is only one project open and there is an active document.  Clicking on that documents source view or sidebar entry or switching tabs does not make the buttons sensitive.  Only when a different item is clicked in the sidebar do the buttons become sensitive.

src/Services/ProjectManager.vala
```diff
@@ -0,0 +1,259 @@
...
+public class Scratch.Services.ProjectManager : Object {
+    public string project_path { get; set; }
+    public bool is_running { get; set; }
+
+    public signal void on_standard_output (string line);
+    public signal void on_standard_error (string line);
+
+    private Pid command_pid;
+
+    public FlatpakManifest? flatpak_manifest () {
```
jeremypw, line 31: Better to use a verb phrase for functions
```suggestion
    public FlatpakManifest? get_flatpak_manifest () {
```

src/Services/ProjectManager.vala
```diff
@@ -0,0 +1,259 @@
...
+
+            return exit_status;
+        } catch (SpawnError e) {
+            warning ("Could not run command: %s\n", e.message);
+            return false;
+        }
+    }
+
+    private async bool build_project () {
+        var flatpak_manifest = flatpak_manifest ();
```
jeremypw, line 189:
```suggestion
        var flatpak_manifest = get_flatpak_manifest ();
```

src/Services/ProjectManager.vala
```diff
@@ -0,0 +1,259 @@
...
+                flatpak_manifest.build_dir,
+                flatpak_manifest.manifest
+            });
+        }
+
+        return false;
+    }
+
+    private async bool run_project () {
+        var flatpak_manifest = flatpak_manifest ();
```
jeremypw, line 203:
```suggestion
        var flatpak_manifest = get_flatpak_manifest ();
```

jeremypw:
I am getting an error if I try the "run" command even though the "build" command is successful.
`error: Build directory /media/jeremy/Shared/shared_data/Vala/gnonograms/.flatpak-builder/rofiles/rofiles-cDbytD not initialised, use flatpak build-init`

I get the same error running `flatpak-builder --run <build-dir> <manifest> <command>` in the terminal. If I run the `flatpak build-init` I get `error: Build directory ./flatpak-build already initialised` so I do not know what is going on ...

meisenzahl:
Added an error dialog

Screenshot from 2021-05-04 20 39 04

meisenzahl:
> I am getting an error if I try the "run" command even though the "build" command is successful.
> `error: Build directory /media/jeremy/Shared/shared_data/Vala/gnonograms/.flatpak-builder/rofiles/rofiles-cDbytD not initialised, use flatpak build-init`
> 
> I get the same error running `flatpak-builder --run <build-dir> <manifest> <command>` in the terminal. If I run the `flatpak build-init` I get `error: Build directory ./flatpak-build already initialised` so I do not know what is going on ...
> 
> If I use `flatpak build-init --update` then it succeeds but the run command still gives the same error.

This seems to be a shared mount of VirtualBox. This can lead to errors. Would be great if you build the project in the local file system.

jeremypw:
@meisenzahl It is not in VirtualBox although the source is on a different partition to the main system.  I have build and installed the flatpak OK in the past but I haven't tried using the --run command before.  I'll try moving the code onto the system partition and try again.

jeremypw:
The dialog box looks nice although it would be better if the details showed 80 characters as the messages can be lengthy.

jeremypw:
Not sure how hard it would be but maybe the secondary text could be taken from the final critical message in the output?

meisenzahl:
@jeremypw et voilà !

Screenshot from 2021-05-05 12 43 40

jeremypw:
Looks good to me! Need to see what the UX team says.  Hopefully the problem I have with the run command is due to my setup not this PR.  I'll test again when I have got manual build/run working.

jeremypw:
So using `flatpak-build --force-clean --install --user <build-dir> <manifest>` does not work if `<build-dir>` and `<manifest>` are on different partitions. At some point a hardlink is created - and this cannot be across partitions.  I thought in the past (about a year ago) I used the same setup successfully but perhaps I am misremembering.  

Using the same command with the mainfest and build directory on the same partition was successful and, moreover `flatpak run <command>` successfully launched the flatpak app.   Just need to get the `flatpak-builder --run` to work now ...

jeremypw:
OK so the problem with the `flatpak-builder --run` command was that the manifest contained `  - '--metadata=X-DConf=migrate-path=<schema>`.  Presumably this also requires the settings to be on the same partition as manifest.  When this line was removed sandbox running was successful :tada:

jeremypw:
I think when the build/run is stopped with the button the dialog should reflect that rather than indicate an error occurred.

jeremypw:
I am inclined to merge this sooner rather than later and iterate as required, because it intersects with my "chooser-project-headerbarcombo" PR which could make use of the Project Manager class introduced here. However, I wonder whether the "GitManager" class could be merged with the "ProjectManager" and/or MonitoredRepository class?

meisenzahl:
@jeremypw thank you so much for looking at this several times now :)

From my point of view, we should not filter the output in the error dialog by `ERROR`. When an error occurs, I as a developer need context what happened before and what led to this error.

I agree with you that there is still a lot to optimize UX-wise. I hope for feedback from @elementary/ux :)

I have now adjusted the logic so that the dialog is only displayed in case of an error.

meisenzahl:
> I am inclined to merge this sooner rather than later and iterate as required, because it intersects with my "chooser-project-headerbarcombo" PR which could make use of the Project Manager class introduced here. However, I wonder whether the "ProjectManager"class could be merged with the "GitManager" and/or MonitoredRepository class?

@jeremypw  would this also work if a project is not managed with git?

jeremypw:
@meisenzahl No doubt the manager class could be generalized to other version control systems if the need arose, however, as @danrabbit has recently emphasised, we should not overcomplicate things in advance of an actual use case (YAGNI principle).

jeremypw:
Current appearance:

### HeaderBar
BuildRunButtons

### Error dialog
FakeBuildError

jeremypw:
@davidmhewitt Do you have any further comments on this? Any reason not to merge and iterate?

Marukesu:
When running code in sandbox, we don't have access to the flatpak installations or to launch a subprocess on the host, we can use the `HostCommand` from `org.freedesktop.Flatpak.Development` (with `--talk-name=org.freedesktop.Flatpak`) thought, I prefer to not make assumptions about what the host have installed.

Also, if we are going to use `flatpak-builder --run`, we need to wait for flatpak-builder 1.0.12 since the bug with the `--metadata` on the manifest was fixed there.

jeremypw:
@Marukesu So we would need to provide an option to install for some flatpaks to run successfully?  I would regard this PR as the first step, providing a minimal safe facility. Is it useful as it is or do more options/advanced facilities need to be added first?

Marukesu:
> So we would need to provide an option to install for some flatpaks to run successfully?

Well, we can still make a copy of the manifest without the `--metadata` on it.

> Is it useful as it is or do more options/advanced facilities need to be added first?

In the actual state, this PR don't work at all with code sandboxed.

We need to add flatpak-builder to the manifest, and probably, allow access to the flatpak directories on host, flatpak-builder known when is running on a sandbox and uses the flatpak on host.

jeremypw:
@meisenzahl Is there anything you can do to address @Marukesu 's concerns?

meisenzahl:
@jeremypw sorry for the late reply.

I tried to integrate `flatpak-builder`. But unfortunately the dependency chain is very long 🙈️

I tried to follow the [Flatpak manifest of Gnome Builder](https://github.com/flathub/org.gnome.Builder/blob/master/org.gnome.Builder.json).

Unfortunately I already get errors when compiling the dependency `polkit`...

@Marukesu I would be very grateful if you could take a look at my implementation 😅️

jeremypw:
Flatpak makes everything so much more complicated ... :grimacing:

Marukesu:
@meisenzahl you need to add `rm-configure: true` in the polkit module, i don't know much about how make works to know why trough.

about the dependencies, flatpak-builder needs libyaml and flatpak, flatpak needs ostree, fuse, polkit and pyparser.

also don't forget to add to the finish-args:  `--filesystem=/var/lib/flatpak`, `--filesystem=~/.local/share/flatpak`, `--share=network` and `--talk-name=org.freedesktop.Flatpak`.

with that the manifest is buildable.

> Flatpak makes everything so much more complicated ...

we can use the dbus interface directly, but it means that the user need to have flatpak-builder on the host.

meisenzahl:
I'm not getting anywhere here at all, unfortunately 😕

@Marukesu I would be totally grateful if you could invest some time and try to build **Code** as a Flatpak.

Marukesu:
@meisenzahl the manifest should build now. however, `flatpak-builder --run` don't seems to work inside the sandbox (the same happens with the [flatpak-builder from flathub](https://github.com/flathub/org.flatpak.Builder)).

jeremypw (changes requested):
After correcting the issue with the private getter, this PR successfully built a flatpak of itself (took quite a long time). However, pressing "run" gave the error
 `error: Build directory /home/jeremy/Documents/Elementary/clones/meisenzahl/code/.flatpak-builder/rofiles/rofiles-MdFy9A not initialised, use flatpak build-init`

The flatpak was not installed but presumably this is intended.

This was running this PR as a native install. Not tested in a flatpak yet.

Need to think about showing progress somehow - preferably show terminal output.  Also about how to prevent the user editing source files in the middle of a build.  Not sure what closing Code in the middle of build will do either.

Presumably "Run" runs the previously build flatpak - but what if the files have been edited?  Either need to rebuild or maybe disable "run" until there is an up to date build.

Making progress!

Marukesu (commented):
Some comments about when running in sandbox.

meisenzahl:
> After correcting the issue with the private getter, this PR successfully built a flatpak of itself (took quite a long time). However, pressing "run" gave the error `error: Build directory /home/jeremy/Documents/Elementary/clones/meisenzahl/code/.flatpak-builder/rofiles/rofiles-MdFy9A not initialised, use flatpak build-init`
> 
> The flatpak was not installed but presumably this is intended.
> 
> This was running this PR as a native install. Not tested in a flatpak yet.

Yes, it is intended that the built app is not installed.

These rofiles errors only appear with some apps that require extended permissions.

> --disable-rofiles-fuse
> Disable the use of rofiles-fuse to optimize the cache use via hardlink checkouts.

Source: [Flatpak Builder Command Reference](https://docs.flatpak.org/en/latest/flatpak-builder-command-reference.html)

Maybe @marukesu can explain it better 😅

The logic also works as a Flatpak. I test the Flatpak version of Code with the following script:

```bash
#!/bin/sh

# fail on first error
set -e

APP=io.elementary.code

flatpak-builder --repo=repo --disable-rofiles-fuse build/flatpak ${APP}.yml --force-clean
flatpak build-bundle repo ${APP}.flatpak --runtime-repo=https://flatpak.elementary.io/repo.flatpakrepo ${APP} master
flatpak install --user -y ${APP}.flatpak
flatpak run ${APP}
```

> Need to think about showing progress somehow - preferably show terminal output.

There is the terminal plugin that we could use, but that is disabled by default. If I'm honest, I'd rather implement the terminal output in a follow up PR. The history of this PR is already quite long and it's getting a bit confusing. The foundation stones with the signals of `ProjectManager` are there...

>Also about how to prevent the user editing source files in the middle of a build. Not sure what closing Code in the middle of build will do either.
> 
> Presumably "Run" runs the previously build flatpak - but what if the files have been edited? Either need to rebuild or maybe disable "run" until there is an up to date build.

If you close Code while building, the process still continues 😬 I will try to fix this.

I wouldn't try to be too clever here either. All the IDEs I have used so far have not prevented the code from being edited while the project is being built.

meisenzahl:
I added [Installer's Terminal widget](https://github.com/elementary/installer/blob/master/src/Widgets/Terminal.vala) to show the output of `ProjectManager`.

Bildschirmfoto von 2021-10-17 09 16 49
Bildschirmfoto von 2021-10-17 09 17 16

I put the terminal widget in a `Gtk.Revealer` and put it in a `Gtk.Grid` together with a `Gtk.Stack` which holds the views. However, an area is always kept free for the 'Gtk.Revealer' in the 'Gtk.Grid' in gray. I would be glad about help.

Bildschirmfoto von 2021-10-17 09 17 39
Bildschirmfoto von 2021-10-17 09 17 24

I would appreciate feedback from @elementary/UX on the implementation.

meisenzahl:
I now use a `Gtk.Paned` instead of a `Gtk.Grid`. This allows to adjust the height of the terminal output.

With the `visible` property of the `Gtk.Revealer` I could achieve that no area is grayed out as placeholder for the widget.

Bildschirmfoto von 2021-10-17 11 37 18

meisenzahl:
> Not sure where to go from here. Is this an upstream problem perhaps?

I'm also stuck on this. @Marukesu do you have any idea how we can prevent the `rofiles` errors?

Marukesu:
> I am still having trouble getting the "Run" tool to work on e.g. the Calculator Flatpak (which seems to be one of the simplest with no unusual permissions). I am still getting an error in the form `.../calculator/.flatpak-builder/rofiles/rofiles-aTZhPj not initialised, use flatpak build-init` Trying different options in `build_project ()` including `--disable-rofiles-fuse` does not fix it. Nor does adding `flatpak-build-init` in `run_project ()` (that results in an `already initialized` error).

are you running on host? that's a problem with the `--metadata` arg in the manifest, older versions of flatpak-builder do not removed them when calling `flatpak build` and `--metadata` has a different meaning there, but that was already fixed upstream. maybe the version on ubuntu repositories don't have the fix.

in the flatpak version, the issue is different, for some reason, it can't access the x11 display (the same happens with the org.flatpak.Builder flatpak).

I saw that GNOME Builder uses `flatpak build` directly instead of `flatpak-builder` for run, we could fixes both issues this way.

src/MainWindow.vala
```diff
@@ -353,6 +398,11 @@ namespace Scratch {
                 on_plugin_toggled (bottombar);
             });
 
+            // Terminal
```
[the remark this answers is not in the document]
igordsm: How about something along the lines of `CommandOutput`?

igordsm:
@meisenzahl Some of these changes are not directly related to build/run Flatpak projects. How about we spin them off in a set of smaller PR so they can be approved faster and use this one only for flatpak related functions?

PR #945, opened 2021-02-08
Allow hidden project folders

### LINKED ISSUE #936: Can't add .hidden folder to sidebar as project folder
## What Happened

I have a .dotfiles repo, and having a .github repo is common to set up org-wide settings. When I try to add one of these folders to the sidebar, they don't appear.

## Expected Behavior

.hidden folders should show as top-level project folders, like any other project folder.

## Steps to Reproduce

1. Create a hidden folder beginning with a `.` character, i.e. `.github`
2. Open it with Code, either via the CLI or GUI
3. See that it doesn't show in the project sidebar

## Platform Information

elementary OS 6 Early Access

jeremypw:
Fixes #936 

This allows using a hidden folder as the project root.

Hidden subfolders are still not shown by default (unless they are hidden only because a parent is hidden, in which case they are shown)


src/FolderManager/File.vala
```diff
@@ -77,19 +77,20 @@ namespace Scratch.FolderManager {
...
-                    return false;
-                }
-
-                if (info.get_file_type () == FileType.DIRECTORY) {
-                    return true;
-                }
+        // Checks if we're dealing with a non-backup directory
+        // Hidden subfolders are not shown by default, but we need to allow hidden top-level folder
+        public bool is_valid_directory (bool allow_hidden = false) {
+            if ((!allow_hidden && name.has_prefix (".")) || //If parent is hidden then inherit validity from parent
```
cassidyjames, line 83:
```suggestion
            if ((!allow_hidden && name.has_prefix (".")) || // If parent is hidden then inherit validity from parent
```

src/FolderManager/FileView.vala
```diff
@@ -138,7 +134,7 @@ namespace Scratch.FolderManager {
             if (is_open (folder)) {
                 warning ("Folder '%s' is already open.", folder.path);
                 return;
-            } else if (!folder.is_valid_directory) {
+            } else if (!folder.is_valid_directory (true)) { //Allow hidden top-level folders
```
cassidyjames, line 137:
```suggestion
            } else if (!folder.is_valid_directory (true)) { // Allow hidden top-level folders
```

cassidyjames (dismissed):
Works and looks sensible.

Screenshot from 2021-02-10 10-45-26

cassidyjames (approved):

PR #951, opened 2021-02-10
Allow launching with pkexec

### LINKED ISSUE #202: Can't create a new file after first run as root
When I try to create a new file from the main window ("new file / open file / new from clipboard"), the app doesn't initialize the text editor window and starts looking like this (still responsive, but I can't write):

from_main_window

When I try to create a new file from an opened file (by clicking on the `+` sign in the tab list), similar story (but I can still edit the previously open file) - the plus sign stays "focused", but nothing else happens:

plus

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48727068-can-t-create-a-new-file-after-first-run-as-root?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #932: Opening Code first time via root results in access denied for new files
I just installed the latest elementary os: elementary OS 5.1.7 Hera

I had accidentally used the wrong computer name. So the first thing I did after the first boot was open a terminal and

`sudo io.elementary.code /etc/hostname`
and
`sudo io.elementary.code /etc/hosts`

After this, I get 

> [FATAL 22:28:43.138423] DocumentView.vala:151: Error opening file “/home/ayoungethan/.local/share/io.elementary.code/unsaved/Text file from 2021-01-26 22:28:43:138313”: Permission denied

When I try to open or edit a new file by default without superuser privileges.

Changing permissions in ~/.local/share/io.elementary.code appears to fix the issue. eg

`sudo chown -R ayoungethan ~/.local/share/io.elementary.code`

I believe if a user opens io.elementary.code using superuser privileges, it should not create a root-level folder in the user's home folder.

danrabbit:
And disallow launching with sudo so you don't get messed up permissions in your home

Fixes #202 
Fixes #932


tintou:
Ideally, we shouldn't even allow this and provide a `Write anyway` that would grant us write access to the file.

danrabbit:
@tintou for sure, but there's been an open issue report for this in both Code and Files for like years now so this is at least a little better than users trashing their home directory permissions

src/Application.vala
```diff
@@ -79,6 +79,12 @@ namespace Scratch {
         }
 
         public override int command_line (GLib.ApplicationCommandLine command_line) {
+            /* Only allow running with root privileges using pkexec, not using sudo */
+            if (Posix.getuid () == 0 && GLib.Environment.get_variable ("PKEXEC_UID") == null) {
+                warning ("Running Code using sudo is not possible. Use: pkexec io.elementary.code");
+                quit ();
+            };
```
jeremypw, line 86: Should `return 1;` after quit (), to avoid unnecessarily completing this function (which results in terminal errors at the moment)

jeremypw (changes requested):
This successfully preccents use of `sudo` and allows use of `pkexec` from the commandline.

Code does not currently give an indication when it runs as root and there is no "New Window as Administrator" option as yet but these could be further PRs.

One minor code issue.

jeremypw (approved):

decathorpe:
Uh, is it expected that this commit does not build yet?
(Due to missing translations for the new policy file, it looks like.)

decathorpe:
Since this PR was merged, my code builds fail with this error:

```
[6/134] /usr/bin/meson --internal msgfmthelper data/io.elementary.code.policy.in data/io.elementary.code.policy xml /builddir/build/BUILD/elementary-code-3.4.1+git210228.000514.6ffb5da7/po/extra
FAILED: data/io.elementary.code.policy 
/usr/bin/meson --internal msgfmthelper data/io.elementary.code.policy.in data/io.elementary.code.policy xml /builddir/build/BUILD/elementary-code-3.4.1+git210228.000514.6ffb5da7/po/extra
msgfmt: cannot locate ITS rules for data/io.elementary.code.policy.in
```

PR #915, opened 2020-12-13
Git manager

jeremypw:
First of a series of PRs adding more Git functionality.  See also #953 #956

Move git related stuff into Service classes.
Intended to match master's behaviour but includes small fixes:

 - New (untracked) child of a new (untracked) folder also gets an "new" icon (Master: no icon)
 - When branch changed a file that was modified in the previous branch but new in the current branch gets "new" icon (Master: gets "modified" icon)
 - Update status only on CHANGES_DONE_HINT to reduce number of signals sent
 - Do not show git status of expanded folders as this duplicates the same information shown on its children

Most of the extra code lines are needed for the GitManager class intended to allow sharing of monitored git repositories across the app.

 - [x] Show git status icons in project folder
 - [x] Update git status icons when changed
 - [x] Deprioritize ignored files
 - [x] Update when branch changes

PR #953, opened 2021-02-13
ProjectFolderItem: Add option to create new branch to context menu

jeremypw:
Based on #915


src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+                _("The branch parent will be '%s'").printf (monitored_repo.current_branch_name),
+                "applications-development",
+                Gtk.ButtonsType.CANCEL
+            );
+
+            var create_button = new Gtk.Button.with_label (_("Create Branch"));
+            create_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
+
+            var entry = new Granite.ValidatedEntry.from_regex (new Regex ("^[a-z]+[a-z0-9--]*$"));
+            entry.bind_property ("is-valid", create_button, "sensitive", BindingFlags.DEFAULT);
```
danrabbit, line 231: It looks like this needs to be `BindingFlags.SYNC_CREATE` or we get a sensitive button for an empty entry. Great use of ValidatedEntry here though! Very cool

```suggestion
            entry.bind_property ("is-valid", create_button, "sensitive", BindingFlags.SYNC_CREATE);
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
-                        try {
-                            monitored_repo.change_branch (branch_name);
-                        } catch (GLib.Error e) {
-                            warning ("Failed to change branch to %s.  %s", name, e.message);
-                        }
-                    });
+                try {
+                    if (!monitored_repo.is_detached_head) {
+                        change_branch_menu.add (new Gtk.SeparatorMenuItem ());
+                        var branch_item = new Gtk.MenuItem.with_label (_("New Branch")) {
```
danrabbit, line 341: Since this opens a dialog that takes more input, it should have ellipsis 
```suggestion
                        var branch_item = new Gtk.MenuItem.with_label (_("New Branch…")) {
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+
+        private string? get_new_branch_name () throws Error {
+            string? name = null;
+
+            var dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                _("Create a new local branch in '%s'").printf (file.name),
+                _("The branch parent will be '%s'").printf (monitored_repo.current_branch_name),
+                "applications-development",
+                Gtk.ButtonsType.CANCEL
+            );
```
danrabbit, line 225: We should make sure to set the parent for this so it gets centered properly
```suggestion
            ) {
                transient_for = (Gtk.Window) get_toplevel ();
            };
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+                warning ("Error creating branch: %s", e.message);
+            }
+        }
+
+        private string? get_new_branch_name () throws Error {
+            string? name = null;
+
+            var dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                _("Create a new local branch in '%s'").printf (file.name),
+                _("The branch parent will be '%s'").printf (monitored_repo.current_branch_name),
```
danrabbit, line 222: These should use typographical quotes:
```suggestion
                _("Create a new local branch in “%s”").printf (file.name),
                _("The branch parent will be “%s”").printf (monitored_repo.current_branch_name),
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+                    _("Please choose what to do '%s'").printf (monitored_repo.current_branch_name),
+                    "dialog-question",
+                    Gtk.ButtonsType.CANCEL
+                );
+
+                var stash_button = new Gtk.Button.with_label (_("Stash"));
+                var commit_button = new Gtk.Button.with_label (_("Commit"));
+
+                dialog.add_action_widget (stash_button, 0);
+                dialog.add_action_widget (commit_button, 1);
```
danrabbit, line 269: Is there a reason you did it this way instead of:
```suggestion
                dialog.add_button (_("Stash"), 0);
                dialog.add_button (_("Commit"), 1);
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+        }
+
+        private void deal_with_uncommitted_changes () throws Ggit.Error, Error {
+            if (monitored_repo.has_uncommitted_changes ()) {
+                var error_message = "";
+                var dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                    _("Branch '%s' has uncommitted changes").printf (monitored_repo.current_branch_name),
+                    _("Please choose what to do '%s'").printf (monitored_repo.current_branch_name),
+                    "dialog-question",
+                    Gtk.ButtonsType.CANCEL
```
danrabbit, line 262: You may have to use `none` here to get the correct button order. It should be:

[Secondary Action] [Cancel] [Primary Action]

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+                switch (response) {
+                    case Gtk.ResponseType.CANCEL:
+                        error_message = _("Cancelled by user");
+                        break;
+
+                    case 0: //Stash
+                        monitored_repo.stash_all ();
+                        return;
+
+                    case 1: //Commit
```
danrabbit, line 284: I guess we could use response type `APPLY` here?

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+
+        private void action_commit () {
+            try {
+                monitored_repo.commit_all_to_head ();
+            } catch (Error e) {
+                //TODO provide user feedback on failure
+            }
+        }
+
+        private void deal_with_uncommitted_changes () throws Ggit.Error, Error {
```
danrabbit, line 255: It seems like this dialog is thrown even when I would be able to switch branches without losing changes via Terminal

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+            if (monitored_repo.has_uncommitted_changes ()) {
+                var error_message = "";
+                var dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                    _("Branch '%s' has uncommitted changes").printf (monitored_repo.current_branch_name),
+                    _("Please choose what to do '%s'").printf (monitored_repo.current_branch_name),
```
danrabbit, line 260: This could probably be more explanatory. Something like:
```suggestion
                    _("Changes can be stashed for later to be applied to any branch, or they can be committed to “%s” before switching branches").printf (monitored_repo.current_branch_name),
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+            } catch (Error e) {
+                //TODO provide user feedback on failure
+            }
+        }
+
+        private void deal_with_uncommitted_changes () throws Ggit.Error, Error {
+            if (monitored_repo.has_uncommitted_changes ()) {
+                var error_message = "";
+                var dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                    _("Branch '%s' has uncommitted changes").printf (monitored_repo.current_branch_name),
```
danrabbit, line 259: Typographical quotes
```suggestion
                    _("Branch “%s” has uncommitted changes").printf (monitored_repo.current_branch_name),
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -108,7 +127,15 @@ namespace Scratch.FolderManager {
...
-                menu.append (new ChangeBranchMenu (monitored_repo));
+                menu.append (new BranchMenu (monitored_repo, actions));
+                if (monitored_repo.has_uncommitted_changes ()) {
+                    menu.insert_action_group (ACTION_DOMAIN, actions);
+                    var commit_item = new Gtk.MenuItem.with_label (_("Commit")) {
+                       action_name = ACTION_PREFIX + ACTION_COMMIT
+                    };
+
+                    menu.append (commit_item);
+                }
```
danrabbit, line 138: Instead of hiding and showing the item, the commit action should probably be enabled/disabled when this changes in `monitored_repo`. This will automatically set the item sensitivity and effect any keyboard shortcuts

src/Services/MonitoredRepository.vala
```diff
@@ -127,7 +104,78 @@ namespace Scratch.Services {
...
+            Ggit.Object git_object = git_repo.get_head ().lookup ();
+            var new_branch = git_repo.create_branch (name, git_object, Ggit.CreateFlags.NONE);
+            git_repo.set_head (((Ggit.Ref)new_branch).get_name ());
+        }
+
+        public void commit_all_to_head () throws Ggit.Error, Error {
+            //TODO Provide means to set various commit information
+            //TODO Warn of untracked files - at present these will be committed and tracked without warning
+            //Use automatic commit message for now
+            var message = ("Commit made at %s").printf (new DateTime.now_local ().to_string ());
```
danrabbit, line 126: Maybe we should hold off on this until we can provide an interface for creating a commit message? Or at least provide a better default commit message? For example, GitHub uses "Create %s" or "Update %s" (where `%s` is the affected filename) as its fallback.

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+                        monitored_repo.stash_all ();
+                        return;
+
+                    case 1: //Commit
+                        monitored_repo.commit_all_to_head ();
+                        return;
+
+                    default:
+                         error_message = _("Unrecognised response from user");
+                        break;
```
danrabbit, line 290: I'm not sure this really makes a ton of sense. The only other response we can get here is `DELETE` which should be handled as `CANCEL`

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+
+        private void action_switch (GLib.SimpleAction action, GLib.Variant? param) {
+            var switch_to_branch_name = param.get_string ();
+            try {
+                deal_with_uncommitted_changes ();
+                monitored_repo.change_branch (switch_to_branch_name);
+            } catch (Error e) {
+                //TODO Provide UI feedback if fails
+                warning ("Error changing branch: %s", e.message);
+            }
```
danrabbit, line 305: This doesn't seem to be handling actual errors. As far as I can tell the only thing this ever catches is when a user cancels the commit/stash dialog

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,29 +202,150 @@ namespace Scratch.FolderManager {
...
+                int response = dialog.run ();
+                dialog.destroy ();
+
+                switch (response) {
+                    case Gtk.ResponseType.CANCEL:
+                        error_message = _("Cancelled by user");
+                        break;
+
+                    case 0: //Stash
+                        monitored_repo.stash_all ();
```
danrabbit, line 281: This doesn't really seem great. The user is offered the option to stash, but there's no way to know where those changes were stashed or how to retrieve them later. Maybe we should save this functionality until we have a more fleshed out workflow for stashing

danrabbit (changes requested):
Left a number of comments. It's definitely cool to see some more advanced git features in here, but I think the experience in this branch feels pretty incomplete and not very predictable. It's pretty jarring to have things stashed or committed with no way to see what the commit message was or where it was stashed to. It kind of feels like this branch wants to do about 4 different things, but can't really do them well without becoming really big.

I think maybe it would be better to break it out into individual branches so we can work through the flow of each feature a little more

jeremypw:
Closed in favour of #997

PR #956, opened 2021-02-22
Fix switch branch

### LINKED ISSUE #954: Changing git branch does not update working directory
## What Happened
Changing the current branch using the ProjectFolderItem context menu does not change the contents of the working directory even if both branches have no uncommitted changes.

## Expected Behavior
When changing from to a fully committed branch to another branch the contents of the working directory (and open documents) should be changed to reflect the new branch.

## Steps to Reproduce
1. Open a repository with at least two fully committed branches (with differences) and no uncommitted changes
2. Using the context meny switch from one branch to another
3. The documents do not changes and the git status indicators change in accordance with the differences in the two branches.

## Platform Information
Elementary 6.0 in VirtualBox VM

### Prior Art
In Geany, switching between fully committed branches from the commandline with `git checkout <branch-name>` marks relevant open documents as 'changed' and allows reloading from disk to reflect the new branch working directory.  The git status of the branch changed to shows no uncommitted changes.

jeremypw:
Fixes #954

Also implements an option to stash uncommitted changes before switching branch (or creating a new one)


danrabbit:
Is the target branch correct here? Should this be proposed to merge into master now?

jeremypw:
After this is merged into the `new-branch` branch then the target will become `master`.  Doing it this way was to keep the diff size down. Creating a new branch involves switching branch so I probably could have done it the one branch.

PR #960, opened 2021-03-04
Flatpak: Add granite master, polkit, bump Hdy

danrabbit:


Marukesu:
AFAIK, flatpak doesn't allow pkexec or sudo, so I think it makes sense to use pkexec only on native compiling.

jeremypw (approved):

PR #971, opened 2021-03-07
Project search

jeremypw:
Simple implementation of a global search within the tracked files of a project

 - [x] Add "Project Search" option to project folder context menu
 - [x] Implement simple dialog to get search term and whether or not to treat as Regex expression
 - [x] Search tracked files (currently limited to `.vala` files)
 - [x] Count matches per file and display as badge
 - [x] Expand project folder to make matched files visible  
 - [x] Allow input of filter for files to search
 - [x] Search within project sub-folder
 - [x] Case insensitive option


jeremypw:
I think this OK for a review now. I have used it a while for my own work and it is quite useful.  There are a number of refinements possible but I do not want to over-complicate this PR.

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -136,6 +136,13 @@ namespace Scratch.FolderManager {
             menu.append (close_item);
             menu.append (close_all_except_item);
             menu.append (delete_item);
+
+            var search_item = new Gtk.MenuItem.with_label (_("Search Project"));
+            search_item.activate.connect (() => { global_search (); });
```
danrabbit, line 141: This should probably be a GLib.Action so that we can assign a keyboard shortcut to it. It looks like VS Code uses `Ctrl + Shift + F` for global search

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -289,6 +298,215 @@ namespace Scratch.FolderManager {
             return false;
         }
 
+        public void global_search (string search_term = "",
+                                   GLib.File? start_folder = file.file,
+                                   bool is_literal = true,
+                                   bool tracked_only = true,
+                                   bool recurse = true) {
```
danrabbit, line 305: It looks like you're only ever using one of these arguments, so we probably don't need all of this complexity

src/Dialogs/GlobalSearchDialog.vala
```diff
@@ -0,0 +1,199 @@
...
+            } else {
+                return filter_combo.active_id.split (";");
+            }
+        }
+    }
+
+    public GlobalSearchDialog (Gtk.Window? parent, string folder_name, bool is_repo) {
+        Object (
+            deletable: false,
+            resizable: false,
```
danrabbit, line 77: `deletable` is already set in granite dialog.

Is there a reason not to allow this to be resizable?

src/Dialogs/GlobalSearchDialog.vala
```diff
@@ -0,0 +1,199 @@
...
+            transient_for: parent,
+            folder_name: folder_name,
+            is_repo: is_repo
+        );
+    }
+
+    construct {
+        border_width = 0;
```
danrabbit, line 85: shouldn't need to mess with this in Granite.Dialog

src/Dialogs/GlobalSearchDialog.vala
```diff
@@ -0,0 +1,199 @@
...
+        match_mode_button = new Granite.Widgets.ModeButton () {
+            margin = 6
+        };
+
+        literal_mode_index = match_mode_button.append_text (_("Text"));
+        match_mode_button.append_text (_("Regex"));
+        match_mode_button.selected = literal_mode_index;
+
+        filter_combo = new Gtk.ComboBoxText.with_entry () {
+            margin = 6
```
danrabbit, line 110: Instead of setting margins on each widget, you should use column and row spacing on the grid

src/Dialogs/GlobalSearchDialog.vala
```diff
@@ -0,0 +1,199 @@
...
+            halign = Gtk.Align.END,
+            margin_start = 6
+        };
+
+        scope_mode_button = new Granite.Widgets.ModeButton () {
+            margin = 6,
+            no_show_all = !is_repo
+        };
+
+        tracked_mode_index = scope_mode_button.append_text (_("Tracked"));
```
danrabbit, line 130: What does "tracked" mean?

src/Dialogs/GlobalSearchDialog.vala
```diff
@@ -0,0 +1,199 @@
...
+        scope_mode_button.append_text (_("All"));
+        scope_mode_button.selected = tracked_mode_index;
+
+        modified_switch = new Gtk.Switch () {
+            margin = 6,
+            halign = Gtk.Align.START, //Stop switch expanding
+            active = true,
```
danrabbit, line 137: I'm not sure this is a good default. We should probably make the defaults more inclusive and let users make them more restrictive as necessary

src/Dialogs/GlobalSearchDialog.vala
```diff
@@ -0,0 +1,199 @@
...
+        layout.attach (scope_mode_button, 3, 2, 1, 1);
+        layout.attach (modified_label, 0, 3, 1, 1);
+        layout.attach (modified_switch, 1, 3, 1, 1);
+        layout.attach (recurse_label, 0, 4, 1, 1);
+        layout.attach (recurse_switch, 1, 4, 1, 1);
+        layout.attach (case_label, 0, 5, 1, 1);
+        layout.attach (case_switch, 1, 5, 1, 1);
+        get_content_area ().add (layout);
+
+        var close_button = (Gtk.Button) add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
```
danrabbit, line 182: It doesn't look like you ever reference the cancel button again, so it's probably not necessary to cache it

danrabbit (changes requested):
Definitely a handy feature! Left a few inline comments

src/FolderManager/ProjectFolderItem.vala
jeremypw: The problem is that Code supports having more than one project open (which can be useful) so we need to determine which project is intended to be searched.  Presumably we could use whichever project the active document belongs to?

src/FolderManager/ProjectFolderItem.vala
jeremypw: This was intended for future development - I envisaged perhaps having some way to launch the global search from the searchbar.  But I can remove them for now.

src/Dialogs/GlobalSearchDialog.vala
jeremypw: The layout of the dialog was temporary - I expect the design team will want to refine it. I didn't think dialogs are generally resizable but I can remove that.

src/Dialogs/GlobalSearchDialog.vala
jeremypw: OK

src/Dialogs/GlobalSearchDialog.vala
jeremypw: It means git is tracking changes in the file.  Alternative?

src/Dialogs/GlobalSearchDialog.vala
jeremypw: OK.  I intend to make user choices persist between uses later.

src/Dialogs/GlobalSearchDialog.vala
jeremypw: OK

src/Dialogs/GlobalSearchDialog.vala
jeremypw: Should probably not show this switch when not a git repo.

src/Dialogs/GlobalSearchDialog.vala
jeremypw: OK

src/FolderManager/ProjectFolderItem.vala
jeremypw: OK implemented an action

src/FolderManager/ProjectFolderItem.vala
jeremypw: Removed unnecessary parameters

src/Dialogs/GlobalSearchDialog.vala
jeremypw: Simplified layout and made resizable

src/Dialogs/GlobalSearchDialog.vala
jeremypw: Changed default

src/Dialogs/GlobalSearchDialog.vala
jeremypw: I have added a tooltip to explain what "Tracked" means.  Also added tooltips to other modes for consistency.

danrabbit: Hm yeah, my first guess was maybe it has to do with indexing. Is this something that you have encountered as a real case or something that was just added because it's possible?

cassidyjames (changes requested):
Brief UX review:

1. ~After performing a single search, I can no longer re-open the Project Search dialog—even after closing and re-opening Code. :o~ Apparently this only works when a text file is already open; it should work when there are no open files.

2. I like using badges in the sidebar to show the number of results, but I wonder if it's not obvious enough… at first, I performed a search and then sat there waiting for something more obvious to happen. I'm not sure what to do about that.

3. There's no obvious way to clear a search (including the badges in the sidebar) after performing it.

4. Should this be more integrated into the existing Find functionality? e.g. an option in the Find bar to "Find in current project" to trigger it?

5. We should badge parent folders with the total number of hits in their child directories, maybe even instead of expanding all directories that have hits.

6. We call the action "Find…" in the headerbar, but then call it a search bar… should we call this "Find in Project" instead of "Search Project"?

7. The menuitem should have an ellipsis since it's opening a dialog and the action is performed there.

Of course, some of these could be follow-up issues if it's too much scope creep.

src/Dialogs/GlobalSearchDialog.vala
jeremypw: I have occasionally wanted to search the C files under the `build` folder for example, which are git-ignored so not tracked.  Although I suppose it would now be possible to search that folder specifically by using its context menu (that wasn't initially possible).  There may be other use cases I haven't thought of.  If you think that when searching a project folder users will only expect tracked files to be searched then I could omit this switch but it is possible users might be puzzled as to why some files in the folder were not searched.

jeremypw:
@cassidyjames Thanks for the review!

1. This is a consequence of using a simple MainWindow action for the menuitem.  If there are several project folders then MainWindow does not know which one to search.  I'll try adding a parameter to the action to get round this.
2. Maybe a toast showing the total number of matches and matching documents?  I was thinking about creating a search results document listing the files and lines containing matches. Geany does something similar (using a separate widget). That would be for another PR.
3. What else needs clearing other than the badges? The searchbar can be cleared in the usual way.  Is there a need to clear the badges - they will be cleared by the next search.  Maybe could clear them when the folder is collapsed?
4. This was planned for a future PR.
5. I disagree with not expanding the folders - I want to see the exact files where there are matches.  The total number of matches within a folder doesn't seem very useful.
6. OK, I can change that.
7. OK, will do

jeremypw:
@cassidyjames  I have addressed your points 1, 6 and 7.  The other points, especially 4 & 5, are best left to other PRs after some design thought.

danrabbit:
Yeah, I think eventually the design of the sidebar should change so that it's clear what project is the current focus of work. This will probably be important for git features as well.

I'm not sure if it's a pre-req for this branch, but it would probably make things easier

jeremypw:
You could have the same number of documents open from two projects so it is not obvious what the focus of work is other than which document is currently active (or maybe the last one that was changed?)

jeremypw:
@danrabbit @cassidyjames Are there any issues that cannot be left to another PR remaining?  I use this all the time for my own development and the main thing I would like would be persistence of the previous search, at least during the session, but I prefer to do that in another PR.

danrabbit:
@jeremypw Sorry, I've created some conflicts here while merging some of your other great fixes. Once those are resolved I'd be happy to take another look at this. Thank you for working on all of these things!

jeremypw:
At the moment global search via keyboard accelerator only works if there is a document open as the current document is used to determine the current project to search.

src/Dialogs/GlobalSearchDialog.vala
danrabbit: My inclination would be to make the search as inclusive as possible and then we can come back and add more options if there's a demand for them in actual usage

jeremypw: OK, I'll try using it with no filters for a while to see whether there are any serious drawbacks. The filters are based on what is supplied in Geany and what I have used in the past.

jeremypw: Do you mean to keep the filters in the dialog but default them all to the most inclusive settings - or remove them altogether?  I notice you have pushed some fixes to this dialog, including the filters.

danrabbit: My personal inclination would be to remove any of the filters that aren't immediately and obviously important. There's a few of these that seem a little hard to describe and may need some more copy writing thought if we want to include them, and then there's translation work etc that could be saved if they end up not being necessary. So we could save ourselves some work here by starting simpler

jeremypw: OK, I'll remove them altogether from the dialog (except, I would propose the Text/Regex switch and the Case Sensitive/Insensitive switch - which I use a lot).  I'll keep some of the infra-structure to make it easier to reinstate them as required.  One thought I had was to put the extra options in an "Advanced" revealer. I think most people familiar with git will understand these options easily enough but people using Code for other purposes do not need to see them.

jeremypw:
Revised Global Search Dialog:
Screenshot from 2021-04-22 17 38 53

jeremypw:
For comparison, this is the equivalent dialog in Geany:

Screenshot from 2021-04-22 18 01 08

danrabbit:
This seems really good to me for a first version here. I think the only thing that feels weird is like Cassidy said there's no obvious way to clear the badges. What would you think about clearing the badges when the find bar is cleared? I feel like that should feel consistent since clicking a badged row would no longer show anything highlighted

It would probably be good to be consistent with the find bar and have case sensitivity disabled by default

Does Regex need to be able to be turned off? In my testing here, I have it enabled and it doesn't seem to be hindering my searches that don't contain any special regex syntax

jeremypw:
I don't think it can be excluded that a user might want to search for a literal string that does contain e.g. "^" or "$" or some other valid Regex.  In Geany at least you had to turn on Regex search if you wanted it.  The only drawback that I haven't yet address is that Regex search is not available in the headerbar so when the search term gets copied over it is not interpreted in the same way.  Similarly, at the moment, the "case sensitive" setting is not copied across to the headerbar.

I think clearing the badges when the search entry is cleared or closed is a good idea.  Will implement.

danrabbit (approved):
Nice work!

PR #975, opened 2021-03-15
Build FlatPak without policykit

jeremypw:


src/Application.vala
```diff
@@ -81,7 +81,11 @@ namespace Scratch {
         public override int command_line (GLib.ApplicationCommandLine command_line) {
             /* Only allow running with root privileges using pkexec, not using sudo */
             if (Posix.getuid () == 0 && GLib.Environment.get_variable ("PKEXEC_UID") == null) {
+#if HAVE_PKEXEC
                 warning ("Running Code using sudo is not possible. Use: pkexec io.elementary.code");
+#else
+                warning ("Running Code using sudo is not possible."); //Pkexec cannot be used in e.g. Flatpak
```
Marukesu, line 87: sudo cannot be used with flatpak either, so this warning seems unecessary.

PR #985, opened 2021-04-09
Better duplication of multiple lines

jeremypw:
Improves the way multiple-lines are duplicated - now behaves like Geany.

If the selection includes multiple lines then *the whole* of the lines including the selection are duplicated and placed on a new line after the last line selected. This means it is no longer required to carefully select the whole of the lines to be duplicated and the duplicate is not appended to the end of the last line selected - which is not expected.

Also the original selection is now selected while the duplicated lines are not selected. This makes it easier to e.g. comment out the original lines before modifying the duplicate lines - this is useful for making experimental changes while keeping the original visible.

An unused (and badly named) property of SourceView is removed in passing

### DIFF
--- a/src/Widgets/SourceView.vala
+++ b/src/Widgets/SourceView.vala
@@ -307,27 +307,42 @@ namespace Scratch.Widgets {
             return 0;
         }
 
-        // Duplicate selected text if exists, else duplicate current line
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
This is massively improved! Great work on this!

PR #988, opened 2021-04-14
Flatpak: Use elementary runtime and Sdk

danrabbit:


io.elementary.code.yml
```diff
@@ -1,9 +1,7 @@
 app-id: io.elementary.code
-runtime: org.gnome.Sdk # outline plugin needs libvala that is only on sdk
-runtime-version: '3.38'
-base: io.elementary.BaseApp
-base-version: juno-20.08
-sdk: org.gnome.Sdk
+runtime: io.elementary.Platform
```
Marukesu, line 2: The need for use the sdk here for outline still
```suggestion
runtime: io.elementary.Sdk
```

Marukesu (changes requested):

Marukesu (commented):
Now it works with all plugins enabled. but i got some new criticals on terminal because of zeitgeist. we need to add `--talk-name=org.gnome.zeitgeist.Engine` to the manifest

Marukesu (approved):

PR #992, opened 2021-04-21
Fix regression due to c35a306f03e3ddef5c2d3c8dff4e69caaa78c667

jeremypw:
Fixes a regression causing the sidebar actionbar not to be displayed.


jeremypw:
Not sure how we missed this regression!

danrabbit (approved):
Nice catch. Oops :grimacing:

PR #997, opened 2021-04-27
Add option to create new git branch

jeremypw:
The option is added to the ProjectFolderItem/Branch submenu and is also accessible with the keyboard shortcut `<Control>B`


src/FolderManager/FileView.vala
```diff
@@ -144,6 +144,16 @@ namespace Scratch.FolderManager {
             find_path (root, path, true);
         }
 
+        public void new_branch (GLib.File current_doc_file) {
+            foreach (var child in root.children) {
+                var item = (ProjectFolderItem)child;
```
tintou, line 149:
```suggestion
                unowned var item = (ProjectFolderItem)child;
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -201,6 +260,15 @@ namespace Scratch.FolderManager {
                     });
                 }
 
+                if (monitored_repo.head_is_branch) {
+                    change_branch_menu.add (new Gtk.SeparatorMenuItem ());
+                    var branch_item = new Gtk.MenuItem.with_label (_("New Branch…")) {
+                        action_name = "win.action_new_branch"
+                    };
+
+                    change_branch_menu.add (branch_item);
```
danrabbit, line 269: Instead of adding or removing the menuitem, we should enable/disable the action. Removing the menuitem when it might be expected can look like a bug, whereas an insensitive menuitem communicates that this option is not available

src/MainWindow.vala
```diff
@@ -170,6 +172,7 @@ namespace Scratch {
             action_accelerators.set (ACTION_NEXT_TAB, "<Control>Tab");
             action_accelerators.set (ACTION_PREVIOUS_TAB, "<Control><Shift>Tab");
             action_accelerators.set (ACTION_CLEAR_LINES, "<Control>K"); //Geany
+            action_accelerators.set (ACTION_NEW_BRANCH, "<Control>B"); //Git action - new branch
```
danrabbit, line 175: This comment seems redundant. If it isn't clear enough from the action name, perhaps the action name should include `GIT`

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,8 +176,67 @@ namespace Scratch.FolderManager {
...
+
+            var create_button = new Gtk.Button.with_label (_("Create Branch"));
+            create_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
+
+            var entry = new Granite.ValidatedEntry.from_regex (new Regex ("^[a-z]+[a-z0-9--]*$"));
+            entry.bind_property (
+                "is-valid", create_button, "sensitive", BindingFlags.DEFAULT | BindingFlags.SYNC_CREATE
+            );
+
+            dialog.add_action_widget (create_button, Gtk.ResponseType.APPLY);
```
danrabbit, line 214: We can cache the return value of `add_button ()` so we don't need to use `add_action_widget ()`

```suggestion
            var create_button = (Gtk.Button) dialog.add_button (_("Create Branch"), Gtk.ResponseType.APPLY);
            create_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);

            var entry = new Granite.ValidatedEntry.from_regex (new Regex ("^[a-z]+[a-z0-9--]*$"));
            entry.bind_property (
                "is-valid", create_button, "sensitive", BindingFlags.DEFAULT | BindingFlags.SYNC_CREATE
            );
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -176,8 +176,67 @@ namespace Scratch.FolderManager {
...
+        public void new_branch () {
+            try {
+                if (monitored_repo.head_is_branch) {
+                    var new_branch_name = get_new_branch_name ();
+                    if (new_branch_name != null) {
+                        monitored_repo.create_new_branch (new_branch_name);
+                    }
+                }
+            } catch (Error e) {
+                warning ("Error creating branch %s", e.message);
```
danrabbit, line 188: This should probably be a user facing error shouldn't it?

danrabbit (changes requested):
A few comments/suggestions :)

src/MainWindow.vala
jeremypw: OK, its gone.

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -405,6 +462,18 @@ namespace Scratch.FolderManager {
                     });
                 }
 
+                var main_window = (MainWindow)(project_folder.view.get_toplevel ());
```
danrabbit, line 465: Would it be more reliable to use `Gtk.Application.get_active_window ()`?

danrabbit (commented):
It seems that creating a new branch fails if I don't have any documents open:
```
scratch_folder_manager_file_view_new_branch: assertion 'current_doc_file != NULL' failed
```

src/FolderManager/ProjectFolderItem.vala
jeremypw: OK, done

src/FolderManager/ProjectFolderItem.vala
jeremypw: Oh yes, that's easier.

jeremypw:
Both this and the project search branch assume there is an active document because that is how the active project folder is known.  Otherwise, when using the shortcut there is no way of telling which project you want the action to be performed on.  

Not sure what the best solution is for this, until we have an "active project" selector in the header.

danrabbit:
@jeremypw if I don't use the shortcut and right click on a toplevel folder, surely we know the project that we're intending to work with right?

danrabbit:
In any case, if we're unable to determine the correct project, there should probably be a user-facing error dialog for this

jeremypw:
The context menu uses the the same MainWindow action as the shortcut but there is currently no way of including the project folder as a parameter of the action. That is something that could be implemented as a stop gap for a more complete solution, if it is thought that is the way to go. And, as you say, a warning dialog could just be shown if the active project is not determinable

src/Dialogs/NewBranchDialog.vala
```diff
@@ -0,0 +1,84 @@
...
+        var create_button = (Gtk.Button) add_button (_("Create Branch"), Gtk.ResponseType.APPLY);
+        create_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
+
+        new_branch_name_entry.bind_property (
+            "is-valid", create_button, "sensitive", BindingFlags.DEFAULT | BindingFlags.SYNC_CREATE
+        );
+
+        new_branch_name_entry.activate.connect (() => {
+            response (new_branch_name_entry.text != "" ? Gtk.ResponseType.ACCEPT : Gtk.ResponseType.CLOSE);
+        });
```
danrabbit, line 80: Instead of this you can set `activates_default = true` on the entry

src/Dialogs/NewBranchDialog.vala
```diff
@@ -0,0 +1,84 @@
...
+            project_list: project_list,
+            image_icon: new ThemedIcon ("git")
+        );
+    }
+
+    construct {
+        if (active_project != null) {
+            assert (active_project.is_git_repo);
+            primary_text = _("Create new branch in project “%s”").printf (active_project.file.file.get_basename ());
+            secondary_text = _("The barnch name must be lower-case, start with a letter, and be at least 3 characters");
```
danrabbit, line 45: typo
```suggestion
            secondary_text = _("The branch name must be lower-case, start with a letter, and be at least 3 characters");
```

danrabbit (changes requested):

src/Dialogs/NewBranchDialog.vala
jeremypw: The problem with that is that the default response is "CANCEL" so pressing Enter would always just cancel the dialog even if the entry contains a valid name.  The response returned needs to depend on whether the entry is valid or not.

src/Dialogs/NewBranchDialog.vala
jeremypw: Good spot

src/Dialogs/NewBranchDialog.vala
danrabbit: Right so then you might need to make sure that the create branch button has `can_default`. Activating the entry shouldn't do two different things depending on its input

jeremypw:
A question: What should be done if there are 2 project folders open and the active document does not belong to either but there is an unfocused tab with a document from one of them?  At the moment, the warning dialog message is `"Open a document in a git project folder in the sidebar or use a project context menu."` which is not quite right. Is it worth trying to deal with this corner case?

danrabbit:
@jeremypw you know, would it be crazy to have a dropdown and ask the user to pick which project they intended if we can't figure it out?

jeremypw:
> would it be crazy to have a dropdown and ask the user to pick which project they intended if we can't figure it out?

Yes, that was going to be a future PR - I have put in place the infrastructure by passing a list of projects to the dialog.  I can implement in this PR if desired.

src/Dialogs/NewBranchDialog.vala
```diff
@@ -0,0 +1,82 @@
...
+            if (project_list.length () == 0) {
+                secondary_text = _("Open a git project folder in the sidebar.");
+            } else {
+                secondary_text = _("Open a document in a git project folder in the sidebar or use a project context menu.");
+            }
+        }
+
+        try {
+            //Branch name must be lower-case, start with a letter and be at least 3 characters long
+            // new_branch_name_entry = new Granite.ValidatedEntry.from_regex (new Regex ("^[a-z].(?=[a-z0-9--]{2,}$)"));
```
danrabbit, line 59: Still need this?

danrabbit (approved):
In that case, let's merge what we've got and we can do that in a follow up branch :)

PR #998, opened 2021-04-28
Update POTFILES

ryonakano:


danrabbit (approved):
Thanks for catching this!

PR #1000, opened 2021-04-28
NewBranchDialog: entry activates default

danrabbit:


jeremypw (approved):
Ah, OK. I did not realize that the ValidatedEntry will not activate when it is not valid.

PR #1001, opened 2021-04-28
NewBranchDialog: Update copy, buttons

danrabbit:
I'm not sure about the copy here on the error side of things. Any suggestions that are less awkward?


jeremypw:
Yes, I found it difficult to come up with something concise, accurate and translatable :grimace:.  
Not sure about "git-using" - might cause some translators difficulties and is not very elegant.
Having a picklist of available projects would make it easier as you can just say "Please choose the project in which to create the new branch" or "No suitable projects are open".

PR #1002, opened 2021-04-28
NewBranchDialog: update primary text

danrabbit:
little more natural sounding, I think


src/Dialogs/NewBranchDialog.vala
```diff
@@ -41,7 +41,7 @@ public class Scratch.Dialogs.NewBranchDialog : Granite.MessageDialog {
     construct {
         if (active_project != null) {
             assert (active_project.is_git_repo);
-            primary_text = _("Create new branch in project “%s”").printf (active_project.file.file.get_basename ());
+            primary_text = _("Create a new branch in “%s”").printf (active_project.file.file.get_basename ());
```
danrabbit, line 44: Actually, should this be "of" instead of "in"?
```suggestion
            primary_text = _("Create a new branch of “%s”").printf (active_project.file.file.get_basename ());
```

cassidyjames (approved):

src/Dialogs/NewBranchDialog.vala
cassidyjames: Hm, both make sense though the imply slightly different things. I think you're right, a branch "of" the project works better.

cassidyjames (approved):

PR #1006, opened 2021-04-28
FolderManager: Add accel to find in project and rename find in folder

danrabbit:
Also be consistent with our separators while here


cassidyjames (approved):
dope

PR #1009, opened 2021-04-28
GlobalSearchDialog: use entry activates default

danrabbit:
get rid of an unnecessary valign while we're here


jeremypw:
I have incorporated similar code into #1011 - would you mind reviewing that first?

PR #1010, opened 2021-04-28
Fix global search when using menuitem and no open files

### LINKED ISSUE #1007: Selecting "Find in Project" with no files open fails silently
## What Happened

When selecting the menu item "Find in Project…" with no files open, Code fails silently.

## Expected Behavior

Since this is a menuitem, we should know which project the user is acting upon and correctly launch the dialog

## Steps to Reproduce

1. Close all files
2. Right click a project folder in the sidebar and select "Find in Project…"
3. Notice that nothing happens

danrabbit:
Fixes #1007


jeremypw (approved):
Works as expected, thanks!

PR #1011, opened 2021-04-29
Fix empty and cancelled global search

### LINKED ISSUE #1008: Global Search: "Search" should be insensitive with empty entry
## What Happened

When performing a global search, you can initiate a search with an empty search field, which highlights ever single file with hundreds of matches

## Expected Behavior

The "Search" button should be insensitive unless there is input

## Steps to Reproduce

1. Have a file open so that you can open the global search dialog
2. Search without entering any input
3. Notice that all files match hundreds of times

jeremypw:
Fixes #1008 

Also rebases the dialog on Granite.MessageDialog, similar to the NewBranchDialog.

### DIFF
--- a/src/Dialogs/GlobalSearchDialog.vala
+++ b/src/Dialogs/GlobalSearchDialog.vala
@@ -19,7 +19,7 @@
 * Authored by: Jeremy Wootten <jeremy@elementaryos.org>
 */
 
-public class Scratch.Dialogs.GlobalSearchDialog : Granite.Dialog {
+public class Scratch.Dialogs.GlobalSearchDialog : Granite.MessageDialog {
     public string folder_name { get; construct; }
     public bool is_repo { get; construct; }
     private Granite.ValidatedEntry search_term_entry;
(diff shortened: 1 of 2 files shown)


jeremypw:
GlobalSearchDialog

src/Dialogs/GlobalSearchDialog.vala
```diff
@@ -51,76 +51,78 @@ public class Scratch.Dialogs.GlobalSearchDialog : Granite.Dialog {
...
-                target_val.set_boolean (src_val.get_string ().length >= 3);
-            }
+        search_term_entry.bind_property (
+            "is-valid", search_button, "sensitive", BindingFlags.DEFAULT | BindingFlags.SYNC_CREATE
         );
 
-        search_term_entry.activate.connect (() => {
-            response (search_term_entry.text != "" ? Gtk.ResponseType.ACCEPT : Gtk.ResponseType.CLOSE);
+        search_term_entry.changed.connect (() => {
+            search_term_entry.is_valid = search_term_entry.text.length >= 3;
```
danrabbit, line 125: Does it really need to be >= 3 or just 0 ?

danrabbit (changes requested):
I opened #1013 with a few requested changes

src/Dialogs/GlobalSearchDialog.vala
jeremypw: The rules for legal branch names are quite complicated, but single letter names are accepted by git (not sure if GitHub imposes any additional restrictions).  Not sure where I got the existing Regex and min length from but they need relaxing to exactly match git rules.  I just noticed that there is a `is_valid_name` static function provided in the git library so I'll see if I can use that.

danrabbit: @jeremypw that sounds like a good idea for the git branch, but this is the global search branch ;) Maybe some coffee or sleep? lol

danrabbit (approved):

PR #1012, opened 2021-04-29
NewBranchDialog: Improve new branch name validation

### LINKED ISSUE #1004: Creating a new branch doesn't check if branch name exists
## What Happened

I created a new branch and named it the same as a branch that already exists. I was presented with an error dialog

## Expected Behavior

We should validate against existing branch names to make sure the user can't input a branch name that already exists

## Steps to Reproduce

1. Select "New Branch…"
2. Input a branch name you know already exists
3. Get error

jeremypw:
Fixes #1004

Also
* Use unowned variables for strings where possible
* MonitoredRepositor: Use list not array for local branch names
* ProjectFolderItem: Public getter function for branch names
* ChangeBranchMenu: Use construct clause

Name validation now enforces the git reference naming rules and checks for matching local branches.


src/Dialogs/NewBranchDialog.vala
```diff
@@ -39,10 +39,14 @@ public class Scratch.Dialogs.NewBranchDialog : Granite.MessageDialog {
...
     construct {
+        unowned List<string> branch_names = null;
         if (active_project != null) {
             assert (active_project.is_git_repo);
+            branch_names = active_project.get_branch_names ();
             primary_text = _("Create a new branch of “%s”").printf (active_project.file.file.get_basename ());
-            secondary_text = _("The branch name must be lower-case, start with a letter, and be at least 3 characters");
+            var secondary_1 = _("The branch name must be lower-case, start with a letter, and be at least 3 characters.");
+            var secondary_2 = _("The name must not already exist");
+            secondary_text = ("%s\n%s").printf (secondary_1, secondary_2);
```
danrabbit, line 49: I'm not sure that we should force a line break here, the text should wrap as necessary
```suggestion
            secondary_text = _("The branch name must be lower-case, start with a letter, and be at least 3 characters. The name must not already exist.");
```

src/Dialogs/NewBranchDialog.vala
```diff
@@ -61,6 +65,17 @@ public class Scratch.Dialogs.NewBranchDialog : Granite.MessageDialog {
...
             };
+
+            new_branch_name_entry.changed.connect (() => {
+                unowned var new_name = new_branch_name_entry.text;
+                foreach (unowned var name in branch_names) {
+                    if (new_name != name) {
+                        continue;
+                    }
+
+                    new_branch_name_entry.is_valid = false;
```
danrabbit, line 76: Shouldn't we break the loop if we find a match?

danrabbit (changes requested):

src/Dialogs/NewBranchDialog.vala
jeremypw: OK, I was concerned that in some locales the break might occur in an awkward position but I guess that could happen anyway.

jeremypw: Oh yes, done.

src/Dialogs/NewBranchDialog.vala
jeremypw: I have now replaced this code with an improved method.

src/Dialogs/NewBranchDialog.vala
jeremypw: Text now updated without new line.

jeremypw:
A possible (future) refinement would be to choose the branch on which to base the new branch.  Currently it is based on the current branch and switches to the new branch like the `git checkout -b new_branch` command.

danrabbit (approved):

PR #1013, opened 2021-04-29
Use fewer grids

danrabbit:
* Only one grid
* Remove the "search for" label, since it seems fairly obvious what to do with the entry
* Use the semantically correct icon name
* typographical quotes


jeremypw (approved):
Thanks! I thought it should be possible with one grid but for some reason I could not get it to look right.

PR #1014, opened 2021-04-29
GitManager: debug on error

danrabbit:
Since it sounds like we expected this error when there is no git repo, move it to debug to avoid log noise


jeremypw (approved):
Makes sense

Issue #991, opened 2021-04-20
A method to indicate the current project


danrabbit:
## Problem

Some more advanced IDE features require us to know which project we're currently working in to function in a predictable manner. Global search, build & run, git features, etc. At the moment, we have no clear way of doing this.

## Proposal

- [ ] Add a dropdown to the headerbar to manage and select projects
- [ ] Only list files and folders from the selected project in the sidebar
- [ ] Remove items that no longer make sense like "close other folders"

This would automatically close:
* #568

![image](https://user-images.githubusercontent.com/7277719/115475669-005bae80-a1f5-11eb-8aa4-eee251a40254.jpeg)

jeremypw:
At the moment git actions are on the context menu of the project folder so would be expected to apply to that project.  However, moving some actions to the headerbar would make them more convenient but require an "active project" concept.

One possibility would be to limit only one project per window - if you require two projects open (which can be useful for comparison, copying code etc) then it would be opened in a new window.

It feels a bit odd to not list files in the sidebar if they are opened in the notebook. Maybe they could be in a "Other Files" section?

PR #1015, opened 2021-04-30
Implement choose project dialog and use for new-branch action

### LINKED ISSUE #1005: New Branch Dialog: offer a dropdown if focused project is unknown
## Problem

When creating a new branch, if Code isn't able to determined the focused branch we get a warning dialog explaining that we must first open a project etc.

## Proposal

Instead, we could offer a dropdown listing the available projects so that the user could choose the one they meant to create a branch for

jeremypw:
Fixes #1005


jeremypw:
ChooseProjectDialog

Maybe do not need both lines of text?

danrabbit:
If we're creating a separate UI anyways, maybe we should stick this as a combobox on the headerbar to solve #991?

jeremypw:
@danrabbit OK, but there are some design questions to answer:

* Where on the headerbar should it appear
* Should it hide or become insensitive if no git projects are loaded - or if only one project is loaded.
* Should it default to the first (or last) opened project if there are  multiple projects but no documents opened?

danrabbit:
@jeremy:

* My inclination would be to put it at the very start of the headerbar to imply that it effects every action in the application, including others that we may move to the headerbar
* Yeah, probably it would make sense to hide for now if there are no projects loaded, but it would make sense to have it when only one project is loaded since it still does indicate the active project.
* My inclination would be that we try to keep it updated with the most recently active project. Basically whatever our guess would have been before, we could do that logic here pro-actively instead of waiting until an action is taken. So I guess we would want to update that when we change the focused document?

jeremypw:
Closing in favour of #1023

PR #1016, opened 2021-05-01
NewBranchDialog: Unify button positions

ryonakano:
In favor of HIG


jeremypw (commented):
Ignore my previous comment  - I must have had the wrong branch running.  In master the button order is incorrect.  Note that this dialog is substantially changed by #1015 so any changes are best proposed there - I'll make sure that PR has the correct button order.

jeremypw:
Also, adding the `Cancel` button should not be inside the `if()` clause at it should be always shown.

ryonakano:
OK, I'll close this PR in favor of your PR #1015 😁️

PR #1022, opened 2021-05-05
Spelling: Git x3, can't already exist.

comradekingu:


src/Dialogs/NewBranchDialog.vala
```diff
@@ -47,7 +47,7 @@ public class Scratch.Dialogs.NewBranchDialog : Granite.MessageDialog {
                 active_project.get_current_branch_name ()
             );
             ///TRANSLATORS "Git" is a proper name and must not be translated
-            secondary_text = _("The branch name must comply with Git rules and must not already exist.");
+            secondary_text = _("The branch name must comply with Git rules and can't already exist.");
```
jeremypw, line 50: I am not a fan of the contracted form "can't" and I do not think it is correct in this case anyway.  "Cannot" means to an inability to do something (perhaps for a purely physical or logical reason) whereas "must not" refers to a restriction due to some kind of rule.

comradekingu, line 50:
```suggestion
            secondary_text = _("The branch name must be unique and comply with Git standards.");
```

src/Dialogs/NewBranchDialog.vala
```diff
@@ -47,7 +47,7 @@ public class Scratch.Dialogs.NewBranchDialog : Granite.MessageDialog {
                 active_project.get_current_branch_name ()
             );
             ///TRANSLATORS "Git" is a proper name and must not be translated
-            secondary_text = _("The branch name must comply with Git rules and must not already exist.");
+            secondary_text = _("The branch name must be unique and comply with Git standards.");
```
danrabbit, line 50: I'm not sure "standards" is the right word here. It is a technical requirement and not a recommendation or guideline.

danrabbit (commented):
Moving things to reflect what they must be instead of what they must not be is probably good. Not sure "standards" is the right word here. Thanks for working to make this more clear!

src/Dialogs/NewBranchDialog.vala
comradekingu: "rules" certainly isn't, since it is not censorship. Nor do I think it is formatted in the form of rules.
Precisely because it is a technical requirement, (must), it is a standard to meet.

Could shorten it to "compliant" I suppose.

danrabbit: "Rules" definitely doesn't imply censorship in English. The git documentation itself uses the word "rules": http://git-scm.com/docs/git-check-ref-format

jeremypw: I agree "unique" is better.  Maybe "requirements" would be clearer than "rules" or "standards"?

comradekingu: @jeremypw In my head one "meets" a requirement however. Still OK?

jeremypw: I would be OK with "meet Git requirements"

meisenzahl: > Git imposes the following rules on how references are named:

https://git-scm.com/docs/git-check-ref-format

jeremypw:
@ryonakano Which option do you think is easier for translators, "follow Git rules" or "meet Git requirements"?

danrabbit (approved):

PR #1023, opened 2021-05-06
Choose project with headerbar combo and use for new branch action

### LINKED ISSUE #1005: New Branch Dialog: offer a dropdown if focused project is unknown

jeremypw:
Fixes #1005
Replaces  #1015

Instead of showing a dialog to choose the active project only if it is not unambiguous from the circumstances, always show the active git project (if any) in the headbar and allow manual override with a combo.  This is consulted whenever a keyboard shortcut (or potentially a headerbar button) initiates a git action instead of showing a dialog.

The active project keeps in sync with the active document but note that if the active project is changed manually before initiating a git action with the keyboard the project chosen overrides that associated with the active document.

Conversely if a git action is initiated by using a project context menu then the setting in the headerbar is ignored.

 - [x] Populate combo on startup and add new projects when added to sidebar
 - [x] Synchronize combo with active document
 - [x] Preference context menu > combo > to active document.
 - [x] New branch action uses combo to determine target project 
 - [x] Remove projects from combo when removed from sidebar


jeremypw:
ProjectCombo

jeremypw:
There is an issue where the popover associated with the combo does not show the first entry when the second entry is active and there are two repos open. This is because the popover scrolls to show the active entry and shows an arrow to indicate that there is another entry hidden above.  This is even though the size of the popover is large enough to show both entries (when the first entry is active, both entries are shown).  Does anyone know how to fix this?

ryonakano (commented):
Don't forget to add `Widgets/ProjectCombo.vala` to POTFILES :)

hanaral:
It's not as simple, but it might be better to add something like the searchable popover list that Code's language selector uses. Apart from not sucking as much as popovers, it is generally more futureproof'ed because it scales really well with the larger project numbers an IDE would theoretically be used for. On top of that there is also the benefit that it helps the HIG steer away from slapping everything in a combo.

jeremypw:
@hanaral Thats a good idea. The language selector does look better and I can copy the code!

danrabbit:
Closing since we merged the popover

PR #1024, opened 2021-05-07
Show project diff in a temporary document

jeremypw:
Adds an action with shortcut and project context menu option that shows the complete diff of the current project in a temporary document that is displayed in the document view.

The document gets the builtin "diff" syntax highlighting but unfortunately this does not currently include coloration of the added lines as green and the deleted lines as red, which would be useful.

Assistance in fixing this would be helpful.


jeremypw:
Closed in favour of PR merging into git-diff

PR #1025, opened 2021-05-07
Add action to show project diff in a temporary document.

jeremypw:
Based on `git-diff` branch.

Adds an action with shortcut and project context menu action to show the whole project diff in a temporary document in the DocumentView.

The diff document gets the builtin syntax highlighting for ".diff" files but unfortunately this does not include coloration of the added and deleted lines green and red respectively.  This would be useful. Help needed to fix that. Do we need to supply a custom SourceStyleScheme file?

ShowDiffAction


jeremypw:
Closing as not found particularly useful.

PR #1029, opened 2021-05-15
Flatpak: Use OpenURI Portal to launch external apps

Marukesu:
Opening as draft because there isn't a libportal package on focal, however i can update the PR to use conditional compilation if perferred. 

it's functional with project folders added via the terminal and the add project folder button.


jeremypw:
@Marukesu Am I right in thinking that `libportal` is now available in Odin?

Marukesu:
@jeremypw I can't see it with `apt list 'libportal*`.

However, I was thinking in making an alternative PR calling the DBus interface directly like in elementary/calendar#681

It's basically an exchange of dependencies form `libportal` to `gdk-x11-3.0` and `gdk-wayland-3.0`.

jeremypw:
The gdk dependencies sound like they may be provided by the elementary flatpak platform? Not sure how to find out exactly what that provides. Anyway sounds like a good idea.

jeremypw:
Superceded by #1162

PR #1032, opened 2021-05-19
CI: Fix YAML Syntax

Marukesu:
Sorry for not notice this before, this should make the ci runnable again.


danrabbit (approved):
My bad for not reading :sweat_smile:

PR #1037, opened 2021-05-19
ChooseProjectButton: valign center

danrabbit:
Instead of hard coding a margin, we can center align for the same effect


jeremypw:
I used margins after @hanaral 's comment about the buttons height not matching the equivalent buttons in the FormatBar.  I'll check whether this changes that (or perhaps it doesnt matter).

jeremypw:
Taller button than FormatBar

ValighChooseProject

jeremypw:
Sorry - pressed wrong button :facepalm:

jeremypw:
Actually I can't really see the difference

danrabbit:
The headerbar isn't forced to be larger in this branch

PR #1038, opened 2021-05-19
ChooseProjectButton: use max_content_height for scrolled

danrabbit:
So that the popover will shrink to fit contents and max out at 350


jeremypw (approved):
Yes, that's a better way.

PR #1041, opened 2021-05-19
ChooseProjectButton: simplify ProjectRow click handling

danrabbit:
It looks like this `selected` property is a duplicate of `active` but it exists only as a setter to workaround toggling the radiobutton.

Instead, we can use `active` for the setter, but change the signal handling of the radiobutton to `button_release_event` so that `activate ()` is only triggered when the button is actually activated by the user


jeremypw (approved):
Yes, saves some lines and no obvious regressions.

PR #1042, opened 2021-05-19
ChooseProjectButton: use a liststore to populate project list

danrabbit:


jeremypw (approved):
Yes, this is a much better way of doing it - thanks!

PR #1043, opened 2021-05-19
ChooseProjectButton: ellipsize long radio items

danrabbit:
Make sure we also handle ellipsizing radio items

Change ellipsize mode to middle to disambiguate cases like `switchboard-plug-foo` where relevant information is at the beginning and end


jeremypw:
I guess this is required as a consequence of constraining the width of the popover to the width of the button.  Is there any downside to allowing the popover to expand to show the full project name?

jeremypw:
I agree with the middle ellipsize for th button though.

jeremypw:
This is how it looks if you remove the size group and instead use the same character with for the search entry as for the button label.

With a long project name
ExpandPopoverLong
Without a long project name
ExpandPopoverShort

jeremypw (commented):
I have suggested an alternative - I think it would be better to see the full name in the popover.

danrabbit:
@jeremypw Removed the sizegroup instead, but kept middle ellipsize :+1:

jeremypw (approved):
:+1:

PR #1044, opened 2021-05-20
GitManager: source of truth for active_project_path

jeremypw:
Using GitManager as source of truth for the active project saves a useful number of lines and increases clarity.

- We may need a "ProjectManager" class as a controller for project objects in the wider sense (not just git repository objects).? #940  goes down this route. In which vase the active project should be controlled there and there are further simplifications possible in other classes.


danrabbit (approved):

danrabbit:
Yeah, it would probably be better to have a single project manager class in the future, but this at least makes sense for now :)

PR #1045, opened 2021-05-20
Allow non-git projects to fix global search

danrabbit:
Fix global search shortcut by allowing non-git projects in the chooser.

Throw a Gdk.beep if trying to use the new branch shortcut in a non-git project instead of crashing

### DIFF
--- a/src/FolderManager/FileView.vala
+++ b/src/FolderManager/FileView.vala
@@ -164,8 +164,13 @@ namespace Scratch.FolderManager {
         }
 
         public void new_branch (string active_project_path) {
-            string? branch_name = null;
             unowned var active_project = (ProjectFolderItem)(find_path (root, active_project_path));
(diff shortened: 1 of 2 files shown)


jeremypw (commented):
You can still global search within folders (including top-level folders) that are not git projects (or any kind of project) by using the context menu. 

I guess it depends how focussed you want Code to be on code development.  If you want Code to be used on other kinds of unversioned projects then this might be useful but it feels odd to have a list of versioned repositories and unversioned folders in the ChooseProjectButton.

danrabbit:
@jeremypw The case I'm currently testing against is a new app project that doesn't have git init'd just yet. I think it's reasonable to expect that other features—like global search, and eventually compilation tools—will function before a local project is pushed to a git repo

jeremypw:
OK, lets try it.

PR #1046, opened 2021-05-21
Fix Gala 1147: Use present_with_time instead of present to focus window

andreasfelix:
I think this should fix https://github.com/elementary/gala/issues/1147

see https://valadoc.org/gtk+-3.0/Gtk.Window.present_with_time.html

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -129,7 +129,7 @@ namespace Scratch {
                 window.show ();
                 window.restore_opened_documents ();
             } else {
-                window.present ();
+                window.present_with_time ((uint32) GLib.get_monotonic_time ());
             }
 
             // Create a new document if requested


andreasfelix:
Weird, I also tried it change `present` method to `present_with_time` in files 

https://github.com/elementary/files/blob/1e0fffe485365bb4584525b9b6e6252fb8b19f9c/src/Application.vala#L239

and in the terminal

https://github.com/elementary/terminal/blob/8376caa849c2625624b9b6e3d8d738faf31c3170/src/Application.vala#L235

and it fixed the issue in both cases for me. Dumb question: Have you made sure to close all instances after recompiling? 

I have to admit I don't have a deep understanding of what's going on. I just noticed hat almost all elementary apps are using `Gtk.Window.present` to raise the window and that the [Gtk doc](https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-present) mentions that some kind of race condition can happen. At least this how I understand the sentence. Therefore I assumed the `present` method could be villain here. Sorry it didn't work out.

Can you name one of the ubuntu apps which doesn't have this issue?

andreasfelix:
also found this https://codereview.chromium.org/299001/diff/4001/chrome/browser/gtk/options/options_window_gtk.cc

andreasfelix:
No, I tried it again, and the fix does not work for me anymore. I am not able to reproduce...

jeremypw:
Yes, that's weird. I guess if its a race condition then it may not be reproducible.

jeremypw:
This does not fix #1147 as that issue affects most apps, not just Code.  Closing as it would be confusing to fix on a per app basis.

PR #1048, opened 2021-05-22
Show symbol outline pane on right of document

### LINKED ISSUE #1033: Move the outline to the right allowing it to be document-dependant
## Problem

The left pane is not really tied to the document context, outline symbols feels out of place there

## Proposal

Move the outline symbols to the right

## Prior Art

My artistic PR: https://github.com/elementary/code/pull/377

jeremypw:
Fixes #1033

* Move outline code into main code and remove the plugin
* Show outline to the right of the document with matching background
* Add preference option to show outline
* Add action to toggle outline visiblity (`<Alt>\`)
* Remove`Document` signal (`doc_saved ()`) that was only used by the plugin


jeremypw:
SymbolsOnRight

jeremypw:
@danrabbit I would appreciate your advice as to how to get the symbol list (which is a `Granite.Widgets.SourceList`) to follow the styling of the document wrt background and text color.

cassidyjames:
@jeremypw it is probably worth looking at and adapting this method, as setting the background to `@tab_base_color` seems to get us most of the way there:

https://github.com/elementary/code/blob/16d771627b6e04407a6779cc4cc70258e85192c9/src/Widgets/DocumentView.vala#L100-L121

Perhaps renaming that method and defining a GTK CSS variable called `@themed_bg` or something could work, then we could use that in the various places we want to use that color?

jeremypw:
@cassidyjames Thanks for the suggestion! I'll try that.

jeremypw:
@cassidyjames  Implemented your suggestion - thanks.

danrabbit:
If we're moving this anyways, can we make it not be a plugin and just have it be part of the app?

I wonder if we should also add a menuitem to toggle its visibility like we have with the sidebar

jeremypw:
Yes, it should not be too hard to move the code into a "SymbolOutline" subfolder of `\src` and bypass the plugin interface.  As you say, an alternative way of toggling it will be needed.

jeremypw:
The latest commit moves the outline code to the `/src`.  Note that you should run it as a flatpak as it is not compatible with the existing plugin code.  Otherwise you will need to delete the outline plugin from `/usr/lib/x86_64-linux-gnu/io.elementary.files/plugins`.  

Should I include code to blacklist/overwrite the outline plugin in the same way as I did in #1151?

jeremypw:
I have taken the opportunity to DRY the code to reduce duplication between the Vala and C outliners.

jeremypw:
Need to fix some regressions.

jeremypw:
I have added code to overwrite the old plugin when not installed as Flatpak in order to make development and testing easier.  This can easily be reverted if considered unnecessary.

jeremypw:
Removed "blacklist" code as it was not needed when the strip-trailing-spaces plugin was moved into main code.

danirabbit (changes requested):
Looks like there's a completely commented file: src/SymbolPane/meson.build

Otherwise, works as expected and looks good!

danirabbit (approved):

PR #1049, opened 2021-05-23
Global search selected

### LINKED ISSUE #1020: GlobalSearch: The dialog should be preloaded with the selected text (similar to searchbar)
## What Happened

1. Select some text
2. Press `<Ctrl><Shift>F` to open the global search dialog
3. The dialog search entry is empty

## Expected Behavior

The selected text should appear in the search entry in the same way and with the same restrictions as occurs when pressing `<Ctrl>F` for a local search

## Platform Information

Elementary 6.0 daily

jeremypw:
Fixes #1020 

* Use selected text or search entry text when available with global search action

The keyboard shortcut searches text based on which widget is focused.  If the document is focused and text is selected then that is the proposed search term.  If the search entry is focused and has text then that is the search term.

At the moment the dialog still appears in order that the user may override the choice or change settings.


src/Widgets/SearchBar.vala
```diff
@@ -114,6 +116,18 @@ namespace Scratch.Widgets {
             );
             case_sensitive_button.clicked.connect (on_search_entry_text_changed);
 
+            global_search_button = new Gtk.Button.from_icon_name ("gnome-globe") {
+                can_focus = false,
+                tooltip_markup = Granite.markup_accel_tooltip ({"<Control><Shift>F"}, "Global Search"),
```
tintou, line 121:
```suggestion
                tooltip_markup = Granite.markup_accel_tooltip ({"<Control><Shift>F"}, _("Global Search")),
```

tintou:
@danrabbit Can you give some input on which icon to use instead of `gnome-globe` or if we need something new?

danrabbit:
Hm yeah this interface is definitely starting to feel a bit awkward. I wonder if we should move search options into a menu and use checkboxes instead of trying to iconify all of the options inline.

I think it feels awkward to open the search dialog when we already have settings here for case sensitivity and regex. So perhaps if we make global search a checkbox option here we could have the keyboard shortcuts just open this one interface with the global search option either checked or unchecked

jeremypw:
Yeah, I guess you are right - the searchbar could also do global search so the dialog would be redundant. I think originally the dialog had more options but it was slimmed down.  I'll split this PR and keep the code to apply the search to the selected text and lose the extra button pending an interface redesign.

jeremypw:
@danrabbit This PR is now restricted to allowing search on selected text - which is quite useful and quicker than typing search terms in the dialog every time.  At the moment you can select any amount of text which fails with there are more than one line - not sure how to deal with that - search on a term up to the first new line?

danirabbit (approved):
Yeah this works quite well! I think we can probably ignore multiline for now, the dialog makes it clear enough what is about to be searched :+1:

PR #1060, opened 2021-06-02
DocumentView: simplify empty check

danrabbit:
There's only two places that we use `if_empty ()` and both of them only use it to signal `empty ()`, so might as well make `if_empty ()` do the thing

* Renamed `if_empty ()` to `request_placeholder_if_empty ()`, which is admittedly a bit verbose. Open to suggestions
* Renamed `empty ()` to `request_placeholder ()`, which is what this signal really does (for the most part)
* Made `request_placeholder_if_empty ()` signal `request_placeholder ()` so that we don't repeat ourselves

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -410,7 +410,7 @@ namespace Scratch {
                 restore_opened_documents ();
             });
 
-            document_view.empty.connect (() => {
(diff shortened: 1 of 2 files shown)


jeremypw (approved):
I guess on the YAGNI principle there will never be any other reason to check whether are zero docs .

PR #1061, opened 2021-06-03
Get diff gutter colors from SourceStyleScheme if possible

jeremypw:
This PR attempts to get colors defined in the current or fallback SourceStyleScheme for ids "diff:added-line" etc otherwise use hard-coded colors.

Unfortunately of the three styles built into Code, only "classic" contains the relevant style ids and none contain a style explicitly for unchanged lines that replace lines that have been deleted, so the gutter colors do not currently actually change when the style is changed.

Not sure whether it is worth using a fallback SourceStyleScheme or should just immediately fallback to hard-coded colors.

### DIFF
--- a/src/Widgets/SourceGutterRenderer.vala
+++ b/src/Widgets/SourceGutterRenderer.vala
@@ -18,11 +18,11 @@ public class Scratch.Widgets.SourceGutterRenderer : Gtk.SourceGutterRenderer {
     }
 
     public void set_style_scheme (Gtk.SourceStyleScheme? scheme) {
-        update_status_color_map (Services.VCStatus.ADDED, get_style (scheme, ADDED_STYLE_ID));
-        update_status_color_map (Services.VCStatus.REMOVED, get_style (scheme, REMOVED_STYLE_ID));
-        update_status_color_map (Services.VCStatus.CHANGED, get_style (scheme, CHANGED_STYLE_ID));
-        update_status_color_map (Services.VCStatus.REPLACES_DELETED, get_style (scheme, REPLACES_DELETED_STYLE_ID));
-        update_status_color_map (Services.VCStatus.NONE, get_style (scheme, NONE_STYLE_ID, false), false);
+        update_status_color_map (Services.VCStatus.ADDED, scheme, ADDED_STYLE_ID);
+        update_status_color_map (Services.VCStatus.REMOVED, scheme, REMOVED_STYLE_ID);
+        update_status_color_map (Services.VCStatus.CHANGED, scheme, CHANGED_STYLE_ID);
+        update_status_color_map (Services.VCStatus.REPLACES_DELETED, scheme, REPLACES_DELETED_STYLE_ID);
+        update_status_color_map (Services.VCStatus.NONE, scheme, NONE_STYLE_ID, false);
     }
 
     static construct {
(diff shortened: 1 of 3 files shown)


src/Widgets/SourceGutterRenderer.vala
```diff
@@ -1,18 +1,81 @@
...
+    public void set_style_scheme (Gtk.SourceStyleScheme? scheme) {
+            update_status_color_map (Services.VCStatus.ADDED, get_style (scheme, ADDED_STYLE_ID), true);
+            update_status_color_map (Services.VCStatus.REMOVED, get_style (scheme, REMOVED_STYLE_ID), true);
+            update_status_color_map (Services.VCStatus.CHANGED, get_style (scheme, CHANGED_STYLE_ID), true);
+            update_status_color_map (Services.VCStatus.REPLACES_DELETED, get_style (scheme, REPLACES_DELETED_STYLE_ID), true);
+            update_status_color_map (Services.VCStatus.NONE, get_style (scheme, NONE_STYLE_ID, false), false);
+    }
+
+    static construct {
+        fallback_scheme = Gtk.SourceStyleSchemeManager.get_default ().get_scheme ("classic"); // Assume this always exists?
```
danrabbit, line 28: Classic is shipped with Gtk.SourceView, so I think it's safe to assume it always exists. We probably don't need null checks etc for this: https://gitlab.gnome.org/GNOME/gtksourceview/-/tree/master/data/styles

danrabbit (changes requested):
Left a couple of inline comments. This looks good and works as expected here! I can see that it is pulling in the colors from the style and updates correctly when changing styles

Since Gtk.SourceView does ship "classic" as part of the library, I don't think we need any fallbacks other than pulling from this style, and I think we can assume it exists and won't return null etc

jeremypw:
Thanks for the review - I did fear this was somewhat over-engineered.  I'll simplify as you suggest.

src/Widgets/SourceGutterRenderer.vala
```diff
@@ -1,18 +1,81 @@
...
             return project != null ? project.top_level_path : "";
         }
     }
 
+    public void set_style_scheme (Gtk.SourceStyleScheme? scheme) {
+            update_status_color_map (Services.VCStatus.ADDED, get_style (scheme, ADDED_STYLE_ID), true);
+            update_status_color_map (Services.VCStatus.REMOVED, get_style (scheme, REMOVED_STYLE_ID), true);
+            update_status_color_map (Services.VCStatus.CHANGED, get_style (scheme, CHANGED_STYLE_ID), true);
+            update_status_color_map (Services.VCStatus.REPLACES_DELETED, get_style (scheme, REPLACES_DELETED_STYLE_ID), true);
+            update_status_color_map (Services.VCStatus.NONE, get_style (scheme, NONE_STYLE_ID, false), false);
```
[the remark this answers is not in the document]
jeremypw: Removed

src/Widgets/SourceGutterRenderer.vala
jeremypw: Builtin fallback removed, checks removed

jeremypw:
I have increased the gutter width to 5px as I was having trouble distinguishing the more muted classic colors with a 3px gutter.

danrabbit:
Looks great. Nice work!

PR #1067, opened 2021-06-11
Use full path as FileItem and FolderItem tooltip

jeremypw:
Tooltips for the ChooseProject button and Notebook tabs already show the full path.  There does not seem any point in just repeating the item name in the tooltip (default behaviour) so this PR puts the full path in. 

Implements #1068 for sidebar tooltips.


cassidyjames:
The point of showing the name on hover is in case of truncation—it's not _entirely_ pointless. :wink: But yeah, this might be more useful than just always showing the name, and feels more similar to other places.

src/FolderManager/Item.vala
```diff
@@ -36,8 +36,15 @@ namespace Scratch.FolderManager {
             editable = true;
             name = file.name;
             icon = file.icon;
-
             edited.connect (rename);
+            tooltip = Scratch.Utils.replace_home_dir_by_tilde (file.path);
```
cassidyjames, line 40: Nit, but this is a little shorter while still being clear and imho sounds more natural

```suggestion
            tooltip = Scratch.Utils.replace_home_with_tilde (file.path);
```

src/FolderManager/Item.vala
```diff
@@ -36,8 +36,15 @@ namespace Scratch.FolderManager {
...
+
+            notify["activatable-tooltip"].connect (() => {
+                tooltip = ("%s\n" + Granite.TOOLTIP_SECONDARY_TEXT_MARKUP).printf (
+                    Scratch.Utils.replace_home_dir_by_tilde (file.path),
```
cassidyjames, line 44:
```suggestion
                    Scratch.Utils.replace_home_with_tilde (file.path),
```

src/Utils.vala
```diff
@@ -133,4 +133,13 @@ namespace Scratch.Utils {
 
         return sb.str;
     }
+
+    public string replace_home_dir_by_tilde (string path) {
```
cassidyjames, line 137:
```suggestion
    public string replace_home_with_tilde (string path) {
```

cassidyjames (dismissed):
This looks good to me and works. Nice improvement! Just a minor nit on the name of the tilde util.

src/FolderManager/Item.vala
jeremypw: The difference between "replaced with" and "replace by" (if any) is subtle and the Internet does not seem to agree entirely.  One opinion is that where one thing (which is still valid) is substituted by something else equally valid (as here) it is "replaced by" whereas where something becomes obsolete, broken or invalid it is replaced with a newer different thing.  However, I am not convinced there is any material difference so I am happy to accept the alternative.

cassidyjames (approved):

PR #1069, opened 2021-06-11
Style color buttons from theme data

danrabbit:
Based on some work David started in #434 

This styles (and tooltips) color buttons from theme data instead of hardcoding it. There's been some work in the stylesheet to make color buttons work without a ton of custom CSS

This doesn't attempt to make the color buttons customizable, but it's a step in that direction

### DIFF
--- a/data/Application.css
+++ b/data/Application.css
@@ -16,43 +16,6 @@
 * Boston, MA 02110-1301 USA
 */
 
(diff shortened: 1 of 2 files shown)


jeremypw (approved):
Nice. It will be good to be able to access other installed SourceStyleSchemes

Issue #1082, opened 2021-07-19
Make header bar more adaptive, e.g. for small screens and tiling


jeremypw:
## Problem

With a screen size of 1366 x 768px Code cannot be half-tiled because there are too many widgets in the headbar. This makes it difficult to view two document at once. (See issue #1080)

It is noted that Code will half-tile at this size when no documents are open because the headerbar does not then contain the Tab, Syntax Highlighting and Line Number tools.

## Proposal

The Tab and Syntax Highlighting tools are not in continuous use so could be moved into the AppMenu to save space. Alternative some other solution to make the required width smaller could be used.  This could happen at all screen/window sizes (easier) or just when Code detects it is running in a small screen/window.

## Prior Art

Geany moves some tools out of the tool bar and into a menu when the window does not have sufficient width so will half-tile at 1366px screen width.

cassidyjames:
I agree making the headerbar more adaptive for smaller screens and tiling is something we should pursue. That could mean moving indentation and syntax highlighting to the app menu as proposed here, or it could be accomplished differently; as such, I'm going to rename this issue to be more specifically about the problem we're trying to solve.

nevillepark:
Seconded. The "no project selected" field also takes up a lot of space and keeps me from resizing the window any smaller. This is a little annoying as I often have a browser taking up about 75% of the screen and a text editor with notes in the other.

jeremypw:
I think Code is really now intended to a be a "main" app on screen intended for code development rather than a simple text editor. Maybe we should provide a minimalist text editor as well?  It is difficult for an app to try and fulfil two roles.  A solution may be to hide the project selector when the sidebar is hidden?  If you are just taking notes then you do not really need all the code development stuff.

nevillepark:
Tha would be a great solution!

I also agree that elementary needs a basic Notepad-like text editor. I've been looking around and this one, [Eddit](https://github.com/maze-n/eddit), is pretty much exactly what I had in mind. Could it be forked? Some minor things to fix/add:

- line soft-wrapping doesn't work (will make an issue for it)
- could add auto-saving and restoring files on startup, to be in line with how Code acts 
- could add support for tabs and a sepia colour scheme, similar to Terminal

But otherwise it's fine.

nevillepark:
With the new update that hides the project field unless you have the sidebar open, the Code window is more compact!

[...]

PR #1093, opened 2021-08-22
Bind visibility of sidebar and choose_project_button

jeremypw:
Partially addresses #1082 by hiding the Project Chooser button in the header whenever the sidebar is hidden.


danrabbit (changes requested):
Can we add a revealer here so we get a nice animation instead of just binding to visible?

jeremypw:
OK, I'll give it a try - do you want a revealer for the sidebar as well? I don't think it has one.

jeremypw:
@danrabbit I've now used a revealer for the chooseproject button.  Using one for the sidebar proved problematic as it is in a Paned so I've left it for now.

cassidyjames (approved):
Nice. Even without the sidebar animation, this feels pretty slick.

PR #1100, opened 2021-09-07
Fix crash when creating new window from tab drop

### LINKED ISSUE #1095: Crashes when dragging tab over to open a new window
## What Happened

When I drag an open tab over to the desktop to create a new window, a new blank window opens but both crash.

## Expected Behavior

Dragging a tab over should open the file in a new window.

## Steps to Reproduce

1. Open file in a new tab
2. Drag the tab out to the desktop

## Platform Information

ElementaryOS 6

igordsm:
Fix #1095 

There's still some strange situations where the new window has the same file opened twice, but at least it does not crash anymore. I'll try finding a reproducible test case later.

### DIFF
--- a/src/Widgets/DocumentView.vala
+++ b/src/Widgets/DocumentView.vala
@@ -288,7 +288,7 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
     }
 
     private bool find_unique_path (File f1, File f2, out string? path1, out string? path2) {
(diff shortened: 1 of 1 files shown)


jeremypw (approved):
This does fix the crash so is probably worth merging as is.  However, the whole process of creating a new window is broken and needs further work.  The new window contains *all* the previous documents + and extra copy of the dragged document (!)

PR #1101, opened 2021-09-07
Update ChooseProjectButton when a project is closed.

### LINKED ISSUE #1075: Project chooser button not updated when project closed
## What Happened

After closing the project currently shown in the project chooser button, the button still those that project as the active one.  

## Expected Behavior
The project chooser should be updated to show the an appropriate project based on remaining open documents, or if none should be blank.

## Steps to Reproduce

1. Open one or more projects
2. Make a project active in the project chooser button.
3. In the sidebar, close the active project.

WrongProjectShown

## Platform Information

Elementary 6 daily

igordsm:
Fix #1075 . When a project is closed check if it is the currently active project (as selected in ChooseProjectButton) and resets active project to no active project state.

### DIFF
--- a/src/Widgets/ChooseProjectButton.vala
+++ b/src/Widgets/ChooseProjectButton.vala
@@ -92,6 +92,16 @@ public class Code.ChooseProjectButton : Gtk.MenuButton {
             create_project_row
         );
 
+        project_listbox.remove.connect ((row) => {
+            var project_row = row as ProjectRow;
+            var current_project = Scratch.Services.GitManager.get_instance ().active_project_path;
+            if (project_row.project_path == current_project) {
+                label_widget.label = _(NO_PROJECT_SELECTED);
+                label_widget.tooltip_text = _("Active Git project: %s").printf (_(NO_PROJECT_SELECTED));
+                Scratch.Services.GitManager.get_instance ().active_project_path = "";
+            }
+        });
+
         project_listbox.row_activated.connect ((row) => {
             var project_entry = ((ProjectRow) row);
             select_project (project_entry);


src/Widgets/ChooseProjectButton.vala
```diff
@@ -92,6 +92,16 @@ public class Code.ChooseProjectButton : Gtk.MenuButton {
             create_project_row
         );
 
+        project_listbox.remove.connect ((row) => {
+            var project_row = row as ProjectRow;
+            var current_project = Scratch.Services.GitManager.get_instance ().active_project_path;
+            if (project_row.project_path == current_project) {
+                label_widget.label = _(NO_PROJECT_SELECTED);
+                label_widget.tooltip_text = _("Active Git project: %s").printf (_(NO_PROJECT_SELECTED));
```
jeremypw, line 100: I think a more helpful tooltip could be given here.  How about "Open a document in a project folder"? As it is, there is only an "active project" when there is a project folder open in the sidebar AND one of its documents is open.  Maybe this logic needs rethinking as it can be a bit confusing.  There is no direct way of "selecting" a project so "No project selected" probably needs changing too, but that could wait for a different PR.

jeremypw (changes requested):
Does the main job.  Just a query about the tooltip.

src/Widgets/ChooseProjectButton.vala
igordsm: Well, I just copied the default label and tooltip set when the widget is created. How about raising a new issue to discuss this? IMHO the tooltip could show different things when there are project folders but the user hasn't selected an active one and no project folders are opened.

jeremypw: OK, if you want to leave it for another PR that's OK.  This PR is still an improvement.

jeremypw (approved):

PR #1108, opened 2021-10-04
Open README.md when opening folder.

### LINKED ISSUE #437: Open README when adding project
I wonder if it would be valuable to automatically open the project README.md when adding a project to the sidebar? This would mean you get straight to any documentation about the project right away, and if we implement something like #340, we could even automatically open the rendered view of it which would feel like magic.

This would be nice for #135 as well instead of passing in a folder, then getting what seems like your previous session with no changes (except a tiny sidebar item non-obviously added).

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58585486-open-readme-when-adding-project?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61917289&utm_medium=issues&utm_source=github).
</bountysource-plugin>

igordsm:
Fix #437


kaixoo:
This should happen only when you open a folder for the first time, not to affect the "Restore Previous App State" behaviour of elementary

jeremypw:
I am not keen on apps that try to be "helpful" - it is easy enough to open the README.md file if you need to.  This at least needs positive input from the design team to proceed.

jeremypw:
Going to close this as the benefit is not worth adding another option for imo.

PR #1109, opened 2021-10-05
Change setting to Visible whitespace.

### LINKED ISSUE #732: Preferences: Swap "Draw spaces" combobox for "Visible whitespace" switch
Follow-up from https://github.com/elementary/code/pull/722#issuecomment-547048283.

>Design-wise, since we've moved from a ternary option to a binary one, we could also revisit how it's displayed. My thought is to just expose it as "Visible whitespace" with a switch: on is "Always," and off is "For Selection".

igordsm:
Fix #732 . I'm not sure this is the best way of doing this. Please let me know if there's a simpler way.

### DIFF
--- a/src/Dialogs/PreferencesDialog.vala
+++ b/src/Dialogs/PreferencesDialog.vala
@@ -116,11 +116,11 @@ namespace Scratch.Dialogs {
             var line_wrap_label = new SettingsLabel (_("Line wrap:"));
             var line_wrap = new SettingsSwitch ("line-wrap");
 
-            var draw_spaces_label = new SettingsLabel (_("Draw Spaces:"));
-            var draw_spaces_combo = new Gtk.ComboBoxText ();
-            draw_spaces_combo.append ("For Selection", _("For selected text"));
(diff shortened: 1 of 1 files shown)


cassidyjames (changes requested):
Ah, you don't need a label next to the switch; since we don't support "never" showing spaces, we can just treat "for selection" as the default (off position) and then the on position is "always".

igordsm:
@cassidyjames  I'm not sure I follow. You mean keeping the label as "Draw spaces"? If not, could you use the images below to point me in the right direction? (sorry for asking that, BTW). Or maybe point me to a place that looks like what you've described.

__Current dialog:__
Screenshot from 2021-10-05 15-56-11

__Proposed dialog:__

Screenshot from 2021-10-04 20-13-47

cassidyjames:
@igordsm ah, I totally misread the PR. :facepalm:. My bad. I'm testing it now and it looks good to me UX-wise, but we'll want to be sure to get a code review, still.

jeremypw:
At the moment having problems loading documents with this PR - investigating.

PR #1114, opened 2021-10-09
Hide non-active documents when project manually chosen

jeremypw:
This attempts to make handling multiple projects more convenient.  I often find it annoying to have to manually close all the documents from one project when I want to switch to another project (but keep the first project available in the sidebar).  Otherwise there can be a plethora of tabs open.

This PR arranges for all the documents of inactive projects to be closed (but remembered) *when the active project is changed manually with the project chooser button*.  When the active project changes automatically by switching documents then all documents are kept open so it is still possible to work on two documents from different projects if required.

Also any document that is closed temporarily by this PR is still saved and restored.

Need to use this PR for a while to see how useful it is and whether there are drawbacks.


danrabbit:
I haven't looked at the code, but I am in favor of this behavior change!

zeebok (approved):
Code looks good. I haven't tested yet but if you've tested enough to feel confident I say go ahead and merge it in

jeremypw:
I havent stress tested with hundreds of hidden and open docs but it seems stable under modest workload at least. Let others hammer it!

PR #1118, opened 2021-10-19
1092 - Add location and count functionality  to Search and Replace

### LINKED ISSUE #1092: Find and Replace should display instance number of text found
## Problem

When using the Find/Replace option, there is no obvious indication of the "instance number" of the text found, 
such as "1 of 17", "2 of 17", "3 of 17", etc.

## Proposal

The "instance number" of the found text should be displayed when using Find/Replace.

## Prior Art

gedit
Most other text editors have this feature.

jpgleeson:
Fixes #1092 Add current location and count of search string functionality to the Find and Replace bar.

There are still some rough edges, the search buffer has to be created for each change to the search, I've attached the label to update on the GtkSearchEntry search_changed signal, which is a 150ms delay from the last change. This gives enough time for the buffer to be created in most cases, but gives the slight delay before the label updates on a new search. I'm developing on a Pinebook Pro and it's a bit more noticeable on slower hardware like this, but nothing major. I think this is a good solution until I can get a patch upstream to give a callback location when the search buffer is created.

Screen showing search bar with count and location index

Screenshot when there is no search term


jeremypw (approved):
This seems to work pretty well now.  I think it is worth merging and iterating on the design if required.

jeremypw:
Going ahead and merging as no objections have been raised.

PR #1119, opened 2021-10-19
Application: force use portals

danrabbit:
Since Code currently doesn't work great in a sandbox, we can use this environmental variable to force use portals even outside of Flatpak

* Opening files and folders works with the native file chooser
* Code doesn't use notifications
* Trashing files works
* Application chooser doesn't use a portal, but it works


tintou:
What was the problem that you're trying to fix with this environment variable?

cassidyjames (approved):
I can confirm this enables using the file chooser portal for opening a folder and opening a file. I can also confirm Open In and Trash continue to work.

The only strange behavior I'm seeing here is that the file chooser portal seems to try to open away from the Code window (if it's not maximized), or at the top-left corner. But I imagine that's a Gala and/or portal issue.

danrabbit:
Yeah I opened an issue against Files about that. It seems to happen with any maximized window

PR #1121, opened 2021-10-31
Make search results easier to see

sryze:
When you search for the next/previous occurrence, Code scrolls to the line containing the text and makes it the last or first line (depending on search direction). Sometimes it's not easy to spot the result because of this, especially when it's the last line. This patch simply makes it scroll 3 more lines, if possible.


jeremypw (approved):
Code makes sense and works as expected.  Nice fix!

PR #1125, opened 2021-11-13
Rewrite word completion to play nice with other completion providers.

### LINKED ISSUE #1147: Word completion plugin does not handle underscores and hyphens as expected
### What Happened?

If you start typing a word that contains underscores then when you type the underscore and the next letter then the completions starting at the beginning of the word disappear and completion starting with the letter after the underscore appears.

### Steps to Reproduce

1. Open a document containing e.g. the words "Example_one_of_two",  Example_one_of_three" and "Example_two_of_two". 
2. Turn on the word completion plugin
3. Type "Example_" - the plugin shows all three possible words
4. Type "Example_o" -  the completion list disappears or shows unrelated words beginning "o" if present.

### Expected Behavior

Completions should always start with the first letter of the word currently being typed.  As hyphens and underscores are commonly used as part of variable and property names they should be regarded as part of the word, not separators.

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

igordsm:
Fixes #1147 
This makes word completion work with Snippets from #1113 . It's an extensive rewrite of the plugin, so there are probably lots of possible improvements. I've been using it without much problem in the last hours, so I'm sending for review.


plugins/word-completion/engine.vala
```diff
@@ -22,74 +22,58 @@ public class Euclide.Completion.Parser : GLib.Object {
...
-                        list.add (word);
-                    }
-                }
-            }
-        }
-
-        return !list.is_empty;
+    public bool get_for_word (string to_find, out List<string> list) {
+        list = prefix_tree.get_all_matches (to_find);
+        return list.length () > 0;
```
jeremypw, line 46: Its more efficient to test whether the beginning of the list is null than to measure its length.

jeremypw (commented):
This looks good to me.  I'll merge it together with the snippets branch into my "merge-dev-fixes" branch and dogfood it for a day or two.

There are a couple of other things that could be improved in the original but they are not related to this issue so I'll submit separate PRs for those.

jeremypw (changes requested):
One issue I noticed that this PR introduces is that in some cases a prefix matches both a word and the same word followed by a space.

On a blank line type a (new) word
On the next line type the same word followed by a space
On the next line start typing the same word again - the completion shows two entries that are visually the same but one contains an extra space

igordsm:
Just pushed changes that should address this problem, @jeremypw . Thank you for the instructions to reproduce it.

jeremypw:
I can confirm that the above issue with spaces is fixed.  I'll dogfood for a day or two to see if anything else arises.

plugins/word-completion/engine.vala
```diff
@@ -43,7 +43,7 @@ public class Euclide.Completion.Parser : GLib.Object {
 
     public bool get_for_word (string to_find, out List<string> list) {
         list = prefix_tree.get_all_matches (to_find);
-        return list.length () > 0;
+        return list.is_empty ();
```
jeremypw, line 46: GLib.List does not have this function.

igordsm: Strange, I looked up at [valadoc](https://valadoc.org/glib-2.0/GLib.List.is_empty.html) and it exists. VLS also suggests it for me, but build is failing... I'm using Vala 0.54, but didn't expect this to create problems.

igordsm: Confirmed that `is_empty` fails build on vala 0.48.

jeremypw: I notice that it is an inline Vala function - there is no C function`g_list_is_empty`  Maybe it requires a later version of Vala (or GLib).

vjr: @igordsm  I noticed that the previous code was like `length > 0` so the `is_empty` check should have been inverted?

But anyway looks like you replaced it with a different check, not sure how safe it is though.

igordsm: @vjr yeah, I noticed that as well when replacing it. Current code is correct, but thanks for the warning.

plugins/word-completion/engine.vala
```diff
@@ -22,74 +22,58 @@ public class Euclide.Completion.Parser : GLib.Object {
     public const int MINIMUM_WORD_LENGTH = 1;
     public const int MAX_TOKENS = 1000000;
 
-    public const string DELIMITERS = " .,;:?{}[]()0123456789+-=&|-<>*\\/\n\t\'\"";
-    public bool is_delimiter (unichar c) {
+    private Scratch.Plugins.PrefixTree prefix_tree;
+
+    public const string DELIMITERS = " .,;:?{}[]()0123456789+-=&|-<>*\\/\r\n\t\'\"`";
```
jeremypw, line 27: I would not consider hyphen and underscore as delimiters (see my separate PR on this issue).  If you agree, could you change it here?

jeremypw (changes requested):

jeremypw:
@igordsm I'd like to merge this as soon as possible - are you able to address the space/hyphen issue?

jeremypw (approved):

PR #1128, opened 2021-11-15
GlobalSearchDialog: set active_window as transient parent

### LINKED ISSUE #1127: global search dialog isn't centered on parent
### What Happened?

The global search dialog gets centered on screen, not the parent window

### Steps to Reproduce

1. Tile code to one side
2. Open global search dialog
3. see that it opens in the center of the screen, not centered over Code

### Expected Behavior

Dialogs should be centered over their parents

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danrabbit:
Fixes #1127

We can't `get_toplevel ()` from `ProjectFolderItem` so it passes null. We can instead get the active window from Gtk.Application


meisenzahl (approved):
You were faster than me...

LGTM and works as expected 👍

PR #599, opened 2018-11-30
Add GVls Symbol Completion Provider plugin

esodan:
GVls is a Vala Language Server, this plugin is a completion provider, allowing to users to select between a list of releated symbols, like methods or properties.

GVls plugin, parses your files as you open them adding any namespace declared by means of `using`directive, so all symbols for, i.e. GLib namespace, will be available.

Once GLib namespace is loaded, you can know all methods for a string variable you have declared.

Any class you declare as you write or from any already opened file, will be available with its methods and properties.

PR #1132, opened 2021-11-19
Gvls plugin

jeremypw:
Simplified version based on #599 and using GVLs v 20

Required building and installation from  source https://gitlab.gnome.org/esodan/gvls/-/tree/gvls-20 `libgvls` `libgvlsp` `libgvlsui`  

Currently builds and runs and connects to server but no completions provided - terminal shows error message
```
gvlsui-project-manager.vala:93: Error while pushing changes to the server: No stream available to deliver invocation
```


jeremypw:
@esodan Anything obvious why this does not work? I haven't had time yet to dig through the library code.   I was wondering whether it would be better just to get a list of candidates from the server and populate the completion ourselves in the same way the word completion and snippet plugins work.  I do not really like passing the sourceview to an external library.

PR #1133, opened 2021-11-19
Fix various typos

luzpaz:
Found via `codespell -q 3 -S *.po,./po`

### DIFF
--- a/src/Widgets/SearchBar.vala
+++ b/src/Widgets/SearchBar.vala
@@ -298,7 +298,7 @@ namespace Scratch.Widgets {
             search_context.highlight = false;
 
             if (text_buffer == null || text_buffer.text == "" || search_string == "") {
-                debug ("Can't search anything in an inexistent buffer and/or without anything to search.");
+                debug ("Can't search anything in an non-existent buffer and/or without anything to search.");
                 search_entry.primary_icon_name = "edit-find-symbolic";
                 return false;
             }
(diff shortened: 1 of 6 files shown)


src/Widgets/SearchBar.vala
```diff
@@ -298,7 +298,7 @@ namespace Scratch.Widgets {
             search_context.highlight = false;
 
             if (text_buffer == null || text_buffer.text == "" || search_string == "") {
-                debug ("Can't search anything in an inexistant buffer and/or without anything to search.");
+                debug ("Can't search anything in an inexistent buffer and/or without anything to search.");
```
jeremypw, line 301: Should actually be `non-existent`

jeremypw (changes requested):
Thanks!  Just one thing the tool corrected in the wrong way.

src/Widgets/SearchBar.vala
luzpaz: @jeremypw fixed in d651bdd

src/Widgets/SearchBar.vala
```diff
@@ -298,7 +298,7 @@ namespace Scratch.Widgets {
             search_context.highlight = false;
 
             if (text_buffer == null || text_buffer.text == "" || search_string == "") {
-                debug ("Can't search anything in an inexistant buffer and/or without anything to search.");
+                debug ("Can't search anything in an non-existent buffer and/or without anything to search.");
```
jeremypw, line 301: Sorry I should have said that the article will need to be changed as well.
```suggestion
                debug ("Can't search anything in a non-existent buffer and/or without anything to search.");
```

jeremypw (approved):

PR #1135, opened 2021-11-23
Document: Set Filechooser Dialog as Transient for Main Window

Marukesu:
This was the only location where the Gtk.FileChooserNative wasn't created with a parent window, fixes https://github.com/elementary/files/pull/1899#issuecomment-973996508


danrabbit (approved):
Hey there we go! This works, but only in conjunction with your Files branch. Nice job :)

PR #1137, opened 2021-12-04
Fix logic for duplicating unselected line.

### LINKED ISSUE #1136: Duplicate line does not work correctly on first line of document or on blank second line
### What Happened?

Trying to duplicate the first line of a document using the shortcut `<Ctrl>D` failed. When the second line contained text it was duplicated. If the second line was blank then both it and the third line was duplicated.

### Steps to Reproduce

1. Create a new document containing just the line number on each line (1,2,3,4 ...).
2. Place the cursor on line 1 and press `<Ctrl>D`
3. Observe that the document now contains five lines with two lines with "2" (1,2,2,3,4 ...)
4. Create a similar document containing a blank second line (1, _, 2,3,4, ...)
5. Place the cursor on line 1 and press `<Ctrl>D`
6. Observe both 2nd and 3rd line duplicated (1, _, 2, _,2,3,4)

### Expected Behavior

Only the line on which the cursor is place is duplicated.

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1136

### DIFF
--- a/src/Widgets/SourceView.vala
+++ b/src/Widgets/SourceView.vala
@@ -344,7 +344,7 @@ namespace Scratch.Widgets {
 
         // If selected text does not exists duplicate current line.
         // If selected text is only in one line duplicate in place.
-        // If seected text covers more than one line, duplicate all lines complete.
+        // If selected text covers more than one line, duplicate all lines complete.
         public void duplicate_selection () {
             Gtk.TextIter? start = null;
             Gtk.TextIter? end = null;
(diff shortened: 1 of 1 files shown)


jeremypw:
Works correctly with mixed English & Japanese/ characters.

davidmhewitt (approved):
Code makes sense and fixes the linked issue with no other regressions found. Thanks!

PR #1138, opened 2021-12-04
Implement navigation between edit points

### LINKED ISSUE #1018: Introduce shortcuts to find previous/next edit point.
## Problem

It is common to want to go back to the previous point in the document you edited and then back to the current edit point. These are often widely separated. Keyboard shortcuts for this would be useful.  

## Proposal

Introduce a history of edit points per document.
Introduce a "previous-edit-point action with keyboard shortcut to scroll the previous (non-consecutive?) edited line on screen.
Introduce a "next-edit-point" action with keyboard shortcut to scroll the next (non-consecutive) edited line on screen.  If the currently edited line is the latest one, it would be scrolled onto the screen if necessary.

jeremypw:
Fixes #1018  

A new gutter on the left is added.  Clicking on this gutter places a TextMark at the beginning of the corresponding line, indicated by an icon in the gutter. Clicking the icon removes the mark.  

Actions to move back and forward between these marks are introduced linked to hotkeys and context menu options.

Potential improvement: Make optional by adding a setting or converting to a plugin (if possible). Left until review.

APPEARANCE
Edit-point-history


leonardo-lemos (approved):
Nice, works as expected and code looks good.

PR #1139, opened 2021-12-05
863: Set window title to title of currently focused document.

### LINKED ISSUE #863: Set window title to currently-focused file

jpgleeson:
Resolves #863 

Multitasking uses the window title for the title shown in the multitasking view, not the toolbar title as we had set. This changes that assignment so we update the window title when we move between documents.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -413,7 +413,7 @@ namespace Scratch {
 
             document_view.request_placeholder.connect (() => {
                 content_stack.visible_child = welcome_view;
-                toolbar.title = app.app_cmd_name;
+                title = _("Code");
                 toolbar.document_available (false);
                 set_widgets_sensitive (false);
             });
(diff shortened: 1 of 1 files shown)


jeremypw (changes requested):
This works when documents are open but you need to change all the occurrence where the toolbar.title is set to cover the case when there are no documents.

jpgleeson:
> This works when documents are open but you need to change all the occurrence where the toolbar.title is set to cover the case when there are no documents.

Addressed in 45abaf7. I used the translatable string instead of the app_cmd_name for i18n reasons, but I have no strong feelings on it. There is still an assignment to toolbar.title when it is created, but that is harmless.

jeremypw (approved):
LGTM.  `app.app_cmd_name` should not have been used for the window title anyway as it is also used as the name of the plugin set to use and could be changed by using the `-s` commandline option.

Thanks for the fix!

PR #1144, opened 2021-12-06
New Window: do not restore docs, do not duplicate projects

### LINKED ISSUE #1140: Open tab in new window also opens other tabs in the new window.
### What Happened?

Trying to move one of the documents into a new Code window created a window that included all the tabs from the original window and duplicated the moved tab.  This happened both with "Drag out of Window" and context menu methods of creating a new window.

### Steps to Reproduce

1. Open Code
2. Open at least two documents
3. Drag one of the tabs out of the window and drop on the desktop to create a new window
4. The new window contains the dragged tab **and** all the tabs in the original window.

### Expected Behavior

The new window contains only the dragged tab.

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #1141: Project list contains duplicates after creating new window
### What Happened?

With several projects and documents open a tab was dragged to the desktop to create a new window.  The project combobox list was then found to have each entry duplicated.

### Steps to Reproduce

1. Open Code
2. Open at least one project
3. Create a new Code window either by dragging a tab or using the dock context menu
4. Open the project list
5. Observe duplicate entries

### Expected Behavior

Project list should remain unaffected by opening new Code windows.

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1140 
Fixes #1141  (bitesize)


src/Application.vala
```diff
@@ -164,8 +164,8 @@ namespace Scratch {
             return windows.length () > 0 ? windows.last ().data as MainWindow : null;
         }
 
-        public MainWindow new_window () {
-            return new MainWindow (this);
+        public MainWindow new_window () { // Only use for additional windows
```
igordsm, line 167: How about calling this `new_additional_window`? This way we emphasize that it should only be used to create *additional* windows and might not even need the comment.

jeremypw:
Should probably make "New Window" an app action but that can be left to another PR.

src/FolderManager/FileItem.vala
```diff
@@ -30,7 +30,7 @@ namespace Scratch.FolderManager {
         public override Gtk.Menu? get_context_menu () {
             var new_window_menuitem = new Gtk.MenuItem.with_label (_("New Window"));
             new_window_menuitem.activate.connect (() => {
-                var new_window = ((Scratch.Application) GLib.Application.get_default ()).new_window ();
+                var new_window = ((Scratch.Application) GLib.Application.get_default ()).new_additional_window ();
```
danrabbit, line 33: Hm, I guess do we need to do all this? Could it just be:
```suggestion
                var new_window = new MainWindow (Application.get_default (), false);
```

danrabbit (commented):
Functionally this works for me, I have one comment about if we need this method on the application class

jeremypw:
Hmm, I guess I always considered the Application class to be responsible for creating and managing windows.  But in this case the Application class does not do anything other than create the window so I guess it does not need to be there.  I'll check if there are any drawbacks in moving it.

jeremypw:
@danrabbit I have taken the opportunity to simplify window creation a little.

src/MainWindow.vala
```diff
@@ -132,13 +136,25 @@ namespace Scratch {
...
+        public MainWindow (bool restore_docs) {
             Object (
-                application: scratch_app,
-                app: scratch_app,
+                application: (Gtk.Application)(GLib.Application.get_default ()),
                 icon_name: Constants.PROJECT_NAME,
                 title: _("Code")
             );
+
+            document_view.realize.connect (() => {
```
danrabbit, line 146: Is there a reason you moved this to the constructor? I think usually we don't want to have things here unless they are actually specific to this constructor as opposed to `construct` which runs for all constructors

jeremypw: I moved it here because it uses `restore_docs` which is not available in `construct` (afaik)

jeremypw:
Closing as too many changes since this was proposed.

PR #1150, opened 2021-12-08
Fix flatpak git

### LINKED ISSUE #1091: Not showing diff when running as Flatpak
## What Happened

When running as flatpak, code don't show the diff of the opened file. looking at the terminal output, it seems it is triggering a assertion on `ggit_diff_foreach`.

```
Ggit-CRITICAL **: 15:26:01.110: ggit_diff_foreach: assertion 'file_cb != NULL && binary_cb != NULL && hunk_cb != NULL && line_cb != NULL' failed

```
## Expected Behavior

That the diff is shown like in the non-sandboxed version.

## Steps to Reproduce

1. Run code flatpak.
2. Open a file from a git project.
3. Modify the file and see that the diff is not shown.

## Platform Information

elementaryOS 6.0 Early Access.
code from master branch.

jeremypw:
Fixes #1091


davidmhewitt:
This appears to be caused by a bug in the library, upstream fix submitted here:
https://gitlab.gnome.org/GNOME/libgit2-glib/-/merge_requests/33

Marukesu:
> Since we're not shipping Flatpak Code, I think we'd rather wait for a fixed version of the library than make our code look messier

We can also patch the library with the merge request in the manifest until there's a release with the fix.

jeremypw:
Closing as will be fixed upstream

PR #1151, opened 2021-12-08
Move strip trailing whitespace into main code; remove plugin

### LINKED ISSUE #581: Automatically clear tabs/spaces from empty lines?
Feature request: Like Visual Studio, it would be awesome if Code could satisfy my OCD by clearing spacing (and tabs) from empty lines, i.e. lines with no text in them. The company I work for has made it part of their coding standards that blank lines should be completely blank, including any excess spacing.

### LINKED ISSUE #1098: Move Strip Trailing Space from the Extensions tab to Behavior or Interface
## Problem

The so-called extensions are not really extensions but just checkboxes for a particular feature.

## Proposal

They better be proper switches in Behavior or Interface tabs.

jeremypw:
Fixes #1098 

A hopefully minor behavior change is that stripping (and saving) now occurs whenever the document loses focus as well as on manual saving.

Stripping does not occur on every autosave as it could interfere unacceptably with ongoing editing.

An issue raised by this is what to do about the already installed plugin - this PR overwrites it so that it does not appear in the Extensions dialog.  Alternative solutions welcome.

I guess that once the app is supplied only as a Flatpak this will not be an issue and the plugin overwriting code can be removed.


jeremypw:
Bug found - converting to draft while fixing.

jeremypw:
Reduced scope of PR so that current behaviour (strip only on manual save or close) is unaffected.  Stripping on autosave would be easy to add later if required.

jeremypw:
The problem with CI appears unrelated to this PR - it compiles OK locally.

src/Services/Document.vala
```diff
@@ -911,5 +932,62 @@ namespace Scratch.Services {
             warning ("Folder containing the file was unmounted");
             mounted = false;
         }
+
+        /* Pull the buffer into an array and then work out which parts are to be deleted.
+         * Do not strip line currently being edited unless forced */
+        private void strip_trailing_spaces () {
+            if (!loaded || source_view.language == null || source_view.language.id != "vala") {
```
zeebok, line 939: Does this mean this "plugin" code only supports vala files?

zeebok (approved):
This looks good to me

src/Services/Document.vala
jeremypw: Hmm, its a while since I wrote this and I don't recall why I put that restriction in - it does not seem to be in the original plugin.   I guess if working with a language where trailing spaces are significant then you would not have this function active anyway.

jeremypw: I have now removed this restriction to avoid behavior change in this PR.

jeremypw:
A possible refinement is to have the "strip trailing whiltespace" functionality independent of saving so that it occurs even if "autosave" is off.

zeebok:
Hmmm how would that look separated from saving? Remove the trailing white space when a user starts a new line?

jeremypw:
No, I found it too intrusive for stripping to occur immediately after leaving a line. For example,  I like to insert a blank line and the closing brace of an `if () {}` clause then go back and fill in the body of the clause - I do not want the indenting spaces to be removed at that point.   The stripping would still occur when losing focus - e.g. switching to another document or a terminal even if the document is not autosaving.  At the moment with master I have to consciously press `<Ctrl>S` on each document before running `vala-lint`.

zeebok:
Gotcha. Do you think that might not do it often enough? If I write a quick fix, or just code for a short time, in a single doc and close Code it sounds like it wouldn't strip. Not a frequent scenario but one I can picture happening.

jeremypw:
I think #1188  might be relevant here - this could also ensure stripping occurs.

zeebok:
Yeah that seems good. Personally the more places it checks and does the stripping without interfering the better so it ensures a consistent experience. Losing focus, on saves, etc are all good places to me.

jeremypw:
@zeebok Is it OK to merge this PR (which should behave pretty much the same as master, without regressions) and make changes to behaviour in future PRs?

danrabbit (changes requested):
I am in favor of moving plugins into the main source for sure :+1: 

Instead of using the phrase "blacklist" here, can we use "blocklist" instead?

Marukesu:
> An issue raised by this is what to do about the already installed plugin - this PR overwrites it so that it does not appear in the Extensions dialog. Alternative solutions welcome.

I think this should be handled from the package side, for source installations, we should recommend doing an `sudo ninja uninstall` before a new `ninja install`, so that old plugins get removed.

For the deb package, aren't plugins installed in the same package? so when an update occurs, they all got removed and only the ones in the new build installed?

the flatpak package aren't affected by this.

jeremypw:
@Marukesu I didn't realise Code has an `uninstall` ninja target - not all projects (e.g. Files) have one. In that case, and in view of the fact that Flatpaks are unaffected I will remove the relevant code from this PR.  I am uncertain whether just updating a deb will remove old plugins though.

danrabbit:
Yeah the Deb package will cleanup old installed files like plugins so this only affects source installs

zeebok (approved):

PR #1154, opened 2021-12-11
Show hidden folders by default

### LINKED ISSUE #1077: Show hidden folders like .github in folder view
## Problem

If I want to edit a GitHub Actions workflow file (located under `.github/workflows`), the only way to do it is to manually locate it in the **Open a file** chooser. This is because the `.github` folder does not show up in the folder view.

## Proposal

It would be helpful to show hidden folders like `.github` in the folder view itself when a project is open, so that a user can open it directly from there and track its Git status like other files.

Continuing the discussion from #936, this could be achieved by:

- showing all hidden folders in the folder view, possibly controlled by a setting, or
- showing specific hidden folders like `.github` by maintaining a list of useful ones, or
- showing all hidden folders that aren't git-ignored (see prior art below)

## Prior Art

VS Code, for instance, shows all directories in its folder view, but greys out the ones that are git-ignored, which does not include `.github`.

igordsm:
Fix #1077 

Right now this only forces code to show hidden files. No different styling is done.

### DIFF
--- a/src/FolderManager/File.vala
+++ b/src/FolderManager/File.vala
@@ -79,7 +79,7 @@ namespace Scratch.FolderManager {
 
         // Checks if we're dealing with a non-backup directory
         // Hidden subfolders are not shown by default, but we need to allow hidden top-level folder
-        public bool is_valid_directory (bool allow_hidden = false) {
+        public bool is_valid_directory (bool allow_hidden = true) {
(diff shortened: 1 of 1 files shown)


jeremypw (approved):
Yes, lets go with this.  If any issue is raised about being togglable then that can be implemented later.  Thanks!

PR #1155, opened 2021-12-12
Remove global tab control from document FormatBar

jeremypw:
See #1082

The Format control in the header bar contains two controls that are linked to the focused document but the tab control governs global settings and its functions are duplicated in the AppMenu so it does not really belong in the header and takes up valuable space.

This PR just removes it.

In order to further reduce minimum width, the Gtk.Separators, which take up considerable space, are replaced with margins.  This reduces the minimum width to 636px


jeremypw:
Closing as the issue has been addressed elsewhere

Issue #134, opened 2017-06-28
Add fuzzy-finder-like way of opening documents within the active project


cassidyjames:
From Launchpad [bug 1547620](https://bugs.launchpad.net/scratch/+bug/1547620):

>Atom has a great built-in "fuzzy-finder" when you hit Ctrl+T. It opens a crazy-fast search through the open projects' filenames and when you hit enter, it opens that one in a new tab. It's a great way to insanely-quickly jump around in a large project.
>
>I'd love to see something similar in Scratch. :)
>
>Their plugin is here: https://github.com/atom/fuzzy-finder

ssedrick:
I would be interesting in tackling this one. Couple questions though:

Would this be better as a plugin, or as part of the core code?
Atom has it as a plugin with config options. 

Atom tries to use ripGrep under the hood when possible. I've set up vim to use fzf and ag for the searching. Should we have options for which util to use under the hood?
I would default to using `find` though.

jeremypw:
If Code is anything like Files then there are problems assigning a short-cut to a plugin.  Maybe better as part of the code under the Scratch.FolderManager namespace?

dar5hak:
@ssedrick Providing an option to choose a searching util is probably not something that needs to be exposed to the user --- not according to the HIG at least. It's both simpler and more user-friendly to pick one that integrates well, and use that.

CodeforEvolution:
Out of curiosity, is anyone still looking at this issue?

ssedrick:
I haven't been able to get anywhere with this, so if you want to take it, @CodeforEvolution, please do.

CodeforEvolution:
Cool! In that case, I believe this ticket will need a "needs design" label, due to possible UI adjustments that would be a part of this ticket.

Otherwise, here is what I had in mind:
* I would like to make use of the pre-existing toolbar that drops down for text searches
* The toolbar would now have two modes that a user could "toggle" between, using a special toggle switch:
    1. The toggle switch would lie on the far left of the toolbar
    2. The toggle switch's knob would have a magnifying glass graphic on it
    3. The "slots" of the toggle switch would have graphics for each mode: A "T" for text searching, and a page for file searching
    4. When the knob hovers over one of these icons, the icon is highlighted, and the magnifying glass and specific mode icon are "merged"/shown on top of one another.
[...]

PR #1159, opened 2021-12-14
Add a fuzzy search dialog

treagod:
This implements a fuzzy search to solve #134.

I adapted the fuzzy search [from this blog](https://www.forrestthewoods.com/blog/reverse_engineering_sublime_texts_fuzzy_match/) and integrated it into a modal, which can be opened with Ctrl+p,

For now I want to implement the search as a plugin so that the user can opt-in as long the fuzzy search is not mature enough - but I think in the long run this functionality should be part of the core of the editor. 

preview


treagod:
Is there a way to get file change of a project?

So far I can listen to changes in a git repository, but I also want to listen to file changes in a directory which is not under a git root

jeremypw (changes requested):
Sorry for not reviewing this sooner - this is some good work. Ignoring code style issues for now, here are some issues I've noticed:
1) The search dialog shows on pressing the shortcut even if there are no projects loaded.
2) The search results are not disambiguated where there are more than one project loaded
3) Clicking on a search result does not load the file (need to press Enter)
4) The search is a bit too fuzzy for my liking - some apparently poor matches appear before good or exact matches.  It might be better to start with a simpler search i.e. only showing files with basenames starting with the typed letters.  Later PR maybe could show different levels of match in different categories in a similar way to Files search.
5) I wonder whether it would be possible to incorporate filename search into the existing global text search function, or piggyback on it somehow.  Both iterate through project files.  It would then be unnecessary to create a new plugin.

tvallois:
Hello @treagod ,

Are you still working on this? I'd be glad to help if you need to.

PR #1160, opened 2021-12-15
Include hyphen and underscore in word to match

### LINKED ISSUE #1147: Word completion plugin does not handle underscores and hyphens as expected

jeremypw:
Fixes #1147 

Replaces `TextIter.backward_word_start ()` with our own function in order to regard hyphens and underscores as part of the string to match for completions.


davidmhewitt (changes requested):
I can't confirm that this fixes the issue following the steps in your issue report, and with regular code files.

I am compiling this on the daily OS 7.0 images, so perhaps there has been a library change that affects this. Are you able to confirm whether this still fixes the issue for you?

davidmhewitt:
@jeremypw Sorry, what I mean is that this issue is still present on elementary OS 7, and this PR does not seem to fix it. I haven't tried on 6.1 yet.

davidmhewitt (approved):
Apologies for the confusion. It was my mistake in not re-installing the branch so that the new plugin was picked up. I can confirm that this works on both elementary 6.1 and 7, and the code looks good and makes sense.

Thanks for your work.

PR #1161, opened 2021-12-15
MonitoredRepository: Allow Changing Branch with Untracked Files

Marukesu:
Untracked files aren't affected by branch changes, ignore thoses when checking for uncommitted changes.

### DIFF
--- a/src/Services/MonitoredRepository.vala
+++ b/src/Services/MonitoredRepository.vala
@@ -75,7 +75,18 @@ namespace Scratch.Services {
 
         public bool has_uncommitted {
             get {
-                return file_status_map.size > 0;
+                return file_status_map.values.any_match ((status) => {
+                    switch (status) {
+                        case Ggit.StatusFlags.INDEX_NEW:
+                        case Ggit.StatusFlags.INDEX_MODIFIED:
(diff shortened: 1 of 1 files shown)


jeremypw:
What kind of untracked files are you thinking of? I find that untracked files (even new ones) that are matched by a line in the `.gitignore` file do not block changing branch.

jeremypw:
If Git has not been instructed to ignore a file then maybe it is the correct behaviour to block changing branch? I think this is what happens on the command line?

Marukesu:
> What kind of untracked files are you thinking of? I find that untracked files (even new ones) that are matched by a line in the `.gitignore` file do not block changing branch.

Untracked files are the files that exists on the working directory that aren't specified on `.gitignore` neither staged for commit.

For example, doing ` touch dummy.txt` will create an untracked file, and that would block the branch submenu on master. However, `git switch` and `git checkout -b` still work in this case.

jeremypw:
Ah yes, you are right untracked files for not block branch change on the commandline. So this PR is a good idea.  I'll review.

jeremypw (approved):
Does the job - nice fix!

PR #1162, opened 2021-12-17
Use `org.freedesktop.portal.OpenURI` for Providing the AppChooser

Marukesu:
supersedes #1029

Make use of the OpenURI portal for showing an AppChooser and for opening the file on the File Manager.

This one can be used on host too.


src/FolderManager/FileItem.vala
```diff
@@ -36,18 +36,8 @@ namespace Scratch.FolderManager {
...
-            files_item_icon.pixel_size = 16;
-
-            var files_item_grid = new Gtk.Grid ();
-            files_item_grid.add (files_item_icon);
-            files_item_grid.add (new Gtk.Label (files_appinfo.get_name ()));
-
-            var files_menuitem = new Gtk.MenuItem ();
-            files_menuitem.add (files_item_grid);
-            files_menuitem.activate.connect (() => launch_app_with_file (files_appinfo, file.file));
+            var files_menuitem = new Gtk.MenuItem.with_label (_("File Manager…"));
```
jeremypw, line 39: There should not be an ellipsis here as that indicates a submenu or chooser will appear

src/FolderManager/Item.vala
```diff
@@ -76,27 +76,69 @@ namespace Scratch.FolderManager {
...
+                var options = new HashTable<string, Variant> (null, null);
+                options["token_handler"] = Portal.generate_token ();
+
+                var fd = Posix.open (file.path, Posix.O_RDONLY | Posix.O_CLOEXEC);
+                if (fd == -1) {
+                    critical ("OpenURI: cannot open file dscriptor for '%s'", file.path);
+                    return;
+                }
+
+                // the OpenDirectory method was added on version 3
```
jeremypw, line 125:
```suggestion
                // the OpenDirectory method was added in version 3 of the portal
```

src/FolderManager/Item.vala
```diff
@@ -76,27 +76,69 @@ namespace Scratch.FolderManager {
...
-            if (dialog.run () == Gtk.ResponseType.OK) {
-                var app_info = dialog.get_app_info ();
-                if (app_info != null) {
-                    launch_app_with_file (app_info, file.file);
+            var window = (MainWindow) ((Gtk.Application) GLib.Application.get_default ()).active_window;
+            try {
+                var portal = Portal.OpenURI.get ();
+                var options = new HashTable<string, Variant> (null, null);
+                options["token_handler"] = Portal.generate_token ();
+                options["writable"] = !file.is_valid_directory ();
```
jeremypw, line 84: Cannot actually assume the file is writable - if it isn't the menu option fails with a terminal warning.  Should probably make the option insensitive (or remove it) for unwritable files.

jeremypw (changes requested):
A few comments but otherwise this works as expected and the code looks good.

Marukesu:
Added a new property on `FolderManager.File` to query write access, this should work with both read-write and read-only files and directories.

jeremypw (commented):
Sorry for overlloking this for so long.

Is there a reason for the context menu for folder items to differ from that for file items (compared to master running natively)?  The former has an ellipsis and launches straight into the AppChooser whereas the latter has a submenu offering an immediate option to open in the FileManager as well as an "Other applications" option (which offers the File Manager).

Otherwise this is looking good.  Will changing to the OS7 platform require any changes?

Marukesu:
> Is there a reason for the context menu for folder items to differ from that for file items (compared to master running natively)? The former has an ellipsis and launches straight into the AppChooser whereas the latter has a submenu offering an immediate option to open in the FileManager as well as an "Other applications" option (which offers the File Manager).

Since we don't have access to the AppInfo inside the sandbox, the submenu in `FolderItem` had only one option (`AppChooser`), so i removed it, while, `FileItem` ended with three (`New Window`, `File Manager`, `AppChooser`).

In master, the file manager entry is created manually by querying the default AppInfo for `inode/directory`, the portal has the  `OpenURI.open_directory()` method to replace this case, so i used it. Also, i can't see the File Manager in the AppChooser for `FileItem` here.

> Will changing to the OS7 platform require any changes?

You say the flatpak runtime? since we are using Gtk3 here, no changes should be needed. For Gtk4, we need to adapt the `MainWindow.export()` code to use `Gdk.Surface` instead of `Gdk.Window`.

Marukesu:
> I still do not really see why a fileitem and a folderitem should have a different "Open in..." submenu. Can we not offer a "Open in FIleManager" item for folder items as well as file items?

we can, but it's less useful for folders, since the AppChooser will show the FileManager anyway. that not the case for files.
 
> It could make sense to offer a "New Window" option for folders as well as files but that is outside the scope of this PR.

Well, projects folders aren't unique per-window, so any new window will already have all the folders anyway.

jeremypw (approved):

tintou:
Sorry for taking so long for reviewing it, but I don't understand what are you solving here?

tintou:
Implementing the OpenURI portal inside Code is not the way to go, a portal implementation has to be out of the container

src/FolderManager/FolderItem.vala
```diff
@@ -66,6 +66,9 @@ namespace Scratch.FolderManager {
         }
 
         public override Gtk.Menu? get_context_menu () {
+            var open_in_item = new Gtk.MenuItem.with_label (_("Open In…"));
```
tintou, line 69: It has to be something like:
```suggestion
            var open_in_item = new Gtk.MenuItem.with_label (_("Open in another application…"));
```

tintou:
The only way to trigger a portal for opening a file would be:
```
GLib.AppInfo.launch_default_for_uri (file.file.get_uri (), null);
```

tintou:
@danrabbit We should make a blog post about how app developers should replace the in-app "Open In…" menus with something that works in containers

danrabbit:
@tintou yeah the problem is there isn't a replacement portal yet :sweat_smile: https://github.com/elementary/portals/pull/38

jeremypw:
@danrabbit So should this PR be blocked pending a wrapper for the OpenUri portal appearing in Gtk or Granite, which it can then use?

danrabbit:
@jeremypw it seems like generally the OpenURI Portal probably isn't the appropriate portal to use here long-term anyways so this is more of a work around to preserve some functionality inside the sandbox. The proper long-term solution would be to create a sharing portal that's actually designed for this. Since we're not shipping Code in a Flatpak currently, I'm not sure this specific workaround is a huge priority? We could probably wait for a new portal?

jeremypw:
The only "Open in ..." functionality I use is "Open in Terminal".

PR #1172, opened 2022-02-23
Remove project filter

jeremypw:
In master, the presence of the filter when the listbox is bound to a model gives rise to a run-time warning:
  `GtkListBox with a model will ignore sort and filter functions` (though the filter does in fact work).

To avoid this code managing the contents, filtering and sorting of the listbox is moved to the ChooseProjectButton from GitManager and the listbox is no longer bound to a model.

A better solution may be possible with Gtk4, using a FilterListModel.


danrabbit:
I know this anectdotal, but I have nearly 60 projects open currently, so the search I think is useful :sweat_smile: I guess the proper way to suppress this warning is to set the filter on the model instead of the list

jeremypw:
Thanks for the review! I found more than a few projects open at once unwieldy so I did not envisage so many would be kept open (probably my machine is too slow).  I'll rethink this.

jeremypw:
So the main use for "Active Project" is as the target for shortcuts performing Git actions:

* New Branch
* Global Search (global within the active project)

Maybe this should be made clearer in the UI?  Is there anything else this concept could be useful for?

jeremypw:
Just realized that #1114 does in fact give another use to the Active Project (when selected with the Chooser button) - it collapses the other projects and hides their documents

danrabbit:
Yeah, I really liked your branch that swaps out the tabs when you change the project selection.

Originally, my thought was to only show the file tree from the active project in the sidebar since I have like 40 collapsed tree headers to scroll through to find my open project 😅

jeremypw:
There is a valid use for having documents from more than one project open at a time - I do this quite often to copy code from one to the other.  Another thought I had was to make non-active documents read-only and non-focusable which would mean you *have* to use the ProjectChooser button, rather than the active project automatically changing to match the focused document.

PR #1174, opened 2022-03-08
Fix occasional crash in SearchBar

### LINKED ISSUE #1173: Crash occasionally when searching
### What Happened?

When performing searches in a fairly large project (Gala), Code periodically hangs for several seconds and occasionally crashes.

The following backtrace was obtained from gdb:
```
Thread 1 "io.elementary.c" received signal SIGSEGV, Segmentation fault.
0x00007ffff76c8594 in ?? () from /lib/x86_64-linux-gnu/libgtk-3.so.0
(gdb) bt
#0  0x00007ffff76c8594 in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#1  0x00007ffff76e16d4 in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#2  0x00007ffff76e6c29 in gtk_text_iter_compare () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#3  0x00007ffff70c5abd in  () at /lib/x86_64-linux-gnu/libgtksourceview-4.so.0
#4  0x00007ffff70c9865 in gtk_source_search_context_get_occurrence_position () at /lib/x86_64-linux-gnu/libgtksourceview-4.so.0
#5  0x00005555555d333c in __lambda101_ (self=0x555555c665d0) at ../src/Widgets/SearchBar.vala:540
#6  0x00005555555d360b in ___lambda101__gsource_func (self=0x555555c665d0) at SearchBar.c:2873
#7  0x00007ffff7edbbe8 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#8  0x00007ffff7edb04e in g_main_context_dispatch () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#9  0x00007ffff7edb400 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#10 0x00007ffff7edb4a3 in g_main_context_iteration () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#11 0x00007ffff7d8afe5 in g_application_run () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#12 0x000055555557494d in scratch_application_main (args=0x7fffffffde18, args_length1=1) at ../src/Application.vala:172
#13 0x000055555557499a in main (argc=1, argv=0x7fffffffde18) at ../src/Application.vala:171
```

### Steps to Reproduce

1. Use Code to perform searches on a large project with many tabs open
2. Wait for it to crash (may take a long time).

### Expected Behavior

Code should not crash.

### OS Version

6.x (Odin)

### Software Version

### Log Output

```shell
See bug description
```

### Hardware Info

Hardware

jeremypw:
Fixes #1173 

Check null search_context before using; 
Cancel search occurence label update before changing context or textview

This PR should not change any functionality.

### DIFF
--- a/src/Widgets/SearchBar.vala
+++ b/src/Widgets/SearchBar.vala
@@ -264,8 +264,8 @@ namespace Scratch.Widgets {
         }
 
         private void on_search_entry_text_changed () {
-            if (search_context == null) {
-                critical ("search entry changed with null context");
+            if (search_context == null) { // This can happen during start up
+                debug ("search entry changed with null context");
                 return;
             }
 
--- a/src/Widgets/SearchBar.vala
+++ b/src/Widgets/SearchBar.vala
@@ -354,7 +354,7 @@ namespace Scratch.Widgets {
         }
 
         public void highlight_none () {
-            if (search_context == null) {
+            if (search_context != null) {
                 search_context.highlight = false;
             }
         }
--- a/src/Widgets/SearchBar.vala
+++ b/src/Widgets/SearchBar.vala
@@ -187,12 +187,17 @@ namespace Scratch.Widgets {
         }
 
         public void set_text_view (Scratch.Widgets.SourceView? text_view) {
+            cancel_update_search_occurence_label ();
+            this.text_view = text_view;
+
             if (text_view == null) {
                 warning ("No SourceView is associated with SearchManager!");
+                search_context = null;
                 return;
+            } else if (this.text_buffer != null) {
+                this.text_buffer.changed.disconnect (on_text_buffer_changed);
             }
 
-            this.text_view = text_view;
             this.text_buffer = text_view.get_buffer ();
             this.text_buffer.changed.connect (on_text_buffer_changed);
             this.search_context = new Gtk.SourceSearchContext (text_buffer as Gtk.SourceBuffer, null);
--- a/src/Widgets/SearchBar.vala
+++ b/src/Widgets/SearchBar.vala
@@ -373,6 +373,7 @@ namespace Scratch.Widgets {
         private bool search_for_iter (Gtk.TextIter? start_iter, out Gtk.TextIter? end_iter) {
             if (search_context == null) {
                 critical ("Trying to search forwards with no search context");
+                return false;
             }
 
             end_iter = start_iter;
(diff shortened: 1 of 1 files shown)


src/Widgets/SearchBar.vala
```diff
@@ -338,20 +354,27 @@ namespace Scratch.Widgets {
...
+
             bool has_wrapped_around;
             Gtk.TextIter? start_iter, end_iter;
             text_buffer.get_start_iter (out start_iter);
             return search_context.forward (start_iter, out start_iter, out end_iter, out has_wrapped_around);
         }
 
         private bool search_for_iter (Gtk.TextIter? start_iter, out Gtk.TextIter? end_iter) {
+            if (search_context == null) {
+                critical ("Trying to search forwards with no search context");
```
davidmhewitt, line 375: I would say this is a warning and the method should return in this case? `search_context` is used inside this method and that `null` access is going to cause issues. So better to log a warning and return from the method than log a critical and potentially crash and lose work?

src/Widgets/SearchBar.vala
```diff
@@ -369,6 +392,10 @@ namespace Scratch.Widgets {
         }
 
         private bool search_for_iter_backward (Gtk.TextIter? start_iter, out Gtk.TextIter? end_iter) {
+            if (search_context == null) {
+                critical ("Trying to search backwards with no search context");
```
davidmhewitt, line 396: Same here

davidmhewitt (changes requested):

jeremypw:
@davidmhewitt  Thanks for the review and for spotting those missing returns!

jeremypw:
A critical message is downgraded as it gets triggered during normal startup due to a toolbar button state being bound to settings before an active document is focused.  It is easier to ignore this rather than try and ensure it doesnt happen.

src/Widgets/SearchBar.vala
```diff
@@ -338,20 +354,28 @@ namespace Scratch.Widgets {
         }
 
         public void highlight_none () {
-            search_context.highlight = false;
+            if (search_context == null) {
```
davidmhewitt, line 357:
```suggestion
            if (search_context != null) {
```

This if statement appears to be inverted compared to its indended use?

davidmhewitt (changes requested):

src/Widgets/SearchBar.vala
jeremypw: Oops!

PR #1177, opened 2022-03-24
Only focus the current document on open

### LINKED ISSUE #1176: Wrong file focused after opening
### What Happened?

I created a new empty file in Files and opened it in Code to edit. When Code opened the empty file in front of me, I started typing, but no text appeared. Then I noticed an asterix on the tab of another previously opened file, indicating it has been modified. Switching to that tab (first on the left btw) I could see the text I was typing was added at the end of that file.

After a bit of testing I noticed **it only happens when I open the file using the context menu**.

### Steps to Reproduce

First, have Code with some files already open in tabs, and have Code closed.

1. Create a new empty file in Files.
2. Open this file in Code **using the right-click context menu** (I cannot reproduce this with just `Enter`).
3. When Code opens, start typing.
4. You should see no text added, but some other tab (probably the left-most tab) will indicate modification with an asterisk.
5. Switch to that tab to see your newly added text.

Snímek z 2022-03-24 11-04-39

### Expected Behavior

The file I am opening should be in focus and I should be able to immediately start typing into it. The changes should not appear in a different tab.

### OS Version

6.x (Odin)

### Software Version

### Log Output

```shell
NA
```

### Hardware Info

Snímek z 2022-03-21 10-48-39

davidmhewitt:
Fixes #1176 

There's a race condition here due to the `Idle.add`. Documents can be restored from saved state, one of which will have `focus == true`. At the same time, we could have started Code with an argument to open another document (i.e. "Open With" from Files), this will also have `focus == true`.

So instead of focusing all documents that have `focus == true` in an unpredictable manner due to the `Idle.add`, only focus the one that's set to `current_document` above in the synchronous code.

### DIFF
--- a/src/Widgets/DocumentView.vala
+++ b/src/Widgets/DocumentView.vala
@@ -202,7 +202,7 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
         Idle.add_full (GLib.Priority.LOW, () => { // This helps ensures new tab is drawn before opening document.
             doc.open.begin (false, (obj, res) => {
                 doc.open.end (res);
-                if (focus) {
+                if (focus && doc == current_document) {
                     doc.focus ();
                 }
 


jeremypw (approved):
Makes sense, thanks.

PR #1178, opened 2022-03-24
Update appdata with recent changes

davidmhewitt:
Noticed that none of the recent changes to Code have been captured in the AppData yet. There's some good stuff in here, so we should probably prepare a release soon.

This includes a line for #1177 which hasn't been merged yet, so ideally would be merged after that.


jeremypw (approved):
Thanks for doing this.

PR #1180, opened 2022-03-31
replace_home_with_tilde: use `substring()` to remove home directory

Marukesu:
`slice()` was removing the last character of the path.


davidmhewitt (approved):
Thanks!

PR #1184, opened 2022-04-03
Fix autosave during close

### LINKED ISSUE #1183: Autosave of unsaved buffers not reliable
### What Happened?

https://youtu.be/LrOn42Y4aIE

This works with other editors like Atom and Sublime Text. It is an essential feature for me!

### Steps to Reproduce

1. Open Code
2. New File
3. Type "something"
4. Close Code
5. Open Code

### Expected Behavior

The typed text should be saved.

### OS Version

Other Linux

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

vjr:
Fixes #1183 

Block on source buffer async save instead of yield.


src/Services/Document.vala
```diff
@@ -369,9 +374,14 @@ namespace Scratch.Services {
...
+            }
+
             bool ret_value = true;
-            if (Scratch.settings.get_boolean ("autosave") && !saved) {
-                save_with_hold ();
+            if (Scratch.settings.get_boolean ("autosave") && (!saved /*|| source_view.buffer.get_modified ()*/)) {
+                save_with_hold (app_closing);
             } else if (app_closing && is_file_temporary && !delete_temporary_file ()) {
                 debug ("Save temporary file!");
                 save_with_hold ();
```
jeremypw, line 387: Should pass `app_closing` parameter here too?

src/Services/Document.vala
```diff
@@ -454,15 +464,27 @@ namespace Scratch.Services {
             this.create_backup ();
 
             // Replace old content with the new one
-            save_cancellable.cancel ();
+            if (save_cancellable != null) {
+                save_cancellable.cancel ();
+            }
```
jeremypw, line 469: Maybe better to cancel saving timeout here rather than earlier in order to catch all cases?

jeremypw:
I'll see if I can come up with something.

jeremypw:
Just pushed a draft PR that makes more deep-rooted changes to the saving logic (which seems rather tangled at the moment). Unfortunately the diff is larger so harder to get merged.  Please see if you can find any regressions if you have time.

vjr:
Thanks Jeremy! I'll take a look, probably this weekend.

jeremypw:
Closing as the linked issue is marked closed by another PR.

PR #1188, opened 2022-04-06
Ensure changes saved when closing

### LINKED ISSUE #1183: Autosave of unsaved buffers not reliable

jeremypw:
Fix #1183 

This makes most saving related functions explicitly asynchronous and where possible simpler (fewer parameters/conditionals).
When closing the app, saving changes is decoupled from actually closing the documents so that any name changes during saving are properly remembered.
The delete signal is blocked - the app can only be destroyed after changes have finished being saved. Obviously, if the app receives the KILL signal changes will not be saved so logging out with unsaved changes will not save them.

A function was renamed to make its action clearer and less ambiguous.

Closing a tab or the app immediately after editing a file should save the changes properly, if necessary showing a dialog (which can stop the app closing).

Stripping trailing spaces does not occur during autosave but always occurs before saving in other circumstances (subject to settings), in which case it no longer triggers a further unnecessary autosave.

Changes exposed a few places where a handler id or timeout id was not zeroed after disconnection or removal causing non-fatal terminal errors; these were fixed.


vjr:
Hello Jeremy, I did a quick test (just edited a text file and quickly closed the app via the close button and also via CTRL+Q shortcut) and it appears to be working, except the "remember opened files" part apparently. If I open a second document and come back to the first test document and edit and close, then subsequent launch of the app returns to the second document and not the first one which I had just edited and closed. The saving part is working though, with no leftover temporary `.goutputstream-XXX` files.

As for the code changes , I only skimmed through and looks like this refactoring is good to have even if it is a bit large, subject to approval from a reviewer more familiar with the codebase.

jeremypw:
Thanks for the quick test Vishal - I'll look into that issue you found.  I think there might be another issue whereby the number of holds on the app does not reduce to zero under some circumstances as well.  I'll leave as draft until more thoroughly tested.

jeremypw:
Just managed to generate a `.goutputstream-XXX` file somehow :disappointed: Not sure how to reproduce yet.  More work needed ...

jeremypw:
Last commit fixes the incorrect focused document on restore issue.

jeremypw:
Not sure how I managed to get duplicate commits!

zeebok (approved):
Nothing stands out. This seems like a really good set of changes!

jeremypw:
@zeebok Thanks for the review!  It has been automatically dismissed I am afraid as I pushed some further commits. Hopefully no further changes will be needed but I intend to test it over the next day or two.

jeremypw:
Last commits introduced a serious bug I need to track down ...

jeremypw:
Closing in favour of a new PR consolidating saving code into DocumentManager, simplifying where possible.

Issue #666, opened 2019-04-15
Help info is not displayed correctly in terminal


jeremypw:
Related to #488 but happens in English environment and formatting is worse.

![Screenshot from 2019-04-15 19 15 15](https://user-images.githubusercontent.com/10513844/56155476-e3ab9100-5fb2-11e9-90d5-2f702dac4760.png)

ryonakano:
@jeremypw I'm sorry, but the title of #488 which I made was not accurate. I wanted to mean that this issue happens if we're running our system in any localized languages, but I completely forgot that there are some localized Englishes like en_GB, en_AU, en_CA, and so on. So I think this issue is duplicate of #488.

jeremypw:
OK, I guess I forgot that English English is a localization!

PR #1192, opened 2022-05-11
Follow system style toggle

### LINKED ISSUE #902: Provide an auto dark style option
## Problem

For the majority of default apps, they will always either prefer dark or adopt to the system dark style preference. Code is a little different though because its style preferences are also tied to syntax highlighting. Changing the syntax highlighting out from underneath a developer could be considered a bad thing.

## Proposal

We should add a toggle switch to Code, much in the style of "Natural Copy Paste" in Terminal that will follow the system dark style preference by default, but enable users to manually choose their preferred color scheme instead

Oowoosh0:
Fixes #902 

Adds a follow system style toggle to the preference menu as mentioned in #902 
Default is off, if enabled Code follows the global dark style setting.

It uses solarized-light for syntax highlighting if follow system style is set to true and system style is set to default. So currently there is no way to use the classic theme for syntax highlighting while having follow system style enabled. But to make that possible one would have to separate the choice of color scheme for syntax highlighting and gtk dark style.


jeremypw:
FollowSystemStyleSwitch

There needs to be some padding between the new switch and it label.

jeremypw (changes requested):
The code looks clean to me and works as expected.  Just some padding required between the switch and the label.  Needs input from design team as well.

jeremypw:
Although the UI closely follows the corresponding Terminal PR (#666) there is a significant difference in behaviour.  Terminal persists the user preference when the "follow-system-style" is toggled - i.e. the original appearance is restored when the switch is toggled off.  This PR does not persist the original setting but overwrites it.  I think Terminal's behaviour is better.

PR #1194, opened 2022-05-13
Fix drag and drop of accented text and between tabs

### LINKED ISSUE #1191: Drag and drop of selected text not working with accented characters
### What Happened?

Drag and drog of accented characters transforms them into their unicode equivalent

For example, **légèreté**  turns into **l\u00e9g\u00e8ret\u00e9** 

### Steps to Reproduce

1 - Select text with accented characters
2 - Drag selected text to other location within document
3 - after dropping text, accented characters are converted to their unicode

### Expected Behavior

words to be moved 'as is'

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

This has been happening on other computers and older OS versions...

### LINKED ISSUE #1193: Dragging text into document in a second window duplicates it
### What Happened?

Two windows of Code were open and text was dragged from one window to the other.  The text dropped into the second window was duplicated.

### Steps to Reproduce

1. Open two windows of Code
2. Open a text document in the first window
3. Create a new document in the second window
4. Select some text in the first window and drag/drop it into the new document
5. Observe that the text is pasted twice into the new document.

### Expected Behavior

The dropped text is the same as that selected, not duplicated.

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1191 
Fixes #1193 

* Use native Gtk.SourceView DnD handling for buffer contents;
* Handle DnD of uri lists with DocumentView

The linked issues are caused by resetting the source view as a drag destination to handle uri lists but without handling text properly.  Attempts were made to handle uris with the sourceview without causing issues for buffer contents but these failed.  It is much easier to keep the native DnD handling of the Gtk.SourceView and handle uri lists with the DocumentView.

The effect of this change is that in order to create a new document by dropping a file item (e.g. from Files) the target is the "Add Tab" button on the tab bar.  Dropping elsewhere (e.g. on the textview) drops the uri as text.

In my opinion, this change is acceptable, despite the smaller target for uris, since (1) Dropping a file onto a sourceview has an ambiguous intent (drop uri as text? drop file contents as text? create new document???) and (2) it is not likely to be a commonly used action anyway.


zeebok (approved):
Code looks good and works as described. My only thought with testing is maybe also open the file in a new tab if the file is dropped anywhere in the tab area instead of only the new document button.

jeremypw:
@zeebok  Actually, you can also drop files onto the blank area of the tabbar to create a tab with this PR.  Changes would have to be made to the Granite library to implement dropping onto an existing tab (and it is not clear what the expected behaviour would be in that case). As Granite DynamicNotebook is likely to be replaced in the not too distant future I do not think it worth progressing this further.

PR #1199, opened 2022-06-22
Explicitly handle missing cases

jeremypw:
To silence terminal compile warnings on OS7.


tintou:
Wouldn't a `default:` be the same here?

jeremypw:
In a previous similar PR, Dani preferred explicitly listing the the cases so that a new warning would appear if a new case was introduced.

danrabbit (approved):

PR #1204, opened 2022-08-13
EditorConfig plugin overrides global settings for indent width and style

### LINKED ISSUE #1203: Editorconfig is no longer respected
### What Happened?

This is probably a regression due to #1145 

Tab width is assumed to be global for the editor when different document types can have different tab widths (such as vala vs xml)

### Steps to Reproduce

1. Open a project with an .editorconfig file that specifies tab width for a certain file type (like 2 spaces for xml)
2. Open an xml file and see that the header says 4 spaces
3. Hit tab in the view and see that you get 4 spaces instead of 2

### Expected Behavior

For projects with editorconfig, we should always follow the editorconfig file.

Probably the most user friendly behavior would be to write to the .editorconfig file if a user changes a setting in the UI.

But the simplest solution would be to make any options overruled by .editorconfig insensitive (such as tab width and insert spaces options)

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1203

* [x] Specify name of editorconfig file so that plugin actually works
* [x] Link editorconfig plugin and format bar
* [x] Expose functions to change indent width and style of current document.
* [x] Make format bar aware if indent width and/or style set by plugin
* [x] If editorconfig plugin sets indent width and/or style, make the formatbar indent button insensitive.
* [x] Amend tab button text and tooltip text accordingly
* [x] Prevent FormatBar changing global settings (this is done via Preferences) 
* [x] Amend `.editorconfig` file so that it does not duplicate/conflict with global setting of indent width and style.

It was noticed that although `indent-width` and `tab-width` are separate properties of SourceView, there is no facility in the existing code for these to be set separately so these are treated the same by the plugin (and only one specified in the config file).  If both were specified in a config file then the last one would have priority. If there are situations where these need to be set separately additional code will be required.

There is no facility to update the `.editorconfig` file through the UI (except by manual editing).  That could be added as a future PR if required.

Input required on the user facing copy.


src/Widgets/FormatBar.vala
```diff
@@ -241,11 +238,41 @@ public class Code.FormatBar : Gtk.Grid {
...
+    }
+
+    public void set_tab_width (int indent_width) {
+        if (space_tab_switch.active) {
+            tab_toggle.text = "%d (Space)".printf (indent_width);
+        } else {
+            tab_toggle.text = "%d (Tab)".printf (indent_width);
+        }
+
+        tab_toggle.text = "%d (%s)".printf (
```
danrabbit, line 260: It seems like this if/else is immediately overwritten right?

danrabbit (changes requested):
Hm, this doesn't seem to work for me. When I open an xml file in a project with an editorconfig (such as code) I still get 4 space indent instead of 2

src/Widgets/FormatBar.vala
jeremypw: Oops - looks like both alternatives were left in.  I will remove one.

jeremypw:
@danrabbit  It works for me under those circumstances.  You have to have 
```
[{*.xml,*.xml.in,*.yml}]
tab_width = 2
```

in the editor-config file (*after* the [*] setting)

jeremypw:
Also you have to install the PR in order to update the plugin.

jeremypw:
Maybe this is another candidate for moving into the main code?

danirabbit (approved):
Works as expected, thanks for fixing this!

PR #1207, opened 2022-10-01
Fix meson warning for run_command

turbobastii:
When running meson build, there is a warning about run_command changing its defaults in the future.
(See: https://github.com/mesonbuild/meson/issues/9300)

~This fixes the warning while keeping the current default.~ UPDATE: after the feedback, I changed the check to `true`.

Without fix:
image

With fix:
Screenshot from 2022-10-01 09-13-24


plugins/outline/meson.build
```diff
@@ -14,7 +14,7 @@ module_files = [
 ]
 
 # We need libvala-X.XX library, but it changes depending on the version that is installed
-vala_version = run_command (meson.get_compiler('vala'), '--api-version').stdout().strip()
+vala_version = run_command (meson.get_compiler('vala'), '--api-version', check: false).stdout().strip()
 vala_dep = dependency('libvala-@0@'.format(vala_version))
```
eli-schwartz, line 18: > This fixes the warning while keeping the current default.

But the entire point of the warning is that the current default is usually bad, and it's provably bad here: if the command fails, then you try to find the `libvala-` dependency, which is totally broken and will fail. Of course, it's unlikely that once valac is found it will error out. But, nevertheless, this should absolutely be `check: true`.

turbobastii: Makes perfect sense, thanks for pointing it out. Changed it accordingly.

tintou (approved):

PR #1217, opened 2023-01-08
Put the ProjectChooserButton inside the sidebar

danirabbit:
Since we hide and show this with the project sidebar, put it in the sidebar


jeremypw (approved):
Not too sure about the start and end margins but that is not blocking.

PR #1218, opened 2023-01-08
Make sidebar pane the toplevel container

danirabbit:
Closes #1217 

Fits the more modern design we've been moving towards with other apps:

Screenshot from 2023-01-08 10 05 14
Screenshot from 2023-01-08 10 05 39


jeremypw:
@danirabbit Please could you fix conflicts, thanks.

danirabbit:
@jeremypw done!

jeremypw (approved):
The only issue I have with this design is that the close window button is no longer visually separated from the sidebar so it looks like a "close sidebar" button. Not sure what the solution is to this (if one is needed) and it is not blocking for me.

Code looks good.  

The function of the project chooser button probably needs review. It always reverts to the project related to the focused document regardless of what selection is attempted.  However, that is for another PR.

PR #1219, opened 2023-01-08
HeaderBar: put find actions in menu

danirabbit:
Screenshot from 2023-01-08 12 43 51

Moves find actions to the menu like Epiphany. Should help towards #1082. Gives a place to expose the global search shortcut


jeremypw (approved):
Code looks good, works as expected without regression.

PR #1222, opened 2023-01-09
Use elementary styles when following system style

### LINKED ISSUE #1214: Following system styles uses solarized dark, not elementary dark
### What Happened?

When following system style and having the dark preference set, solarized dark is used. When follow system style is off, the dark style offered is different

### Steps to Reproduce

* dark theme on
* tick and untick follow system style 

### Expected Behavior

There should be a single dark style, or it should be a preference so you can have the new dark style when following system prefs.

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #1214


danirabbit:
Just gonna go ahead and merge this since it's a pretty obvious fix

PR #1224, opened 2023-01-10
Headerbar: code style fixes

danirabbit:
Review with "hide whitespace changes"

* Inline namespace
* Scope down permission to toolbar items
* Use box instead of grid
* Code style


src/Widgets/HeaderBar.vala
```diff
@@ -19,364 +18,373 @@
...
-            Object (
-                has_subtitle: false,
-                show_close_button: true
-            );
-        }
+public class Scratch.Widgets.HeaderBar : Hdy.HeaderBar {
+    public Code.FormatBar format_bar { get; private set; }
+    public Gtk.Button templates_button { get; private set; }
+    public Gtk.MenuButton share_app_menu { get; private set; }
+    public Gtk.Menu share_menu { get; private set; }
```
jeremypw, line 25: I think these can all be `{ get; construct; }` . Compiles and runs if this change is made.

src/Widgets/HeaderBar.vala
```diff
@@ -19,364 +18,373 @@
...
+
+        var revert_button = new Gtk.Button.from_icon_name ("document-revert", Gtk.IconSize.LARGE_TOOLBAR) {
+            action_name = MainWindow.ACTION_PREFIX + MainWindow.ACTION_REVERT
+        };
+        revert_button.tooltip_markup = Granite.markup_accel_tooltip (
+            app_instance.get_accels_for_action (revert_button.action_name),
+            _("Restore this file")
+        );
+
+        share_menu = new Gtk.Menu ();
```
jeremypw, line 79: `share_menu`  and `share_app_menu` were introduced in 2013 but seems to have never been used.

src/Widgets/HeaderBar.vala
```diff
@@ -1,7 +1,6 @@
-// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
 /*
-* Copyright (c) 2013 Mario Guerriero <mefrio.g@gmail.com>
-*               2017–2018 elementary, Inc. <https://elementary.io>
+* Copyright 2013 Mario Guerriero <mefrio.g@gmail.com>
+*           2017–2022 elementary, Inc. <https://elementary.io>
```
jeremypw, line 3: should be 2023?

jeremypw (changes requested):
Just a few small comments inline otherwise looks good.

src/Widgets/HeaderBar.vala
danirabbit: None of these are constructor properties though, we don't currently pass any arguments into the constructor

src/Widgets/HeaderBar.vala
Marukesu: IIRC, they are used by the pastebin plugin.

jeremypw: Your right.  I did a global search for `share_app_menu` but not `share_menu` :disappointed:

src/Widgets/HeaderBar.vala
jeremypw: Maybe I have misunderstood `constructor property` then.  I use it to mean something initialised in the constructor which is intended to be immutable.

jeremypw:
@danirabbit Conflicts need resolving.

danirabbit:
Closing since headerbar is moving around a lot right now

PR #1225, opened 2023-01-10
Move Terminal to core

### LINKED ISSUE #1082: Make header bar more adaptive, e.g. for small screens and tiling

danirabbit:
Instead of being a plugin, move the Terminal into the main src directory. Also remove the unused bottom notebook which only this plugin was using.

Eventually can be used with #1223 to help fix #1082


data/io.elementary.code.gschema.xml
```diff
@@ -47,6 +47,11 @@
       <summary>Symbol outline visibility</summary>
       <description>Whether or not the symbol outline is visible</description>
     </key>
+    <key name="last-opened-path" type="s">
+      <default>''</default>
+      <summary>Last opened path</summary>
+      <description>Last opened path in the terminal plugin</description>
```
jeremypw, line 53: Terminal is no longer a plugin.  Change to "built in terminal"?

jeremypw (changes requested):
Only one tiny comment inline.  Otherwise looks good as a starting point for a review of the terminal functionality.

Possible refinements for future PRs:
* Easy way of navigating to (and/or opening at) active project's folder
* Separate terminal for each project open
* Built in common (customizable) commands e.g. for compiling, testing, debugging, building Flatpak, git etc

jeremypw:
CI fail not due to these changes.

jeremypw (approved):

Issue #29, opened 2017-05-04
Selected text and highlighted matching brackets are impossible to differentiate.


elementaryBot:
We either need to:
a) Disable highlighting matching brackets when other pieces of text are selected.
b) Have a different style for highlighting matching brackets.

Launchpad Details: [#LP1457744](https://bugs.launchpad.net/bugs/1457744) Lewis Goddard - 2015-05-22 05:50:46 +0000

nathandyer:
I just wanted to add my $0.02 to this issue and say that I think this should be considered a relatively high priority. Today while using it I kept thinking that my keyboard was in insert mode. It lead to me constantly overwriting variables, stomping over close parenthesis and braces, and just generally made for a mildly irritating experience.

joe247:
Yes this is sometimes pretty annoying when you're unsure if you've selected the brackets or not:
![Screenshot from 2020-01-11 12-19-46](https://user-images.githubusercontent.com/38415384/72202565-fe019500-3486-11ea-9c07-3b7206cdbf73.png)
Having a different highlight method... either by different **colour** or _underline_ or [box] will be helpful.

PR #1226, opened 2023-01-12
Styles: misc refinements

danirabbit:
* Mix light background colors between Latte and White, Silver 100 is not quite white so it desaturates in a weird way
* Make selection more contrasty and dark both in light and dark styles
* Fix single-line selection in light style
* Make git and line number gutters the same color


danirabbit:
Adding a note to make sure this doesn't break #29

PR #1227, opened 2023-01-13
FormatBar: use SwitchModelButton in tab popover

danirabbit:
Use Granite's SwitchModelButton so that for example clicking on the label activates the switch. Cleans up code a bit

## BEFORE
Screenshot from 2023-01-12 18 37 46

## AFTER
Screenshot from 2023-01-12 18 33 22


jeremypw (approved):
LGTM. No change in behaviour, code good.

PR #1231, opened 2023-01-13
FormatBar: use menubuttons instead of ToggleButtons

danirabbit:
Also:
* Remove the 6px margin
* Subclass Box instead of Grid
* Use tabular numbers to keep size from shifting around so much when line number changes


jeremypw (changes requested):
Works as expected. Only one very picky comment; the variable names `*_toggle` should be changed e.g. to `*_menubutton` as they are no longer toggles. 

There seems an issue with the labelling of the tab menubutton when tabs are being used rather than spaces.  It changes from e.g. "4 spaces" to "4 tabs" but regardless of the settings the indent width is in fact 4 spaces; but it is achieved in different ways (4 space characters or 1 tab character).  Not sure what the correct label should be though. Maybe "4 (space)" and "4 (tab)".   Anyway, something for a separate PR probably.

jeremypw (approved):
Approving as bugs found were present in master.

PR #1236, opened 2023-01-14
FormatBar: Set sensitivity of affected items, fix binding

### LINKED ISSUE #1232: FormatBar tab button tooltip incorrect sometimes
### What Happened?

Switching between a document from a project where the indenting was specified in the `.editorconfig` file to a document that was not in a controlled project did not cause the tooltip of the tab button in the FormatBar to change accordingly.  It still stated that the indenting was controlled by the `editorconfig` file.

### Steps to Reproduce

1. Open a document controlled by an `.editorconfig` file and one that is not.
2. Switch focus between the two.
3. Observe that the tooltip of the tab button in the FormatBar does not change.

### Expected Behavior

The tooltip should accurately reflect the source of the indent setting being used for the currently focused document.

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #1232
* Fix a binding that was inverted
* Only set insensitive menuitems which are set from editconfig, not the whole menu


jeremypw (changes requested):
This addresses the tooltip and sensitivity issues mentioned.  There is still something odd about the bindings (e.g. changing the tab width via the popover does not actually change the tab width whereas doing via the preference menu does). However best have a PR devoted to bindings.

When the tooltip is null, are the settings definitely from the app settings?  We do not seem to honor Emacs and vim mode line comments in individual documents at present and there does not seem to be another way of having these settings individualized.
So is there any reason why a tooltip stating the source of settings when not `editorconfig`? Otherwise it is a bit confusing to sometimes have a tooltip and sometimes not.

danirabbit:
Yeah this might need a larger rework. It's got multiple things trying to touch these switches which seems bad.

Maybe? But it seems weird to have a tooltip that says something that is expected. I think it really only makes sense to provide an explanation for why user settings are being ignored. Thinking about it, a tooltip is probably not right for this anyways since it would be better to have an explanation next to the insensitive items

danirabbit:
Screenshot from 2023-01-14 11 48 55

jeremypw:
I think it debatable what is expected of the FormatBar, at least by a new user, as it can update depending on the document showing so appears to relate to the document not the app.  The same is not true of the Preference menu, which is independent of the focussed document.  Additional info in the popover to resolve the ambiguity is acceptable though.

jeremypw (approved):
I'll approve this on the expectation that the binding problems mentioned earlier, particular with the indent spin button, will be addressed in a later PR (they were not introduced by this PR)

PR #1237, opened 2023-01-14
Redesign - FormatBar: absorb line width guide settings

danirabbit:
Move this setting to the FormatBar since people may want to adjust this per-file and can be overwritten by editorconfig.

In a future branch, set item sensitivity when overwritten by editorconfig


jeremypw:
Not sure about discoverabiity here - max line width seems unrelated to current line number and goto line number. I agree that as it a possible `.editorconfig` setting it belongs in the FormatBar though.  Adding another button would conflict with attempts to reduce the minimum size so I am not sure what the best solution is.

 Removing it from the preferences menu seems inconsistent to what was done with indent-size and indent-size which are still there.

Marukesu:
> Not sure about discoverabiity here - max line width seems unrelated to current line number and goto line number. I agree that as it a possible `.editorconfig` setting it belongs in the FormatBar though. Adding another button would conflict with attempts to reduce the minimum size so I am not sure what the best solution is.

Maybe together with the tab size and tab style? since they all are about code style.

jeremypw:
@danirabbit What do you think of @Marukesu 's idea?

jeremypw (changes requested):
Uncertainties regarding appropriate location and discovery need addressing (see comments)

jeremypw:
The `.editorconfig` setting `max_line_length` is intended to force "hard line wrapping" at the specified line length according to https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties.  Code does not currently support this.  It does not even support soft line wrapping at a specified width - only at the edge of the widget.  The only thing the "max width" setting currently does is allow drawing of a shaded line width guide. I guess we could (mis)use the editor-config setting to set the line-width guide on a per document basis.  If this can change per document then there is probably no choice but to add another button to the FormatBar to make this visible to the user. If necessary the FormatBar could split like the SearchBar when the window width is insufficient.

The first task is to actually make the line-width guide settable per document and to follow the editorconfig setting in preference to the global app setting (like tab width).

jeremypw:
Converting to draft as there a unresolved questions regarding the design and groundwork needs doing.

jeremypw:
@danirabbit Would it be a good idea to move this to a discussion so the issues raised can be resolved? Maybe revisit after port?

PR #1243, opened 2023-01-14
Application: use built-in GLib functions in Activate

danirabbit:


Marukesu:
Note that this has a little change in multiple windows behaviour, `get_last_window()` always return the last created window, while `active_window` is the last interacted window.

So, when there's two or more windows, the old code use the more recent open, but the new code will use the more recent focused.

jeremypw (approved):
I think the changed behaviour is not unreasonable and anyway only affects a corner case.

PR #1245, opened 2023-01-15
Make Control-Scroll zooming controllable

### LINKED ISSUE #1244: Changing zoom with scroll lags compared with AppMenu and has no feedback
### What Happened?

Code opened with a modest number of tabs; zoom changed using `<Ctrl>Scroll`; the font size did not immediately change and when it did it overshot as there was larger than desired as to the setting while scrolling.  It is difficult to get the desired font size this way with modestly powered equipment and a reasonable number of documents open.  It occurred even with all plugins inactive.

### Steps to Reproduce

1. Open Code on modestly powered equiplement and/or with several documents open.
2. Operation `<Crtl>Scroll` to change the font size
3. The font size does not change for several 100 msec and there is no instantaneous feed back as to the percentage font size, unlike on the AppMenu.
4. Change the font size via the AppMenu;  the response is much faster and there is feedback

### Expected Behavior

The font size should change in a controllable manner with `<Ctrl>Scroll` regardless of equipment speed and number/size of documents open (within reason).

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1244 

Showing the percentage zoom while Control-Scroll zooming will be left for a separate PR as that is a UI change which may not be desired.

### DIFF
--- a/src/Widgets/SourceView.vala
+++ b/src/Widgets/SourceView.vala
@@ -132,10 +132,10 @@ namespace Scratch.Widgets {
                 if (Gdk.ModifierType.CONTROL_MASK in key_event.state) {
                     total_delta += key_event.delta_y;
                     if (total_delta < -SCROLL_THRESHOLD) {
-                        GLib.Application.get_default ().activate_action (MainWindow.ACTION_PREFIX + MainWindow.ACTION_ZOOM_IN, null);
+                        get_action_group (MainWindow.ACTION_GROUP).activate_action (MainWindow.ACTION_ZOOM_IN, null);
                         total_delta = 0.0;
                     } else if (total_delta > SCROLL_THRESHOLD) {
-                        GLib.Application.get_default ().activate_action (MainWindow.ACTION_PREFIX + MainWindow.ACTION_ZOOM_OUT, null);
+                        get_action_group (MainWindow.ACTION_GROUP).activate_action (MainWindow.ACTION_ZOOM_OUT, null);
                         total_delta = 0.0;
                     }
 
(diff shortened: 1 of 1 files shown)


src/Widgets/SourceView.vala
```diff
@@ -126,11 +128,17 @@ namespace Scratch.Widgets {
...
-                if ((Gdk.ModifierType.CONTROL_MASK in key_event.state) && key_event.delta_y < 0) {
-                    ((Scratch.Application) GLib.Application.get_default ()).get_last_window ().action_zoom_in ();
-                    return true;
-                } else if ((Gdk.ModifierType.CONTROL_MASK in key_event.state) && key_event.delta_y > 0) {
-                    ((Scratch.Application) GLib.Application.get_default ()).get_last_window ().action_zoom_out ();
+                var handled = false;
+                if (Gdk.ModifierType.CONTROL_MASK in key_event.state) {
+                    total_delta += key_event.delta_y;
+                    if (total_delta < -SCROLL_THRESHOLD) {
+                        ((Scratch.Application) GLib.Application.get_default ()).get_last_window ().action_zoom_in ();
```
danirabbit, line 135: While we're here, can we
```suggestion
                        GLib.Application.get_default ().activate_action (MainWindow.ACTION_PREFIX + MainWindow.ACTION_ZOOM_IN, null);
```

src/Widgets/SourceView.vala
```diff
@@ -126,11 +128,17 @@ namespace Scratch.Widgets {
...
+                        total_delta = 0.0;
+                    } else if (total_delta > SCROLL_THRESHOLD) {
+                        ((Scratch.Application) GLib.Application.get_default ()).get_last_window ().action_zoom_out ();
```
danirabbit, line 138: And also

```suggestion
                        GLib.Application.get_default ().activate_action (MainWindow.ACTION_PREFIX + MainWindow.ACTION_ZOOM_OUT, null);
```

danirabbit (changes requested):

danirabbit (approved):

PR #1250, opened 2023-01-20
Fix startup warnings

jeremypw:
Fixes startup warnings in terminal

Still a close down warning that I have not been able to track down yet.

One warning required reworking the ChooseProject Button (cannot have both a bound model and a filter func on a listbox).


danirabbit (approved):
Nice job!

PR #1251, opened 2023-01-20
Check for null doc in action_find_global

### LINKED ISSUE #1248: Clicking "Find in Project…" in the welcome screen causes the app crashing
### What Happened?

The "Find in Project…" button is sensitive in the welcome screen although we have no stuff to find there, and clicking it causes Code crashing.

スクリーンショット 2023-01-20 20 30 46

### Steps to Reproduce

1. Open Code
2. Click the menu in the titlebar
3. Click "Find in Project…"
4. See Code crashes

### Expected Behavior

* Code should not crash on clicking "Find in Project…"
* That button should be insensitive in the welcome screen because it's useless

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

```shell
** (io.elementary.code:5923): CRITICAL **: 20:29:04.228: scratch_services_document_get_selected_text: assertion 'self != NULL' failed

(io.elementary.code:5923): GLib-CRITICAL **: 20:29:04.228: g_strsplit: assertion 'string != NULL' failed
Segmentation fault
```

### Hardware Info

_No response_

jeremypw:
Fixes #1248

 - [x] Guard against null current doc
 - [ ] Disable action (insensitive button) when no docs

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -990,6 +990,9 @@ namespace Scratch {
 
         private void action_find_global (SimpleAction action, Variant? param) {
             var current_doc = get_current_document ();
+            if (current_doc == null) {
+                return;
+            }
             var selected_text = current_doc.get_selected_text (false);
             if (selected_text != "") {
                 selected_text = selected_text.split ("\n", 2)[0];


ryonakano:
Thank you for working on this issue!

You might be already noticed, but with this branch we can't perform "Find in Project…" when some project open but no docs open:

VirtualBox_elementary OS 7 Daily_21_01_2023_08_50_46

jeremypw:
@ryonakano I think the same restriction applies in master doesn't it? However, I can try and lift it here if you like.

jeremypw:
@ryonakano Fairly simple to fix that restriction. Marking ready for review now.

danirabbit (approved):

PR #1252, opened 2023-01-22
Ensure folder items in sidebar always expandable

### LINKED ISSUE #739: No expander on folders that contain no text files but expander on empty folders
I have searched open and closed issues for duplicates.

## Describe the bug
In the Folder Manager,  subfolders that contain files but not text files or folders appear without an expander and cannot be expanded.  However, subfolders that contain no files at all have an expander and can be expanded.

### Expected behavior
Expanders should be shown consistently on folders that contain no editable files or folders (to be specified by UX team)

# Platform Information
Elementary Juno unstable.
I'm using the latest version from git that I've manually compiled

jeremypw:
Fixes #739 

Needed to relation to issues with expanding and closing folders in the sidebar as well as accomodating files being added removed and renamed while correctly handling presence/absence of dummy item.  The behaviour of the sidebar is now similar to that of Files in relation to empty folders.

 - [x] Ensure that empty folders AND folders with no visible children have a dummy entry and therefore expandable
 - [x] Simplify handling of created/deleted/renamed files 
 - [x] Only load all children once; thereafter keep tree in sync with changes 
 - [x] Prevent stray output stream files being handled or showing.
 - [x] Put single use functions inline 
 - [x] Prevent terminal warning due to attempted updating git status of non-git folders

Test by creating a project with a mixture of subfolders - empty, only non-text files, text files.  Add, remove, rename files in each case to check correct behaviour.


src/FolderManager/FolderItem.vala
```diff
@@ -186,79 +208,70 @@ namespace Scratch.FolderManager {
...
             }
 
-            base.remove (item);
+            item.badge = "";
         }
 
-        public void remove_all_badges () {
-            foreach (var child in children) {
-                remove_badge (child);
+        private new void add (Granite.Widgets.SourceList.Item item) {
```
zeebok, line 225: What does the `new` keyword do here?

zeebok:
Finally getting to look at this. Code looks good, I just gotta test it.

src/FolderManager/FolderItem.vala
jeremypw: I think this was to ensure overriding the `add` method of the base class (Granite.Widgets.SourceList.ExpandableItem) and to match the pre-existing `private new remove ()` method.  I am not sure `new` is essential though as the base method is public and this is private (?)  Probably clearer just to use a different method name.

jeremypw: On reflection there seems no good reason not to override the base  public add and remove methods for FolderItem.  Having different public and private `add` and `remove` methods seems risky.  I haven't found any problem with overriding the public methods so far.

zeebok:
When I loaded a folder with a mix of file and folders (the source folder for Code), I noticed non-text files in the build/src directory weren't visible and when I removed what was there, the dummy was not created and thus would not let you expand it. If I restored the deleted files it immediately detected and was expandable. 

In my test with fresh new folders, adding and removing things would properly keep the expand functionality. Not sure what is different.

zeebok:
Just looked at it again after realizing I was running it from the build directory instead of installing it. If I Install it works correctly!

zeebok (approved):

jeremypw:
It should not matter whether running installed or not (as long there isnt a pre-existing master version running) as no plugins are involved.  You can verify which version is running by using `meson configure -Ddevelopment=true` before building and then checking the window title in multitasking  view.  Development builds show the branch name in the window title.  I am always getting confused about which version of Code is running when developing Code with Code!

Issue #933, opened 2021-01-30
Add whole word search


mcclurgm:
## Problem

Especially when there's no way to search for specific symbols, it's often useful to search for an exact term when writing code. This is a term that's bounded on either side, and doesn't match if there's more non-space text on either side of the term. For example, say I have a variable called `foo` and a function `do_thing_to_foo`. If I want to find where in my code `foo` is used, I would use a whole word search to find instances of the variable. Without this feature, I would also have to deal with every other place in my code that the word `foo` appears, including declaring and calling the function.

## Proposal

Similar to how we have the case-insensitive option in the search bar, it would be great to have a button for word bounded search.

## Prior Art

Here's an example in VS Code (the button with the bars above and below the Ab):
![Screenshot von 2021-01-30 08 39 46](https://user-images.githubusercontent.com/29022469/106360994-bdb2c600-62d8-11eb-8c4b-982add23e908.png)

simonkrauter:
I miss this option too.
Notes:
In Mousepad it's called "Match whole word".
In gedit it's called "Match entire word only".

jeremypw:
As a workround you can insert spaces in the search term e.g. searching for " foo " will exclude `do_thing_to_foo` (but not `foo ()`). Could also use a regex search.  But yes, a button would be much easier.

jeremypw:
Added "Needs Design" for design team input regarding how this should be exposed.  Space is short in the searchbar as we need to be able to half-tile the window on a minimum recommended resolution screen.

Issue #809, opened 2020-05-04
Sort symbols pane by location


mcclurgm:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature
The symbols pane is very useful, but I find it difficult to use because it sorts the symbols alphabetically. In other code editors (like Builder and Visual Studio), the symbols outline is sorted by where it appears in the code, or there's an option to do so. So functions that appear at the top of the file appear at the top of the outline, and so on. I think of my code spatially, so this is much more natural to me than alphabetically hunting down the function I'm looking for.

Here's what it looks like in Visual Studio Code. I've included the file with functions collapsed, so you can see how the outline maps functions to their locations in code exactly.
![Screenshot from 2020-05-04 10 28 41](https://user-images.githubusercontent.com/29022469/80989519-47893d80-8df2-11ea-8e6e-8c3da97e67ef.png)

I don't expect this to be the only option, so it would probably require a switching interface of some sort. Again, here's what Visual Studio Code does:
![Screenshot from 2020-05-04 10-21-48_modified](https://user-images.githubusercontent.com/29022469/80989975-e150ea80-8df2-11ea-83ba-de0571f3ab64.png)

PR #1253, opened 2023-01-24
Add symbol filter to outline pane

jeremypw:
The is a speculative step towards fixing issues such as #933, #809 and generally increasing the utility of the symbol pane over just searching the text.

 - [x] Add a toolbar header
 - [x] Search Entry in toolbar (currently only search within)
 - - [ ] Option to search start only  (?)
 - [x]  Add Filter menu
 - - [x] Add SymbolType enum and classify Vala and C symbols with symbol type 
 - - [x] Allow filtering by symbol type
 - - [x] Add Action bar to popover for bulk select/deselect filter options
 - - [x] Better icon for filter menu?
 - [ ] Choosing e.g. "method" should automatically choose "class" (or "class" always included in filter?)
 

Note: To show class methods both the "Class" and "Methods" selections must be active.

### No filtering:
Screenshot from 2023-03-21 11 55 57

### Filtering class properties
Screenshot from 2023-03-21 11 56 27

### Searching for text "filter"
Screenshot from 2023-03-21 11 57 24


jeremypw:
### No filtering:
Screenshot from 2023-01-24 16 09 27

### Filtering class properties
Screenshot from 2023-01-24 16 10 11

### Filtering methods and properties containing text "filter"
Screenshot from 2023-01-24 16 30 00

jeremypw:
Looks like the regression is due to recent changes to master :disappointed: not this PR.

jeremypw:
Regression now fixed by #1254.  Leaving this as draft as that PR should be prioritised.

zeebok (approved):
It appears to work as intended!

jeremypw:
@danirabbit Is this something worth progressing? I wasn't sure what icon should be used for the "Select all" button (or whether such a button is needed)

danirabbit:
@jeremypw sorry for taking a while to follow up on this! I pushed a commit with the filter icon we're using in Mail and fixing the sidebar background styles.

I think for the popover I would just a regular menu here and use regular text menu items for the "select all" and "deselect" all items since those are probably more clearly communicated with text and we're not particularly constrained for space in a menu

danirabbit (approved):
Works as expected afaict. Nice job :)

PR #1258, opened 2023-01-28
Show running branch if not master

jeremypw:
Another speculative development aid.

This makes it discoverable which branch of Code is running if not master.  This can be useful when switching between master and the development branch for comparison and there is no other visual cue.  Also for knowing which branch you are currently dogfooding.

For simplicity and compatibility with other apps, the branch is exposed in the window title.

In order for the modified window title to be shown, the project needs to be built with the meson option `-Ddevelopment=true`.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -200,8 +200,7 @@ namespace Scratch {
             );
 
             if (Constants.BRANCH != "") {
-                // Put longer app info on second line to avoid ellipsizing
-                base_title = "\n%s (%s)".printf (Constants.PROJECT_NAME, Constants.BRANCH);
+                base_title = _("Code (%s)").printf (Constants.BRANCH);
             } else {
                 base_title = _("Code");
             }
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -200,7 +200,8 @@ namespace Scratch {
             );
 
             if (Constants.BRANCH != "") {
-                base_title = _("Code") + " (%s)".printf (Constants.BRANCH);
+                // Put longer app info on second line to avoid ellipsizing
+                base_title = "\n%s (%s)".printf (Constants.PROJECT_NAME, Constants.BRANCH);
             } else {
                 base_title = _("Code");
             }
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -23,6 +23,7 @@ namespace Scratch {
         public const int FONT_SIZE_MAX = 72;
         public const int FONT_SIZE_MIN = 7;
         private const uint MAX_SEARCH_TEXT_LENGTH = 255;
+        private static string base_title;
 
         public Scratch.Application app { get; private set; }
         public bool restore_docs { get; construct; }
@@ -148,7 +149,6 @@ namespace Scratch {
         public MainWindow (bool restore_docs) {
             Object (
                 icon_name: Constants.PROJECT_NAME,
-                title: _("Code"),
                 restore_docs: restore_docs
             );
         }
@@ -199,12 +199,14 @@ namespace Scratch {
                 Gdk.Screen.get_default (), provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
             );
 
+            base_title = _("Code") + " (%s)".printf (Constants.BRANCH);
             Hdy.init ();
         }
 
         construct {
             application = ((Gtk.Application)(GLib.Application.get_default ()));
             app = (Scratch.Application)application;
+            title = base_title;
 
             weak Gtk.IconTheme default_theme = Gtk.IconTheme.get_default ();
             default_theme.add_resource_path ("/io/elementary/code");
@@ -340,7 +342,7 @@ namespace Scratch {
 
         private void init_layout () {
             toolbar = new Scratch.HeaderBar ();
-            toolbar.title = title;
+            toolbar.title = base_title;
 
             // SearchBar
             search_bar = new Scratch.Widgets.SearchBar (this);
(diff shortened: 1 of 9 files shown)


src/Widgets/HeaderBar.vala
```diff
@@ -250,6 +250,21 @@ public class Scratch.HeaderBar : Hdy.HeaderBar {
         pack_start (save_button);
         pack_start (save_as_button);
         pack_start (revert_button);
+
+        if (Constants.BRANCH != null &&
+            Constants.BRANCH != "" &&
+            Constants.BRANCH != "master" &&
+            Constants.BRANCH != "main") {
```
danirabbit, line 257: Hrm, Is there no way to ask git which branch is the development focus? For some repos this isn't always a common, predictable name (like for example https://github.com/elementary/seeds)

jeremypw: I'll look into it, but as we would be adding this to each repository separately we should know which branches should be ignored in each case.  Agreed, if that were to change in a repo then an update to this code would also be required.  Probably also need a way to ignore it when git is not available or, in general, when not running in a development environment.

jeremypw:
Converting to draft while fixing CI (absence of git) issue.

Marukesu:
Do the branch name is needed? or only a indicative that we are running a development version? i think we can use the `.devel` css-class for differencing between a development window and a stable one. but that won't say the actual branch this was compiled from.

jeremypw:
> Do the branch name is needed? or only a indicative that we are running a development version? i think we can use the `.devel` css-class for differencing between a development window and a stable one. but that won't say the actual branch this was compiled from.

For the case of installing and testing a branch of an app when you are working on several branches of that app at once it would be helpful to know which branch is actually installed.  However, your solution would be simpler and would not need to be customised for each app.

jeremypw:
@danirabbit There does not seem to be any simple reliable way to extract the development focus using `git` (perhaps someone else can find one).  The only way I could see would be to analyse the contents of `.github/release.yml` which would be tricky with meson.  The simplest way seems to be to hard code it into the configuration and change it on the rare occasion it is required (`release.yml` has to be changed in anycase).

src/Widgets/HeaderBar.vala
jeremypw: See comment below

jeremypw:
@Marukesu Would we need a CssProvider for the `.devel` class? I cannot find this class in the elementary stylesheet.

Marukesu:
@jeremypw for now, i believe. the `.devel` css-class is defined by Adwaita to be appiled to windows to add "development" styles to it.

if this end being a common pattern in elementary apps, they could be implemented in the elementary stylesheet itself as it is in Adwaita.

> For the case of installing and testing a branch of an app when you are working on several branches of that app at once it would be helpful to know which branch is actually installed.

one way to implement the branch name i believe is to put it in the window title/desktop file if built with a `-ddevelopment=true` meson option, instead of checking if it's a "stable" branch build. if both stable and developement version should run in parallel a different application_id would be needed too.

jeremypw:
@Marukesu Thanks for suggestions! I'll try to implement those.

src/MainWindow.vala
```diff
@@ -199,12 +199,20 @@ namespace Scratch {
                 Gdk.Screen.get_default (), provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
             );
 
+            if (Constants.BRANCH != "") {
+                // Put longer app info on second line to avoid ellipsizing
+                base_title = "\n%s (%s)".printf (Constants.PROJECT_NAME, Constants.BRANCH);
```
Marukesu, line 204: shouldn't this be `Code (%s)`?, Constants.PROJECT_NAME is `"io.elementary.terminal` and the `\n` cause the seems to be needed only when we have a document name no?

Marukesu:
one thing i noticed is that the window title don't appears in the dock label. Also, it would be good if we put a label with the title in the header bar when we are in the placeholder view.

src/MainWindow.vala
jeremypw: > if both stable and development version should run in parallel a different application_id would be needed too.

Maybe I misunderstood your comment about application id?   I can change it back to a fixed app name again.  Is there something else I should do to address that comment?

 I'll make the new line conditional on the presence of a current document.

jeremypw:
> one thing i noticed is that the window title don't appears in the dock label. Also, it would be good if we put a label with the title in the header bar when we are in the placeholder view.

That's a good idea - I'll do that.

jeremypw:
@Marukesu  I am not familiar with Plank code. I can see that one way of changing the tooltip would be to generate a different "Name" key in the `.desktop` file for development versions. This would only work for installed versions however, not if running uninstalled from the terminal.  Is there a better way of doing this?

src/MainWindow.vala
Marukesu: it was about the application_id property in the Application class. Something like this:
```vala

construct {
    application_id = Constants.PROJECT_NAME;
    flags: ApplicationFlags.HANDLE_COMMAND_LINE;

    if (Constants.BRANCH != "") {
        application_id += "." + Constants.BRANCH.replace ("/", ".").replace ("-", "_");
    }

    ...
}
```

that would allow us to have different versions open at the same time.

jeremypw: Done

jeremypw:
I can confirm that it is now possible to run a branch instance alongside a master instance.

Marukesu (approved):
Code looks good. One thing that could be done in a future PR is allowing not only parallel execution, but parallel installation too.

> … This, of course, prevents translation but I assume that is not a problem. Translation occurs as normal when not installing a branch with the development option.

I would assume `Code` isn't meant to be translated anyway. so yeah, i don't see a problem either.

jeremypw:
Thanks @Marukesu. I'll merge now and we can refine as required in another PR.  I guess for dual installation we will have to generate a desktop file with a different name? Would a single name such as `io.elementary.code.devel.desktop` suffice? I do not want to clutter the `applications` directory with a different name for each branch.

PR #1260, opened 2023-01-30
Stop infobar continually showing after external changes detected

### LINKED ISSUE #17: "File x was modified..." infobar keeps showing after continuing to edit the file
If a file is changed outside of Scratch and I go to edit it, Scratch shows the infobar informing me to reload the file. However, if I reload the file, it keeps popping up saying the file has been modified.

Launchpad Details: [#LP1261616](https://bugs.launchpad.net/bugs/1261616) Cassidy James Blaede - 2013-12-17 04:16:54 +0000

jeremypw:
Fixes #17 

Avoids getting into a loop if "Continue" is pressed after external changes detected.

It would be expected that after pressing "Continue" the user would either overwrite the external changes or save the document with a different name to avoid the warning re-showing if the tab or window is re-focused.


zeebok:
Does simply moving the code to a delegate function really fix that? I don't even see the infobar tell me the file changed externally when I try to do this so I can't test.

jeremypw:
@zeebok Hmm - its a while since I wrote this and I assumed it was finished. Looking at it again perhaps it isn't.  Sorry for the premature review request!  I'll have another look it later today.

jeremypw:
I think commit 598bbe99757eabb29aab7fbf8116d84beb609642 has changed the behaviour of master so I am not sure the issue (and this PR) is valid anymore.  As you say, the infobar is not showing when externally changes occur (which is a regression I guess).  Code just overwrites any external changes now!

jeremypw:
The infobar only shows under certain conditions:

1. AUTOSAVE is off
2. No unsaved internal changes.
3. Focus out - make and save changes with another text editor
4. Focus back in

Under these conditions the infobar shows and pressing continue does not stop the infobar reappearing as soon as the document is refocused.

I am going to submit another (finished) PR based on the current master that fixes this.

PR #1265, opened 2023-02-03
Make symbol outline aware of follow-system-style setting

### LINKED ISSUE #1263: Symbol outline does not respect system style
### What Happened?

The theme color of the Symbol outline pane doesn't fallow the system dark style preference. Instead, it's background color is whatever the theme color was before toggling on the "Follow system style" toggle.
image

### Steps to Reproduce

1. Open Code while the system is in dark mode
2. Switch it's theme to Light or Elementary light
3. Switch on the Follow system style preference
4. Notice that while the whole app switches it's colors to dark, the symbol outline panel will stay with light background

### Expected Behavior

The background color should change.

### OS Version

7.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1263 

Just copied the corresponding code from DocumentView.

### DIFF
--- a/src/SymbolPane/SymbolOutline.vala
+++ b/src/SymbolPane/SymbolOutline.vala
@@ -45,11 +45,23 @@ public abstract class Scratch.Services.SymbolOutline : Object {
 
         update_source_list_colors ();
         Scratch.settings.changed["style-scheme"].connect (update_source_list_colors);
+        Scratch.settings.changed["follow-system-style"].connect (update_source_list_colors);
     }
 
     protected void update_source_list_colors () {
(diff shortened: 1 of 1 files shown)


Marukesu:
I think this is missing a check for when the system style changes, no?

jeremypw:
@Marukesu Thanks for pointing that out! I will not have time to fix immediately so converting to draft.

jeremypw:
Made use of the `style-changed` signal to avoid repeating some code and ensure style of outline and document match.

Marukesu (approved):
code works as expected.

PR #1276, opened 2023-02-26
Additional Search options in menu

### LINKED ISSUE #933: Add whole word search

### LINKED ISSUE #1175: Search occurence label does not update when active document changed
### What Happened?

After performing a search in a document the search occurence label shows the number of hits.  If another document is focused, the label does not update.

### Steps to Reproduce

1. Open two documents
2. Search for a term found only in one document
3. Switch to the other document
4. The search occurence label (and arrows) do not change.

### Expected Behavior

The label and arrows should show information relevant to the current document.

Either have a per document search entry or perform the current search on each document as it is viewed.  Update the search occurence label and arrows accordingly.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #1275: Case sensitive/insensitive search does not work as expected with mixed case search term
### What Happened?

Case sensitivity works as expected only when the search term is entered all lower case or all upper case.  When entering a mixed case search term and case sensitivity is "off" then search does not find strings matching the search term apart from case.

### Steps to Reproduce

1. Create a document containing the strings (e.g.) "class", "CLASS and "Class"
2. Enter search term "Class"
3. Only one hit is found regardless of whether case sensitivity is on or off.
4. The document however highlights *all* strings that match apart from case regardless of sensitivity setting.
5. 

### Expected Behavior

1. If a mixed case search term is entered then the expected number of hits should be found if case sensitivity is "off".
2. The highlighting should match the hits (maybe a separate issue)

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #933
Fixes #1275 
Fixes #1175 (bitesize)

Puts all current search options (Cyclic, Case Sensitive, Regex) into a menu and adds another (Whole Word).  Also an additional Case Sensitive sub-option is implemented making the current Emacs mode optional.  To this end a combobox with three choices replaces one of the switches.

Screenshot from 2023-02-26 17 26 56

Design input required:
 - Correct Icon for button?
 - Radio Buttons for case sensitive modes?
 - Should all options be backed up into gsettings (currently only cyclic search is) - maybe outside scope of this PR


danirabbit:
@jeremypw can you resolve conflicts here? This is a huge improvement and I'd love to get this merged :)

Also, can you make sure to add lines in metainfo.xml for these fixed issues

jeremypw:
Checking for regressions due to merging master.

jeremypw:
There are some quirks in the searching but they seem to be present in master as well so can be fixed separately.

danirabbit (dismissed):
Touched up some style stuff, this looks awesome. Nice work!

danirabbit (approved):

PR #1283, opened 2023-03-15
Implement option to show files with unknown characters as new document

### LINKED ISSUE #1282: Cannot open files with unknown characters
### What Happened?

Code cannot open files that contain unknown characters.
I've tried to open .bash_history, which contained some of these for some unkown reason (corrupted save?), but it only displays an error message saying the file cannot be read or is corrupted.

### Steps to Reproduce

Try to open this file: [invalid_test.txt](https://github.com/elementary/code/files/10957534/invalid_test.txt)

### Expected Behavior

The file should open normally. Possibly indicate that there are weird characters

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1282

It turns out we are loading corrupt or non-text files into a buffer anyway and I presume `Gtk.SourceFileLoader` can be trusted to produce a valid (if not very readable) text file.  This PR adds an option to the AlertView to show what was loaded when a non-empty string was produced.  Because we do not want the original file to be inadvertently overwritten, the loaded text is used to create a new document.

When no text is loaded we can assume it was because the content of the file was not accessible due to permissions or other problem (an empty readable file is always valid).  The text and presence of the option is therefore modified depending on whether any text was loaded.


zeebok (approved):

jeremypw:
On initial opening of corrupted file:
Screenshot from 2023-03-17 17 41 40

After loading as text:
Screenshot from 2023-03-17 17 49 51

Attempt to load file with no read permission:
Screenshot from 2023-03-17 17 47 19

jeremypw:
@danirabbit You may wish to comment on the text or other design matters.

src/Services/Document.vala
```diff
@@ -733,10 +731,36 @@ namespace Scratch.Services {
         }
 
         // Show an error view which says "Hey, I cannot read that file!"
-        private void show_default_load_error_view () {
-            var title = _("File \"%s\" Cannot Be Read").printf (get_basename ());
-            var description = _("It may be corrupt or you don't have permission to read it.");
+        private void show_default_load_error_view (string invalid_content = "") {
+            var title = _("Cannot read text in file \"%s\"").printf (get_basename ());
```
danirabbit, line 735: Instead of escaping programmatic quotes, please use typographical quotes :)

src/Services/Document.vala
```diff
@@ -733,10 +731,36 @@ namespace Scratch.Services {
...
+            if (invalid_content == "") {
+                description = _("You may not have permission to read the file.");
+            } else {
+                description = _("The file may be corrupt or may not be a text file");
+            }
             var alert_view = new Granite.Widgets.AlertView (title, description, "dialog-error");
+            // Lack of read permission results in empty content string. Do not give option to open
+            // in new document in that case.
+            if (invalid_content != "") {
+                alert_view.show_action (_("Convert to text and show in a new document"));
```
danirabbit, line 746: This is a really long button label. I wonder if it would be acceptable to say something like "Show Anyway"?

danirabbit (commented):
Just a couple small comments!

src/Services/Document.vala
jeremypw: Yes, "Show Anyway" was my first choice but then I was concerned that the user would not realise that we are not actually opening the selected file but creating a new one with processed content.  However, if you think that is not a problem I will revert to the short label.

danirabbit: Yeah let's go with the simpler label and then if we get feedback that it's not clear enough we can always revisit

danirabbit (approved):
Works as expected. Nice fix!

PR #1285, opened 2023-03-16
Set search entry icon and styleclass according to results

jeremypw:
Consolidate setting of search entry icon and styleclass with updating other search widgets


danirabbit:
@jeremypw can you resolve conflicts here please :)

src/Widgets/SearchBar.vala
```diff
@@ -623,6 +608,25 @@ namespace Scratch.Widgets {
...
+                    search_entry.set_icon_tooltip_text (
+                        Gtk.EntryIconPosition.PRIMARY, _("No matches")
+                    );
+                } else if (ctx.has_class (Gtk.STYLE_CLASS_ERROR)) {
+                    ctx.remove_class (Gtk.STYLE_CLASS_ERROR);
+                    search_entry.primary_icon_name = "edit-find-symbolic";
+                    search_entry.primary_icon_sensitive = true;
+                    search_entry.set_icon_tooltip_text (
+                        Gtk.EntryIconPosition.PRIMARY, _("Click to find next match")
+                    );
```
danirabbit, line 627: There's already buttons for next and previous search, and there's no precedent for making the primary icon do this kind of thing. I don't really think this makes sense

src/Widgets/SearchBar.vala
```diff
@@ -623,6 +608,25 @@ namespace Scratch.Widgets {
...
+                // Update appearance of search entry
+                var ctx = search_entry.get_style_context ();
+
+                if (search_entry.text != "" && count_of_search == 0) {
+                    ctx.add_class (Gtk.STYLE_CLASS_ERROR);
+                    search_entry.primary_icon_name = "dialog-error-symbolic";
+                    search_entry.primary_icon_sensitive = false;
+                    search_entry.set_icon_tooltip_text (
+                        Gtk.EntryIconPosition.PRIMARY, _("No matches")
+                    );
```
danirabbit, line 620: This seems like duplicate information since we already show the number of results, including "no results" next to the entry

danirabbit (changes requested):
Couple inline comments

danirabbit (approved):
Looks good 👍

PR #1290, opened 2023-03-26
Stop word selection plugin highlighting conflicting with searchbar

### LINKED ISSUE #807: Highlight Word Selection plugin interferes with case-sensitive Find
If you're doing a case-sensitive find, the Highlight Word Selection plugin still highlights case-insensitive matches. We should probably disable highlighting word selection when performing a find, as it's redundant.

jeremypw:
Fixes #807 

 - Remove word selection plugin highlighting if there are current search matches
 - Make plugin honor case-sensitive setting of searchbar, if searchbar has a search context (to be confirmed).

It is unclear whether the plugin should follow the searchbar settings with respect to case sensitivity or always be case insensitive as at present.  Need design input.


danirabbit:
I can't confirm the original issue. Can you provide steps to reproduce? Is there something I'm missing here?

Screenshot from 2023-04-24 17 10 56

jeremypw:
@danirabbit The problem occurs if you 
 * Set search option to be case sensitive
 * Select a word  (plugin highlights all case-insensitive occurrences )
 * Press `Ctrl-F` to find case-sensitive matches of selected word
 * The case-insensitive matches are still highlighted.

This PR makes the plugin highlighting match the search case-sensitivity to fix this.

zeebok (approved):

jeremypw:
@zeebok Thanks for the review!

@danirabbit There is an open design question here regarding whether or not the word highlighting plugin should still follow the search options *when there is no current search term*.  In this PR it now does.  As this is a change in behaviour design input would be appreciated. 

One drawback of this is the search options are hidden in a popover so it may not be immediately obvious why some words are not highlighted if the user forgets they changed the options to do a search.  On the other hand the alternative would mean the behaviour of the plugin changing depending the presence or absence of a search term which is also non-obvious.

jeremypw:
I'll merge this now, any refinements to behaviour can be done in a separate PR.

PR #1294, opened 2023-03-26
Do not change document search results unexpectedly on focus in

### LINKED ISSUE #1293: Search results change when document focused in
### What Happened?

A case-sensitive search was performed finding 3 results.  When the document was clicked on further (incorrect) matches were highlighted.

### Steps to Reproduce

1. Create a document containing the words "TEST" and "test"
2. Set case sensitive search mode to "ALWAYS"
3. Enter "TEST" into the search entry - 1 result
4. Click on the document
5. Both "TEST" and "test" are highlighted - 2 results
6. 

### Expected Behavior

Switching focus between the search entry and the document should not change the search result.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
FIxes #1293

### DIFF
--- a/src/Widgets/SearchBar.vala
+++ b/src/Widgets/SearchBar.vala
@@ -217,6 +217,11 @@ namespace Scratch.Widgets {
         }
 
         public void set_text_view (Scratch.Widgets.SourceView? text_view) {
+            if (this.text_view == text_view) {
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
This makes sense and I can confirm it fixes the issue

PR #1296, opened 2023-03-27
Simplify settings handling

danirabbit:
We can sync the `active-id` property to the `nick` value since these are both strings


jeremypw (approved):
Much better - thanks!

PR #1300, opened 2023-04-01
Do not highlight neighbouring words of selected space

### LINKED ISSUE #1299: Highlight selection plugin highlights neighbouring words when whitespace selected
### What Happened?

Selected some whitespace prior to deleting or overwriting it.  An amount of text either side of the selection was highlighted unexpectedly.

### Steps to Reproduce

See above

### Expected Behavior

When the selection is not a word  then the plugin should not highlight anything.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1299 

The selection extension algorithm is reworked so that leading and trailing space is ignored and the search context is not extended to neighbouring words


danirabbit:
I think I can confirm the original issue was fixed, but there's a couple new issues here that aren't in master:

Now selecting whitespace highlights all instances of whitespace. It seems like we should never highlight white space right?

If you highlight some whitespace, whatever comes after that whitespace will count as a selection:
Screenshot from 2023-04-24 17 03 15

jeremypw:
@danirabbit Hopefully the algorithm now works as expected.  It is not entirely clear how a partial selection should be auto-extended by this plugin. Perhaps it should be documented in the code if it can be decided.  This PR now autoextends the selection to word boundaries as determined by a couple of UnicodeBreakTypes but will not extend to include e.g. enclosing brackets.

Nothing happens now if you just select space.

Leading and trailing space is ignored even if manually selected (is this correct?)

It is difficult to test this with all possible characters though.

danirabbit (approved):
This does seem to solve the originally reported issue and works better than current master. I think extending a partial selection at all is probably a bad idea and will end up causing issues. It seems like trying to be too clever instead of just doing exactly what was input

PR #1301, opened 2023-04-02
Bump runtime to 7.2

danirabbit:


colinkiama:
I'm not sure if this is still being worked on however adding libsoup-2.4 to the flatpak manifest appears to fix the flatpak build errors here:

```yml
  - name: libsoup-2.4
    buildsystem: meson
    sources:
      - type: git
        url: https://gitlab.gnome.org/GNOME/libsoup.git
        tag: '2.74.3'
    config-opts:
      - '-Dvapi=enabled'
  - name: code
    buildsystem: meson
    config-opts:
     - '-Dhave_pkexec=false'
    sources:
      - type: dir
        path: .
```

jeremypw:
The Flatpak version of Code needs a lot of work as I recall to make it fully functional.  Several things do not currently work in a sandbox.

Marukesu:
IIRC, the issues with the flatpak version compared with the host one was the empty "Open In" menu, not having the real path to files to disambiguate the labels, and a libgit2-glib code/vapi missmatch.

for the "Open In" menu, a solution using the OpenURI portal was rejected in favor of using a future "Share" portal (see: https://github.com/elementary/code/pull/1162#issuecomment-1126774276).

for the libgit2-glib issue, we could fix it with a patch in the manifest, or in code itself (see: #1091).

for the file paths, there's nothing we can do really as it's a limitation from using portals.

So, i would say it pretty complete already.

jeremypw:
There is also possible problems around development workflows  like mine where you are using Flatpak Code and non-Flatpak Terminal (and occasionally non-Flatpak Files). These could go away if and when Code becomes a completely self contained IDE.  I'll try to give Flatpak some more attention.

jeremypw:
The tab disambiguation problem could maybe be fixed by using the project name as the root of the path.

jeremypw:
@danirabbit I think this can be closed now as the runtime has been bumped to 8 now.

PR #1308, opened 2023-04-11
Ask save location with dialog

jeremypw:
Speculative PR investigating possibility of replacing the infobar (with limited layout possibilities) with a dialog when it comes to dealing with problems saving to the current location.

Because of the focus changes caused by the dialog (which trigger handlers in document) several other significant changes were needed, principally giving the document a "locked" state in which it cannot be saved or autosaved and does not react to focus changes nor is saved when closed.  The document enters this state (shown as a lock in the tab) whenever it cannot be saved to the current location and before the dialog is shown.  The document remains in that state until "save as" is performed.

The ask save location dialog is called from fewer places than previously - now only after a save failure and after a status change to "not exist" or "unwritable". 

External edits to the document still show an infobar - that can be revisited in another PR.

As now, the document is always opened unlocked and and does not become locked until edited (or unmounted or deleted).  This allows viewing read-only documents without unnecessary interruption.

It needs to be decided if the more intrusive dialog is a better option.  It can then be smartened up.


danirabbit (changes requested):
Left a few comments. This seems to work well, nice job!

src/Services/Document.vala
```diff
@@ -794,114 +846,120 @@ namespace Scratch.Services {
...
+            string error_text = ""
+        ) {
+            locked = true;
+            string primary_text;
+            if (source_view.buffer.get_modified ()) {
+                primary_text = _("The document cannot be saved");
             } else {
-                Utils.action_from_group (MainWindow.ACTION_SAVE, actions).set_enabled (true);
-                this.source_view.editable = true;
+                primary_text = _("The changes to the document cannot be saved");
```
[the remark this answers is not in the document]
jeremypw: It was so I could distinguish which part of the code was causing the dialog to show.  You are probably right that from a user point of view there is no difference. I'll remove the distinction before merging.

src/Dialogs/AskSaveLocationDialog.vala
```diff
@@ -0,0 +1,28 @@
...
+
+        if (error_message != "") {
+            show_error_details (error_message);
+        }
+    }
+
+    construct {
+        var app_instance = (Gtk.Application) GLib.Application.get_default ();
+        transient_for = app_instance.active_window;
+        image_icon = new ThemedIcon ("dialog-warning");
```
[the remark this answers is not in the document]
jeremypw: Done

jeremypw:
@danirabbit Thanks for the review comments! I have actioned most of them.  There are couple that require further discussion.

src/Services/Document.vala
```diff
@@ -637,13 +687,15 @@ namespace Scratch.Services {
         public string get_tab_tooltip () {
             if (is_file_temporary) {
                 return _("New Document"); //No path for a new document
+            } else if (locked) {
+                return _("Cannot save this document to %s").printf (Scratch.Utils.replace_home_with_tilde (file.get_path ()));
```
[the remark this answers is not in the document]
danirabbit: Naw DynamicNotebook is gone in Gtk4 anyways so wouldn't make sense to open a new issue

src/Services/Document.vala
```diff
@@ -794,114 +846,113 @@ namespace Scratch.Services {
...
+                                if (save_as_with_hold.end (res)) {
+                                    locked = false;
+                                }
+                            });
+                            break;
+                        case Gtk.ResponseType.REJECT:
+                            break;
+                        default:
+                            assert_not_reached ();
+                    }
```
danirabbit, line 949: This seems excessive and makes Code crash if you close the window with the escape key. Probably can just be:
```suggestion
                    if (id == Gtk.ResponseType.ACCEPT) {
                        save_as_with_hold.begin ((obj, res) => {
                            if (save_as_with_hold.end (res)) {
                                locked = false;
                            }
                        });
                    }
```

src/Services/Document.vala
```diff
@@ -794,114 +846,113 @@ namespace Scratch.Services {
...
-            if (info_bar.visible == true) {
-                return;
-            }
+        private void ask_save_location (
+            string details,
+            string error_text = ""
+        ) {
+            locked = true;
+            var dialog = new Scratch.Dialogs.AskSaveLocationDialog (
+                _("“%s” can't be saved here. Save a duplicate somewhere else?").printf (file.get_path ()),
```
danirabbit, line 929: Seems like this should be

```suggestion
                _("“%s” can't be saved here. Save a duplicate somewhere else?").printf (file.get_basename ()),
```

src/Dialogs/AskSaveLocationDialog.vala
```diff
@@ -0,0 +1,28 @@
+/* Copyright 2023 elementary, Inc. <https://elementary.io>
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+public class Scratch.Dialogs.AskSaveLocationDialog : Granite.MessageDialog {
```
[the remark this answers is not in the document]
danirabbit: That makes sense, but `ask_save_location ()` is already an abstraction for this dialog so to add another layer of abstraction here seems like we're creating lasagna and we should have one or the other

src/Services/Document.vala
jeremypw: Yes - `assert_not_reached ()` should not really be used in released code. I put it in so any unhandled code paths are obvious when testing - and you have found one.  Will fix.

src/Services/Document.vala
jeremypw: I think there was a user complaint that the basename was not sufficient - will check back.

src/Dialogs/AskSaveLocationDialog.vala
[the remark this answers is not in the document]
jeremypw: OK, will merge these one way or another.

src/Services/Document.vala
jeremypw: I was thinking of https://github.com/elementary/files/issues/2164 when I made this change.  Not exactly the same issue but related I think.

jeremypw: I removed the "Save duplicate here?" text to re-shorten the prompt as it seems to repeat the text on the button.  Alternatively it could be put on a separate line?

jeremypw: Actually, the full path is given in the secondary text anyway so you are right the primary should just show the basename.

jeremypw:
I have now inlined the new dialog and made the other requested changes.  Also fixed some new critical warnings on startup.

danirabbit (approved):
Looks good and works as far as I can tell!

src/Services/Document.vala
```diff
@@ -794,114 +852,127 @@ namespace Scratch.Services {
...
-                        critical (e.message);
-                        show_default_load_error_view ();
-                        return;
-                    }
+                ask_save_location (details.printf ("<b>%s</b>".printf (get_basename ())));
+            } else if (loaded) { // Check external changes after loading
+                if (!locked && !can_write () && source_view.buffer.get_modified ()) {
+                // The file has become unwritable while changes are pending
+                    locked = true;
+                    var details = _("File “%s” was does not have write permission.");
```
TomiOhl, line 871: Bit late to the party but this sentence should probably be rephrased

jeremypw: Oops! Thanks for noticing.

PR #1309, opened 2023-04-16
Use dialog for external changes

### LINKED ISSUE #965: Undo-ing sometimes deletes the entire contents of a file
## What Happened

I was just editing a HTML document, undid something, realised I could no longer open the file in my browser from the sidebar, reopened the app to see if it would fix it and turns out the text is all gone. 
I didn't have any recent backups, which is my fault - but this is pretty bad behaviour if it's constantly autosaving.
I've also realised that undoing every so often causes all the text in a file to disappear, but unoing again or redoing seems to fix that. 

## Expected Behavior

Undoing or redoing should not randomly delete all your work. There should also be copies saved periodically since it's autosaving everything.

## Steps to Reproduce

1. Edit a text document and repeatedly undo/redo
2. Keep making changes until erronious behaviour

## Platform Information

elementary OS 6 Daily

### LINKED ISSUE #1327: Git branch change can result in unwanted warnings
### What Happened?

A change of git branch resulted in one of the open files disappearing resulting in a dialog with the options "Ignore" or "Save Duplicate elsewhere".  While these options are appropriate for files deleted externally in this case the most likely desired action is to close the tab but this option is not given.

### Steps to Reproduce

* Open Code master branch and open e.g. "Document.vala"
* Switch to the `deb-packaging` branch
* A dialog appears warning that "Document.vala" has been deleted
* To close the tab you have to press "Ignore" and then close the tab manually

### Expected Behavior

Possible solutions:

1. Add an option to close the tab to the dialog.  The shortcoming of this is that if there are many files open that disappear you have to close each one separately
2. When the git branch is to be changed, all open documents are *first* saved and then closed.  Any documents that exist in the new branch could then be re-opened although there is no guarantee that the user wants to work on the same documents.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #965
Fixes #1327 

 Following #1308, use dialog instead of infobar when there are external changes

 * Add external change dialog inline
* Connect focus in signal at same time as focus out signal
 * Check file status before autosaving
 * Set buffer modified state to false after save and after loading external changes
 * Remove dependence on autosave setting
 * Dialog text reflects whether there are unsaved internal changes
 * Fix issue with new documents by opening them the same way as existing documents
 * Do not show dialog when no unsaved changes e.g. when git branch changed
 * Loading external changes is undoable
 * Lose now unused infobar

The next Code release is waiting for this to be merged.


jeremypw:
An issue that has shown up dogfooding this branch is that if you change git branch with affected files open then this is treated as an external change and unwanted dialogs appear.  Need to ignore the changes made by git branch change.

jeremypw:
@danirabbit In master, with autosave "on" (which is default) external changes are immediately loaded without a dialog so the git branch change issue does not normally arise.  I am not sure if this behaviour is correct but I guess it is based on the assumption that the document is being edited by only one user and conflicting edits are virtually impossible.  Shall I continue the master behaviour here?

jeremypw:
I have gone ahead and reproduced the behaviour in master (without referencing the autosave setting). Not sure if the immediately loaded external changes need to be flagged up with a toast or something but that can be left for a later PR.

jeremypw:
There is a small improvement wrt master in that the behaviour is the same regardless of the autosave setting - even if autosave is "off" external changes are loaded immediately if there are no unsaved internal edit and in this case the document is *not* marked as "unsaved".

src/Services/Document.vala
```diff
@@ -92,6 +92,7 @@ namespace Scratch.Services {
 
         // Locked documents can be edited but cannot be (auto)saved to the current file.
         // Locked documents can be saved to a different file (when they will be unlocked)
+        // Create as locked so focus events ingb=nored. Unlock when content is loaded
```
zeebok, line 95:
```suggestion
        // Create as locked so focus events ignored. Unlock when content is loaded
```

zeebok (commented):
Code looks good. Gonna give a test over the next day or two for you.

zeebok:
Trying to test by editing the same file in nano is causing two things:

1. I don't see the dialog appear, it simply refreshes the doc with the new version. If autosave is off it still refreshes but the document tab has a * indicating that it isn't saved.
2. If I get to the scenario of the doc refreshing from an external change and perform an undo, the whole document clears. If I undo again the previous version comes back as would be expected.

jeremypw:
@zeebok Thanks very much for testing this out.  Autoloading of external changes is mention in the comments above and is motivated by the need to change git branches without triggering the dialog unnecessarily.  The changes should only autoload if there are no unsaved changes to the document.

> There is a small improvement wrt master in that the behaviour is the same regardless of the autosave setting - even if autosave is "off" external changes are loaded immediately if there are no unsaved internal edit and in this case the document is not marked as "unsaved".

There is an open question as to whether there needs to be a toast or something indicating the changes have been loaded. Since the changes were presumably made by the same user not sure this is needed.

Undoing shouldn't clear the document though - I'll look into that.

You should get the dialog if there are conflicting changes made internally and externally which in effect means "autosave" is off.

I used `gedit` when testing as that has its own conflict warning system.

jeremypw:
Regarding the `undo` issue: It seems that setting the buffer text programmatically records to actions in the buffer undo manager - "delete old text" and "insert new text", hence it takes two `<Ctrl>Z` to get back to the old text.  Not sure if there is a way round this.

jeremypw:
I have achieved single-click undo of autoload by using copy/paste rather than just setting the buffer text. This is at the cost of more code and dealing with the asynchronous nature of pasting.  Its not quite perfect because after undo the old text remains selected but selection is not recorded by the undo manager so cannot be undone.  Questionable whether this is worth the extra complexity as this would be likely to be used very rarely in practice.

zeebok:
Would a better path be to clear the undo history so trying to undo doesn't do anything?

zeebok:
Okay now that I know that I can confirm the detection and dialog work as describe!

jeremypw:
It would be simpler to disable undoing changes made by autoloading external changes, if that is acceptable.

zeebok:
I think it is but I am just a single user

jeremypw:
@zeebok I agree with you. Most users would probably rarely, if ever, trigger this code anyway.

zeebok (approved):

PR #1318, opened 2023-05-12
Fix creating duplicates of unwritable files

### LINKED ISSUE #1316: Fails to create duplicate if unwritable
### What Happened?

If a file's write permission is changed during editing then the "Save Duplicate" option of the resulting dialog fails to create a new file in the selected location even though the tab is renamed.

### Steps to Reproduce

 * Create and open a writable text file
 * Edit it with the "Save changes" option on
 * Remove the write permission of the file
 * Edit some more
 * In the resulting dialiog press "Save Duplicate"
 * Select the same (writable) folder but a different file name
 * No file is created and the tab remains locked but renamed

### Expected Behavior

A *writable* duplicate should be created in the requested location and the tab lock removed.

### OS Version

7.x (Horus)

### Software Version

Compiled from master

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1316

The issue was caused by the `focus-in-event` handler checking the file status before the duplicate has been created resulting in a further error dialog. This was fixed by putting the document in a `working` state while duplicate creation is in progress and ignoring focus in events in that state.  The file status is checked after save-as completes instead.

### DIFF
--- a/src/Services/Document.vala
+++ b/src/Services/Document.vala
@@ -372,7 +372,7 @@ namespace Scratch.Services {
 
             // Focus in event for SourceView
             this.source_view.focus_in_event.connect (() => {
-                if (!locked) {
+                if (!working && !locked) {
                     check_file_status ();
                     check_undoable_actions ();
                 }
@@ -503,12 +503,15 @@ namespace Scratch.Services {
             var old_uri = file.get_uri ();
             var old_locked = locked;
             locked = false;  // Can always try to save as a different file
+            working = true; // Prevent premature status check when focus in after dialog closes
             var result = yield save_with_hold (true, true);
             if (!result) {
                 file = File.new_for_uri (old_uri);
                 locked = old_locked;
             }
 
+            working = false;
+            check_file_status ();
             return result;
         }
 


src/Services/Document.vala
```diff
@@ -503,12 +503,15 @@ namespace Scratch.Services {
             var old_uri = file.get_uri ();
             var old_locked = locked;
             locked = false;  // Can always try to save as a different file
+            working = true; // Prevent premature status check when focus in after dialog closes
             var result = yield save_with_hold (true, true);
```
zeebok, line 507: Just to be sure since it seems `working` is not a new variable, is there any kind of situation that this call or altering the value for the variable here might cause undesired behavior?

zeebok (approved):
Looks good. Just the one question but I feel like I am being paranoid :sweat_smile:

src/Services/Document.vala
jeremypw: `working` is a property of Granite.Widgets.Tab and causes the spinner to show.  It is only used internally to Document.vala to stop certain things happening when the document is a transitional state, originally when loading but extended here to include being saved into new file.  I think its use here is correct.

PR #1320, opened 2023-05-21
Ensure correct sidebar item is focused, or none, when tab removed

### LINKED ISSUE #1319: Document does not open when clicked if already focused
### What Happened?

Occasionally clicking on an item in the sidebar has no effect - the document does not open.  This happens if the item is already focused.

### Steps to Reproduce

* Open a project but no documents
* Open one document with the sidebar
* Close that document with the tab (focus remains in sidebar)
* Click on that sidebar item again
* The document does not reopen

### Expected Behavior

The document always opens when the sidebar item is clicked on (if it is not already open) regardless of what is focussed.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1319 

This fixes the issue by ensuring that a document that is not open in the document viewer is not selected in the sidebar so that it always activates and opens when clicked. 

It also ensures the sidebar selects the current document when a tab is closed.


zeebok (approved):
Nice fix! I keep forgetting this was thing and am glad to see it fixed

Issue #1322, opened 2023-05-31
Man page contains unwanted option


jeremypw:
### What Happened?

The man page for Code was seen to contain the option `-s, --set=plugin           Set of plug-ins` but the code does nothing with this option (although still treats it as valid).

### Steps to Reproduce

1. Open terminal 
2. Run `man io.elementary.code`
3. Compare with code in `Application.vala`

### Expected Behavior

The man page should only show options that work.

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

danirabbit:
Probably not 😅

Marukesu:
actually, the option is passed by the MainWindow class to the PluginManager, so it's used. it's a folder name that contains extra plugins (see the creation code in PluginManager.vala). if there's already a running instance, the set defined in that instance is the one used.

i don't think it makes sense in keeping it, even with a plugin system. this option could only make sense if every window was a process by itself.

jeremypw:
@Marukesu Yes, the issue was badly worded. The use of custom plugin sets in a subfolder of `plugins` is unused and not officially supported by elementary even though the code does support it.   I agree that there is no reason not to remove this feature and simplify the code.

jeremypw:
CLosing as fixed

PR #1324, opened 2023-06-01
Remove plugin set option

jeremypw:
Towards addressing #1322 

A separate PR is required to remove the option from the man page,

### DIFF
--- a/data/io.elementary.code.1
+++ b/data/io.elementary.code.1
@@ -33,11 +33,6 @@ Open a new window\.
 Show a help message and exit\.
 .
 .TP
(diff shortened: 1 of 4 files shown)


danirabbit:
@jeremypw does this still require a separate PR since #1325 was merged?

jeremypw:
#1325 Did not alter the man page only changed the way it is installed so yes, after code support for this option is removed by this PR, the manpage needs updating.  I could do it here as well though.

jeremypw:
Oh, I cannot amend the man page here - it targets the deb-packaging branch.

Marukesu (approved):
LGTM.

Issue #408, opened 2018-04-28
Folders are selectable in the sidebar


jwinnie:
It doesn't make sense that folders are selectable:
![folderclick](https://user-images.githubusercontent.com/20506149/39391105-45b4781a-4a53-11e8-8dba-f1ed9b9b6583.gif)
Folders should expand/collapse when clicked.

berliner:
Depends, for things like https://github.com/elementary/code/issues/418 it might actually make sense to be able to select folders by clicking on them. In TextMate 2 (mac), that's a way of selecting the search scope and it's working really well.

marbetschar:
I agree. Stumbling a lot across folders which do not automatically open in Code which is quite annoying. Event though it *might* make sense to select folders for search, I'd argue that use case happens way less often than expanding/collapsing folders.

That said, the default behavior in my opinion should be expanding/collapsing. If selecting makes sense, then this could be done using Ctrl + click. That's also the way it works in the Files app.

jeremypw:
This also happens in granite-demo so seems to be a "feature" of the underlying widget used (Granite.Widgets.SourceList).  Maybe able to override, but if not then will need to move issue to Granite project.

PR #1330, opened 2023-06-07
Make folders renameable

### LINKED ISSUE #1328: Cannot rename folders from Code
### What Happened?

I cannot rename any folders from within the Code application which is the default text editor in Elementary OS. Files can easily be renamed from within the application.

### Steps to Reproduce

1. Open a folder inside the Code (default editor) text editor
2. Try to rename the folder
3. Nothing will happen.

### Expected Behavior

The folder should be easily be able to be renamed and then the new name can be saved.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

Processor: Intel Core i5-4200m
Graphics: Intel HD 4600
RAM: 16GB
SSD: 256GB

jeremypw:
Fixes #1328 

This is the simple solution.  There does not seem to be any particular downside to making folders selectable but reviewers should test that.

### DIFF
--- a/src/FolderManager/FolderItem.vala
+++ b/src/FolderManager/FolderItem.vala
@@ -38,8 +38,6 @@ namespace Scratch.FolderManager {
         }
 
         construct {
-            selectable = false;
-
             dummy = new Granite.Widgets.SourceList.Item ("");
             // Must add dummy on unexpanded folders else expander will not show
             ((Granite.Widgets.SourceList.ExpandableItem)this).add (dummy);


danirabbit:
Looks like this was done in #763 to fix #408

jeremypw:
Hmm,  folders were made unselectable so they expand when clicked over their whole width; when they are selectable you have to click the expander (like Files List view).  Which is better?  Is renamability more important?  Maybe I can make them temporarily selectable when the rename option is selected.

jeremypw:
I succeeded with making the folder item temporarily selectable in order to edit it but there is a problem in reverting to unselectable when the edit is cancelled with the `Escape` key.  In this case there is no signal emitted to connect that I can find (when an edit is performed then an `edited` signal is emitted and I can connect to that to revert the selectable state).
Have to look at Granite code to see what can be done.

jeremypw:
I cannot see a work around for this without changing Granite so that there is a signal when editing ends regardless of whether it was successful or cancelled.  The sourcelist widget seems not exist in Granite 7 anyway?

PR #1347, opened 2023-07-07
Sync tab settings

### LINKED ISSUE #1340: Tab width menu at the top doesn´t work
### What Happened?

When trying to change the tab width using the quick menu at the top of the window, it won't change the tab width. The +- menu will change, but the label won't and the actual tab width doesn't change either. It can only be changed in the preferences menu.

### Steps to Reproduce

1. Fresh install on Elementary OS 7 Horus
2. Open Code.
3. Default tab width will be 4 spaces. You cannot change this using the quick menu at the top.

### Expected Behavior

Changing the tab width according to the new setting.

### OS Version

7.x (Horus)

### Software Version

### Log Output

```shell
n/a
```

### Hardware Info

Lenovo C340 14-IWL
i3-8145u
8GB Ram
128GB SSD

Elementary OS freshly installed using mutschler.dev's install guide for eOS 6.1

jeremypw:
Fixes #1340 

* Ensure the settings, appmenu preferences, format bar and tab popover stay in sync and are applied to doc when editorconfig not overriding
* Allow overriding tab width and tab style separately
* Ensure correct settings displayed and used when switching documents while editorconfig active.
* Ensure correct symbol used when drawing spaces when switching between spaces and tabs

There is a pre-existing issue that the range of the tab width spinbutton is up to 24 while the maximum tab width set by editorconfig plugin is 16.  This should be resolved here or later.

As this PR changes a plugin, it needs to be installed for testing.


src/Widgets/FormatBar.vala
```diff
@@ -180,27 +184,42 @@ public class Code.FormatBar : Gtk.Box {
...
 
-        bind_property ("tab-set-by-editor-config", editorconfig_infobar, "revealed", BindingFlags.SYNC_CREATE);
-        bind_property ("tab-set-by-editor-config", space_tab_modelbutton, "sensitive", BindingFlags.INVERT_BOOLEAN | BindingFlags.SYNC_CREATE);
-        bind_property ("tab-set-by-editor-config", tab_box, "sensitive", BindingFlags.INVERT_BOOLEAN | BindingFlags.SYNC_CREATE);
+        format_tab_header_from_global_settings ();
+        width_spinbutton.value_changed.connect (() => {
+            if (!tab_width_set_by_editor_config) {
+                Scratch.settings.set_int (
+                    "indent-width",
+                    ((int)width_spinbutton.@value).clamp (2, 16)
```
zeebok, line 196: If we are clamping to 16 both in editorconfig and here, should we just remove this clamp and limit the spinbutton?

zeebok (approved):
Code looks good and works as advertised. Besides my one question I think this is good to merge.

src/Widgets/FormatBar.vala
jeremypw: I don't see why not. Do you agree that `16` is a reasonable maximum for the tab width?

jeremypw:
@zeebok I have removed the clamp as you suggested and added a corresponding forced range to the schema.  Also changed the range in the plugin to correspond.

jeremypw:
I notice that `gedit` only allows tabs in the range 2 - 8 and also only even numbers.

zeebok (approved):
> I notice that `gedit` only allows tabs in the range 2 - 8 and also only even numbers.

Even seems nice but I don't feel strongly enough to not approve this. I think this is great and good customization.

PR #1348, opened 2023-07-07
Draw spaces on current line option

### LINKED ISSUE #570: Draw Spaces On Current Line
Heyo! Feature request for Code in regards to the drawing spaces function. Currently we have it off, for selected text, or all the time. Would it be possible to do this for the current line specifically? Would be useful when jumping about through a document, but I'm not sure how difficult this might be to add or if it's really useful for anyone else. Thanks in advance either way, Code is fantastic.

jeremypw:
Fixes #570 

Adds a third draw-space mode as requested.  In this mode spaces are always drawn on the current line.  They are also drawn on any selections.  To allow the user to choose between three states, a switch is replaces by a combobox.

Screenshot from 2024-02-11 18 32 33


zeebok:
I am noticing that if I have whitespace at the end of a line, it will still draw the spaces when on Current and on a different line:
image

jeremypw:
@zeebok Looks like the same thing happens in `master`.  If it can be fixed here without significantly increasing the diff I will.

jeremypw:
@zeebok:  Looks like this is deliberate behaviour (when draw spaces is ON at all) - see #433

zeebok (approved):

PR #1354, opened 2023-07-20
Fix spurious external change warnings

jeremypw:
Since recent changes to external change handling it was noticed that spurious warnings of external changes were produced while editing.  This was traced to file status checking not completing before autosaving commenced because source loading is asynchronous.

This PR makes sure file status checking completes before saving starts when autosaving.

### DIFF
--- a/src/Services/Document.vala
+++ b/src/Services/Document.vala
@@ -209,8 +209,8 @@ namespace Scratch.Services {
             // Check if file changed externally or permissions changed
             this.source_view.focus_in_event.connect (() => {
                 if (!locked && !is_file_temporary) {
-                    check_file_status ();
                     check_undoable_actions ();
+                    check_file_status.begin ();
                 }
 
                 return false;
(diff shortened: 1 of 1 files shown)


jeremypw:
As this issue depends on semi-random events order and timing, it was tested by using it for intensive coding for a while without spurious warnings.

jeremypw:
Converting to draft as started getting unexplained closing of Code. Investigating.

jeremypw:
OK, its to do with a certain file causing problems with the Symbol Pane - nothing to do with this PR.

jeremypw:
It could be something to do with this although it only seems to affect a certain file - need to investigate effect of changes in timing of autosave on the outline plugin. Seems to sometimes only partially parse the file - or not at all.

jeremypw:
Now confident that the outline issue is unrelated to the PR.

zeebok (approved):
Tried to repro for like 5-10 minutes but it wasn't happening. The code seems to make sense though. If you really want me to try and see if it works, I can keep testing but for now I approve.

jeremypw:
@zeebok  Thanks for testing. Its one of those annoying timing issues that is not always reproducible. I'd guess it might be more likely to happen with a large document that takes a longer time to status check.

jeremypw:
I think I'll merge this but wait a while for further dog-fooding before finalizing the release.

PR #1356, opened 2023-07-20
Fix crash in vala outline when entering some invalid code

### LINKED ISSUE #1355: Vala scanner goes into loop and may crash with certain code
### What Happened?

Inserted some code in the form:
```
try {
---
---
} catch () {
---
---
}
```

into a Vala file.  On opening the symbol outline it remained blank and inspecting the terminal output indicated it had gone into a loop.   Repeated Terminal warning: `vala_scanner_read_token: assertion 'self != NULL' failed`

Editing the file with "autosave" on resulted in random crashes which back traced to the outline plugin.

### Steps to Reproduce

1. Edit some code containing lines in the form mentioned above with the outline plugin on

### Expected Behavior

The outline plugin should handle invalid Vala code without crashing or going into a loop.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1355 

The root of the issue is in the libvala library - the parser never finishes on some invalid code lines and there is no obvious way of cancelling it.

To fix the crashes this PR ensures that if one parsing thread stalls and never finishes, a new one is not started.

A limitation of this is that if the code is edited to fix the problematic line, the outline is not re-parsed until the outline pane is hidden and then shown again (or the document closed and reopened - this PR ensures the outline is destroyed on closing the document).

It is questionable whether this issue is likely to occur often enough warrant implementing a more complex fix.

Any advice on how to force the thread to end when the parser has stalled would be welcome.


jeremypw:
Looks like the original issue was fixed upstream so closing.

PR #1374, opened 2023-09-22
Improve show_replace action

### LINKED ISSUE #1371: Highlighted text isn't used with Ctrl-R shortcut
### What Happened?

If you use Ctrl-R to open the replace dialog with a word highlighted, it does not use the highlighted text as the search keyword.

### Steps to Reproduce

1. Highlight some text
2. Press Ctrl-R to open replace search
3. Note that the "Find" field does not update to the highlighted text

### Expected Behavior

I would expect it to behave like the Find action and use the highlighted text for the keyword.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1371 

 - [x] Use selected text as search term 
 - [x] Focus replace entry


zeebok (commented):
This looks great. I am going to test this out for a bit on my main machine and let you know!

zeebok:
So far this is exactly what I would expect and I am really loving it!

zeebok (approved):
i am inclined to approve this. The behavior fits intuitively for me but i am curious what your thoughts are on this.

PR #1379, opened 2023-11-21
Fix-folder-search-filtering-1198

### LINKED ISSUE #1198: Searching within a folder should ignore global filters
### What Happened?

When performing a global search Code only searches in `.vala` files and ignores untracked files.

The option menu for folders includes "Find in Folder" - if that option is chosen Code still applies those filters even if it means ignoring the folder clicked on

### Steps to Reproduce

1. Right click on an untracked folder e.g. `build`
2. Choose "Find in folder"
3. Search for a string known to be in the clicked folder
4. The string is not found (the folder is not searched)

### Expected Behavior

If a user explicitly wants to search within a certain folder that should be honored (or a warning given).

### OS Version

7.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

colinkiama:
Fixes #1198

### DIFF
--- a/src/FolderManager/FileView.vala
+++ b/src/FolderManager/FileView.vala
@@ -197,7 +197,8 @@ public class Scratch.FolderManager.FileView : Granite.Widgets.SourceList, Code.P
                     ? item_for_path.file.file
                     : search_root.file.file;
 
-                search_root.global_search (start_folder, term);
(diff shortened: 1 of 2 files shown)


jeremypw (approved):
Fixes the issue nicely - good job!

PR #1386, opened 2023-12-03
Fix select after rename; fix rename folder; disallow rename if unsaved edits

### LINKED ISSUE #1385: Click on sidebar item after cancelled rename does not open document
### What Happened?

After cancelling a rename operation on a sidebar fileitem, a different item was clicked.  The item was selected but the corresponding document did not open. Clicking again on the same item had no effect.  To open the document it was necessary to select a different item then the desired item .

### Steps to Reproduce

1.  Open a project with multiple documents
2. Start to rename one document bookmark but cancel the rename with `Esc`
3. Click on a different document bookmark
4. The second document does not open

### Expected Behavior

Document should open as usual after a renaming is cancelled.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1385 

 - [x] Instead of ignoring the next select when a fileitem is renamed, the renamed item is (closed) and reopened.
 - [x] If a fileitem rename is cancelled, the item is deselected so it can be clicked on to open
 - [x] The rename menu option is insensitive if the file has unsaved changes in any Code window 
 - [x] Code to rename folders was pre-existing but did not work. This PR also fixes that (needs checking for side-effects - should we allow renaming of folders at all?)

A Timeout had to be used to detect when renaming was cancelled as Granite Sourcelist does not produce any notification of this :disappointed: so to the `editing` property (read-only) had to be polled.


colinkiama:
Is this still a draft?

jeremypw:
@colinkiama Oops, no - I forgot to change the status.

colinkiama (changes requested):
Seems like you fixed the issue with the files.

However, after you rename a folder, the first file you select doesn't open in the editor.

Maybe `view.unselect_call ()` needs to be called again after you finish editing the folder name?

jeremypw:
@colinkiama Thanks for the review!  Your right that everything should be deselected after renaming - I'll check why a file may become selected after a folder rename.

jeremypw:
@colinkiama  Hopefully fixed now.

colinkiama:
Great! That issue has been fixed however, there's now another problem:
If you cancel a folder rename operation with the "Escape" key, you won't be able to select files. The files do not open the editor either.

jeremypw:
@colinkiama Thanks for being thorough! This is puzzling as the item is set back to being unselectable after editing ends yet it appears still selected and stops the listbox selecting anything else for some reason.  Investigating.

jeremypw:
@colinkiama I think I fixed it - it was due to some bug in SourceList I think. Seems to work if you set "selectable" to false *after* unselecting all :shrug:

colinkiama (approved):

PR #1389, opened 2023-12-15
Synchronise and save symbol outline width.

### LINKED ISSUE #1329: Symbol pane changes size after reopening code
### What Happened?

When closing Code with the Symbol Outline pane showing and then reopening it the Symbol Outline pane has increased in size, filling over halve of the editor window.

Window before closing:
image

Window after reopening:
image

This only affects the last opened vala file. For all of the other vala files the size of the Symbol Outline pane actually decreases.
Window after reopening (other vala files):
image

### Steps to Reproduce

1. Open a vala file in Code and open the right hand Symbol Outline pane.
2. Close Code.
3. Reopen Code and the Symbol Outline pane has increased in size.

### Expected Behavior

The Symbol Outline pane should consistently keep its size even after closing and reopening code. You should not have to resize it for every file after reopening code because it is now either to large or to small.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1329 

 - Add a new saved-state key `outline-width`
 - Bind to corresponding property of DocumentView
 - Keep Documents, DocumentView and saved-state settings in sync

A consequence of the PR having a single setting is that the outline pane width is always the same in all documents rather than individually adjustable but this is unlikely to be an issue.

### DIFF
--- a/data/io.elementary.code.gschema.xml
+++ b/data/io.elementary.code.gschema.xml
@@ -47,6 +47,11 @@
       <summary>Symbol outline visibility</summary>
       <description>Whether or not the symbol outline is visible</description>
     </key>
+    <key name="outline-width" type="i">
+      <default>160</default>
+      <summary>Symbol outline width</summary>
+      <description>Width of the symbol outline sidebar</description>
+    </key>
     <key name="last-opened-path" type="s">
       <default>''</default>
       <summary>Last opened path</summary>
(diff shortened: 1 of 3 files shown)


colinkiama (approved):
Looks good to me! Awesome!

PR #1395, opened 2024-01-04
Expose fuzzy find function and accelerator in sidebar action menu

jeremypw:
- Improves the discoverability of the plugin a little by adding a menu item to the folder manager action menu when it is active.  DRY shared code.
 - Consistently use "Find Project Files" to describe function
 -  Move the SearchProject class out of the main  plugin file and into its own file for clarity.


jeremypw:
The UX of this function can be revisited should it be promoted to the main code.

colinkiama (approved):
Awesome!

PR #1396, opened 2024-01-04
Associate popover with foldermanager

jeremypw:
- Point popover to the folder manager since that is where it is searching
 - Show folder manager temporarily if it is hidden when popover is shown (else points to random position)

I am not sure about this one.  The issue I am trying to address is where the popover should point if the foldermanager or searchbar is hidden. Feel free to reject.

### DIFF
--- a/plugins/fuzzy-search/fuzzy-search-popover.vala
+++ b/plugins/fuzzy-search/fuzzy-search-popover.vala
@@ -29,8 +29,7 @@ public class Scratch.FuzzySearchPopover : Gtk.Popover {
     public FuzzySearchPopover (Gee.HashMap<string, Services.SearchProject> pps, Scratch.MainWindow window) {
         Object (
             modal: true,
-            relative_to: window.sidebar,
-            constrain_to: Gtk.PopoverConstraint.WINDOW,
+            relative_to: window.document_view,
             width_request: 500,
             current_window: window
         );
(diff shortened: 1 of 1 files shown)


colinkiama (changes requested):
I think that it's a good idea but it's really distracting as it is now because of how the popover keeps shrinking and growing from the centre.

On my machine I've tried to manually set the `pointed_to` property of the popover so that the popover is near the top of the sidebar:
Screenshot from 2024-01-04 21 29 53

The popover grows and shrinks from the top now however, the initial position of the popover is wrong if the app starts with the sidebar hidden:
Screenshot from 2024-01-04 21 49 18

Change I made to achieve this:
```vala
public FuzzySearchPopover (Gee.HashMap<string, Services.SearchProject> pps, Scratch.MainWindow window) {
    Object (
        modal: true,
        relative_to: window.sidebar,
        constrain_to: Gtk.PopoverConstraint.WINDOW,
        width_request: 500,
        current_window: window,
        position: Gtk.PositionType.LEFT,
        pointing_to: Gdk.Rectangle () { x = window.sidebar.get_allocated_width (), y = 36 }
    );

    //...
```
I think that if the issue with the initial position is solved then this is okay to merge

jeremypw:
@colinkiama Thanks for looking at this. I considered pointing to a fixed rectangle but as you say it does not work so well if the sidebar is hidden.  We could point to the left-hand edge of the window or move the rectangle before showing the popover (but that feels hacky).  I'll work on this some more.

jeremypw:
I have changed it so that the popover points to the top left corner of the document now, whether or not the sidebar is open.  Its not ideal but I do not think it worth doing more at this stage. It will have to be reworked anyway when moved to main code and/or ported to Gtk4.  In Gtk4 there is more control over the positioning and appearance of popovers.

colinkiama (approved):
I agree

PR #1397, opened 2024-01-10
Make fuzzy-find less fuzzy in some respects

jeremypw:
- Prioritise sequential matches at start of filename
 - Match folder exactly if specified
 - Do not split matched parts across directory separator in path

With this PR if you enter something like "src/c" it will find files starting c (or containing) in or below the "src" folder.  

"src/w" will also match all files in "src/Widgets/" and any other folder containing "w" and any file below "src" containing "w" which I am not sure is a desirable behaviour but that can be fixed separately if not.

Multiple directory separators are now accepted but note that only the first folder must be an exact match.  So "src/w/s"
Will match all filenames containing "s" in foldername containing "w" below "src" .

 "cool" in "code/po/lu.po" is not now matched

### DIFF
--- a/plugins/fuzzy-search/fuzzy-finder.vala
+++ b/plugins/fuzzy-search/fuzzy-finder.vala
@@ -6,11 +6,11 @@
  *              Colin Kiama <colinkiama@gmail.com>
  */
 
-const int SEQUENTIAL_BONUS = 15; // bonus for adjacent matches
 const int SEPARATOR_BONUS = 30; // bonus if match occurs after a separator
+const int SEQUENTIAL_BONUS = 40; // bonus for adjacent matches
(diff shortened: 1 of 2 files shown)


colinkiama (approved):
This works great! You get the results you expect in fewer strokes then before.

However, I do miss the flexibility offered by the old search algorithm. Maybe the old algorithm could return as an option one day?

jeremypw:
> However, I do miss the flexibility offered by the old search algorithm. Maybe the old algorithm could return as an option one day?

Could you raise an issue indicating the kind of search / results that you wish to support? Thanks.  I did think about adding some controls to the popover or appmenu but didn't want to complicate things too much at this stage.  It would be easier when this feature is absorbed into main code.

PR #1399, opened 2024-01-11
Allow match start not at beginning

jeremypw:
Corrects unintended restriction on fuzzy find.

This allows the search string (after any first specified folder) to match at any position in the target string

Test search "ym" and "src/ym" in the Code project for example


colinkiama (approved):
Ah much better

PR #1405, opened 2024-01-15
[Feature]Implement "open-in-terminal" action

### LINKED ISSUE #58: Terminal extention open in current directory.

### LINKED ISSUE #600: Terminal automatically "cd" to project folder in the sidebar
It would be nice if Code terminal could "cd" to a project folder in the sidebar I choose to go to. At least an option to do that when I right-click the folder would be fine!

jeremypw:
Fixes #600 
Fixes #58 

 - Add menuitem to project folder item to open in terminal pane
 - Add hot-key for action
 - Trigger action on manual project change if terminal is visible
 - Add menu option to all sidebar items (doc items open their parent folder)

This is a convenient way of navigating the terminal pane to the project root folder when the active project is changed.
Current limitations:  If a foreground process is running in the terminal then it is killed without warning when the active project changes. It seems unlikely that a user would change project while e.g. compiling but if necessary additional code to show a warning can be added (like Terminal).

Possible future refinements: 
 - Have option/setting to navigate to the `/build` folder instead when active project changes


src/FolderManager/ProjectFolderItem.vala
```diff
@@ -116,6 +116,11 @@ namespace Scratch.FolderManager {
         }
 
         public override Gtk.Menu? get_context_menu () {
+            var open_in_terminal_pane_item = new Gtk.MenuItem.with_label (_("Open in Terminal Pane")) {
+                action_name = MainWindow.ACTION_PREFIX + MainWindow.ACTION_OPEN_IN_TERMINAL,
+                action_target = new Variant.string (file.file.get_path ())
+            };
+
```
colinkiama, line 123: This is great! Just one more thing. Add an `Granite.AccelLabel` to this menu item.

The keyboard shortcut for the "Open In Terminal" action isn't discoverable otherwise.

colinkiama (changes requested):

src/FolderManager/FileItem.vala
```diff
@@ -28,6 +28,10 @@ namespace Scratch.FolderManager {
         }
 
         public override Gtk.Menu? get_context_menu () {
+            var open_in_terminal_pane_item = new Gtk.MenuItem.with_label (_("Open in Terminal Pane")) {
+                action_name = MainWindow.ACTION_PREFIX + MainWindow.ACTION_OPEN_IN_TERMINAL,
+                action_target = new Variant.string (file.file.get_parent ().get_path ())
+            };
```
colinkiama, line 34:
```suggestion
            };

```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -116,6 +116,16 @@ namespace Scratch.FolderManager {
         }
 
         public override Gtk.Menu? get_context_menu () {
+            var open_in_terminal_pane_label = new Granite.AccelLabel.from_action_name (
+               _("Open in Terminal Pane"),
+                MainWindow.ACTION_PREFIX + MainWindow.ACTION_OPEN_IN_TERMINAL + "::"
+            );
```
colinkiama, line 122:
```suggestion
            );

```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -116,6 +116,16 @@ namespace Scratch.FolderManager {
...
+            var open_in_terminal_pane_item = new Gtk.MenuItem () {
+                action_name = MainWindow.ACTION_PREFIX + MainWindow.ACTION_OPEN_IN_TERMINAL,
+                action_target = new Variant.string (file.file.get_path ())
+            };
```
colinkiama, line 126:
```suggestion
            };

```

colinkiama (changes requested):
Just need to fix these small code style issues then it's all good!

src/FolderManager/ProjectFolderItem.vala
jeremypw: I don't think elementary code-style requires a blank line after a statement - even multi-line ones? They are only mandated after `if {}` and other blocks that do not end in `;`

jeremypw:
@colinkiama I have questioned those style changes.  It is usual for statements that relate to the same object not to be separated by blank lines, I think.

colinkiama (approved):

jeremypw:
@colinkiama Well, I could be wrong - I was going on what I've seen Danielle do in her recent PRs.  But yes, clarification of docs would be good.

jeremypw:
@colinkiama Looking at the docs it says that an empty line should follow a line consisting of `}` or `)` but in this case the line consists of `};` so strictly speaking the rule does not apply.  Could be clearer though.

PR #1409, opened 2024-01-19
ValaSymbolOutline: Lose global header

### LINKED ISSUE #378: Symbols plugin has redundant header
screenshot from 2018-03-22 18 29 10 2x

The label "Symbols" and the ability to collapse that header doesn't seem terribly useful.

jeremypw:
Fixes #378 

Because SourceList requires the toplevel items to be headers, it seems, it was necessary to add another header "Other Symbols" when there are symbols that would not otherwise appear (e.g. those outside of classes or an existing namespace).  This complicates the code significantly.

The CSymbolOutline was not touched as there are usually a large number of symbols at the top level.

### DIFF
--- a/src/SymbolPane/SymbolOutline.vala
+++ b/src/SymbolPane/SymbolOutline.vala
@@ -27,8 +27,6 @@ public abstract class Scratch.Services.SymbolOutline : Object {
 
     construct {
         store = new Granite.Widgets.SourceList ();
-        root = new Granite.Widgets.SourceList.ExpandableItem (_("Symbols"));
-        store.root.add (root);
 
         set_up_css ();
     }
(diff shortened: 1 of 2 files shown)


danirabbit:
I'm not sure if this really solves the issue since in most cases we just end up with another single toplevel item. Perhaps we should revisit this after/as part of GTK 4 porting not using SourceList?

jeremypw:
@danirabbit  I think solves the issue as strictly stated in the original report in that it gets rid of the redundant "Symbols" header which has nothing to do with the structure of the code.  The presence of remaining topevel items reflects the structure of the code as returned by the parser so cannot be fixed without applying further rules to hide some Vala symbols and not others which I am  :-1: on but, if desired, should be raised in (another) issue.

You're probably right that it will be easier to address the original issue by changing the widget in Gtk4 but that will not in itself fix the expanded issue.

jeremypw:
Closing as design reservations expressed.

PR #1414, opened 2024-01-23
ChooseProjectButton: Deactivate all project rows before selecting new active project

### LINKED ISSUE #1413: ProjectChooser: More than one active project selected under some circumstances
### What Happened?

After opening another project (with one already open) the second project was manually selected with the projectchooser button.  When the projectchooser was later reopened it was noticed that both open open projects were selected (checked).

### Steps to Reproduce

1. Have one project open
2. Close all open documents 
3. Open another project
4. Use ProjectChooser to select second project
5. Reopen ProjectChooser and note both entries selected

### Expected Behavior

Only one project should be selected (corresponding to the actual active project).

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1413 

Although the list box is in SINGLE selection mode, the radio buttons in the project rows are not affected by this so have to be explicitly unchecked when a new active project is chosen.

This PR makes the GitManager the source of truth regarding the active_project_path.  Some public access is removed. It also simplifies the ProjectRows and makes sure they remain grouped even when projects are removed and are re-added.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -107,6 +107,7 @@ namespace Scratch {
 
         private ulong color_scheme_listener_handler_id = 0;
 
+        private Services.GitManager git_manager;
 
         private const ActionEntry[] ACTION_ENTRIES = {
             { ACTION_FIND, action_fetch, "s" },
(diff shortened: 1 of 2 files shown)


colinkiama:
I can still produce the issue:
Screenshot from 2024-01-24 09 47 28

jeremypw:
That's weird - I was sure it was fixed but now it isnt :disappointed:

jeremypw:
This proving trickier than expected to fix in all circumstances so I'm going to convert to draft after have a deeper rethink.

jeremypw:
@colinkiama Hopefully fixed now!

PR #1415, opened 2024-01-24
[Feature] Implement default build directory setting

jeremypw:
This PR allows the setting of a global default build directory (relative to the root of projects) so that when the path of the terminal pane is changed by manually selecting the active project or using the ProjectFolderItem context menu, the desired build location within the project is automatically used.

Screenshot from 2024-01-24 18 48 32

### DIFF
--- a/data/io.elementary.code.gschema.xml
+++ b/data/io.elementary.code.gschema.xml
@@ -151,6 +151,11 @@
       <summary>Remember the last focused document.</summary>
       <description>Restore the focused document from a previous session when opening Code.</description>
     </key>
+    <key name="default-build-directory" type="s">
+      <default>''</default>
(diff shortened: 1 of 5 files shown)


colinkiama (changes requested):
Great! Except for one thing: The "Open In Terminal Pane" keyboard shortcut (CTRL + ALT + SHIFT + T) doesn't use the default build directory.

jeremypw:
@colinkiama Oops - forgot that.  Now fixed :smile:  

Ideally we want to be able to override the default on a per project basis because e.g. a Flatpak project will have a different default to a native project but we can iterate on this simple implementation.

colinkiama (approved):
Awesome. it's all fixed now! 👍️

PR #1417, opened 2024-01-25
Fix hang when two project names share the same prefix

jeremypw:
* Match complete directory name for prefix in two places
* Ensure set active path only activates one row


colinkiama (approved):

jeremypw:
@colinkiama Thanks for your prompt action!   I don't think this problem occurs in the last release fortunately :relieved:  but I'd like to get another release out fairly soon to showcase our latest features!

Issue #1157, opened 2021-12-13
Code


danrabbit:

colinkiama:
> Make Granite.SourceList Internal

So does this mean that Granite.SourceList should be moved into the elementary Code codebase?

danirabbit:
@colinkiama yes since there is no SourceList in Granite 7, Code will need its own sidebar class

colinkiama:
> Things to do to prepare:
> 
> * [ ]  Use Hdy.Tabbar
> * [x]  Make Granite.SourceList Internal
> * [ ]  Replace Gtk.MenuItem with GLib.MenuModel

Does "Replace Gtk.MenuItem with GLib.MenuModel" include the SourceList widget? 
It's quite a bit of effort for something that will be completely be rewritten in the GTK 4 port. We also have the issue of `Gtk.PopoverMenu` not using elementary styling in GTK 3 (unless I'm missing something):

![PopoverMenu in sidebar which has GNOME styling instead of the elementary widget styling](https://github.com/elementary/code/assets/16501024/141ee045-c94f-4240-98f6-b43322ba44f3)

danirabbit:
@colinkiama you can use Glib.Menu with Gtk.Menu for Gtk3 and then the port from Gtk.Menu to Gtk.PopoverMenu in Gtk4 is much smaller than going from Gtk.MenuItem all the way to menu models at the same time. Basically just trying to break things down as much as possible into reviewable branches so we're not trying to merge a several thousand line diff that nobody can reasonably review

PR #1419, opened 2024-01-25
[GTK4 Prep] Moved Granite.Widgets.SourceList source code into codebase

colinkiama:
Part of #1157


jeremypw:
Looks like Mail took this route too - see https://github.com/elementary/mail/pull/827.

The SourceList code will need some significant changes to work in Gtk4 though, e.g. Drag and Drop but it may still be easier than finding or build a native Gtk4 replacement.

I suggest you create a separate `SourceList` folder rather than mixing with other widgets.

colinkiama:
SourceList files are in a `SourceList` folder now.

jeremypw (approved):

PR #1423, opened 2024-02-22
Put branch name on separate line

### LINKED ISSUE #574: UX: git branch not readable if dir name is too long.
If the dir name is too long, the git branch is too truncated or hidden.
In this case you cant see the current git branch.

Possible solution:
Have a second line just for the name of the current git branch.

jeremypw:
Fixes #574 

Makes current branch visible without excessive sidebar width when both project and branch have long names
Screenshot from 2024-02-22 11 03 04

### DIFF
--- a/src/FolderManager/ProjectFolderItem.vala
+++ b/src/FolderManager/ProjectFolderItem.vala
@@ -59,7 +59,7 @@ namespace Scratch.FolderManager {
             if (monitored_repo != null) {
                 //As SourceList items are not widgets we have to use markup to change appearance of text.
                 if (monitored_repo.head_is_branch) {
-                    markup = "%s <span size='small' weight='normal'>%s</span>".printf (
+                    markup = "<span size='larger'>%s</span>\n<span size='small' weight='normal'>%s</span>".printf (
                         name, monitored_repo.branch_name
                     );
                 } else { //Distinguish detached heads visually
--- a/src/FolderManager/ProjectFolderItem.vala
+++ b/src/FolderManager/ProjectFolderItem.vala
@@ -59,7 +59,7 @@ namespace Scratch.FolderManager {
             if (monitored_repo != null) {
                 //As SourceList items are not widgets we have to use markup to change appearance of text.
                 if (monitored_repo.head_is_branch) {
-                    markup = "<span size='larger'>%s</span>\n<span size='small' weight='normal'>%s</span>".printf (
+                    markup = "<span size='medium'>%s</span>\n<span size='small' weight='normal'>%s</span>".printf (
                         name, monitored_repo.branch_name
                     );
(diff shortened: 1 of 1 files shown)


zeebok (approved):
Code looks good. Is this something that needs to be run by UX?

danirabbit (changes requested):
I think the newline for the branch name makes sense here, but I'm -1 on making the project name even larger

jeremypw:
I find it useful to have the project names more prominent and easily readable but maybe I need new glasses ...

jeremypw:
I have left this project name size slightly increased  as it was smaller than the file items before (!).  I really think it needs to stand out at least a little.  This is how it looks now:
Screenshot from 2024-03-13 19 45 53

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -59,7 +59,7 @@ namespace Scratch.FolderManager {
             if (monitored_repo != null) {
                 //As SourceList items are not widgets we have to use markup to change appearance of text.
                 if (monitored_repo.head_is_branch) {
-                    markup = "<span size='larger'>%s</span>\n<span size='small' weight='normal'>%s</span>".printf (
+                    markup = "<span size='medium'>%s</span>\n<span size='small' weight='normal'>%s</span>".printf (
```
danirabbit, line 62: I don't think the `span` is needed here :p 
```suggestion
                    markup = "%s\n<span size='small' weight='normal'>%s</span>".printf (
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -59,7 +59,7 @@ namespace Scratch.FolderManager {
             if (monitored_repo != null) {
                 //As SourceList items are not widgets we have to use markup to change appearance of text.
                 if (monitored_repo.head_is_branch) {
-                    markup = "%s <span size='small' weight='normal'>%s</span>".printf (
+                    markup = "<span size='medium'>%s</span>\n<span size='small' weight='normal'>%s</span>".printf (
                         name, monitored_repo.branch_name
                     );
                 } else { //Distinguish detached heads visually
```
danirabbit, line 65: don't forget to add newline to the line below this also

src/FolderManager/ProjectFolderItem.vala
jeremypw: Its not larger than the folder names which was my point, although it is bold.  I would prefer it more prominent myself but in the interest of getting something merged I'll take your suggestion.

danirabbit: If sidebar headers need to be adjusted this should be done in the platform stylesheet

danirabbit (approved):

PR #1424, opened 2024-02-29
[GTK4 Prep] Update DocumentView to use Hdy.TabView

colinkiama:
DocumentView widget has been rewritten to use the `Hdy.TabView` widget. 

The majority of the work is done. There are a few more things that left to do but I haven't figured out how to do these yet (Any help would be very appreciated!):
- [x] Add "Open in a New Window" menu option to tab view in DocumentView.
- [x] Support opening drag and drop operations with tabs
  - [x] Drag tab out of window to open tab in new window
  - [x] Drag tab into window to move tab from one window to another

Part of #1157


src/Services/Document.vala
```diff
@@ -113,6 +142,20 @@ namespace Scratch.Services {
             }
         }
 
+        private bool _loading = false;
```
jeremypw, line 145: Do we need this private variable? Can we not just return `tab.page` (or false if `tab` is null).  Can `loading` be replaced entirely with `tab.working` inline?

src/Services/Document.vala
```diff
@@ -61,10 +65,21 @@ namespace Scratch.Services {
             }
         }
 
-        public string tab_name {
+         public string tab_name {
```
jeremypw, line 68: Do we need both `tab_name` and `title`?

src/Widgets/DocumentView.vala
```diff
@@ -115,78 +140,61 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
...
+    public void duplicate_tab () {
+        if (current_document != null) {
+            new_document_from_clipboard (current_document.get_text ());
+        }
+    }
 
-        string new_text_file = _("Text file from %s:%d").printf (
-                                    timestamp.format ("%Y-%m-%d %H:%M:%S"), timestamp.get_microsecond ()
-                                );
+    public Services.Document search_for_document_in_tab (Hdy.TabPage tab) {
```
jeremypw, line 176: I don't understand what this does I'm afraid.  Can't we just get the document from the child of the `TabPage` parameter?

jeremypw (commented):
Just a few initial comment from looking at the code.

src/Services/Document.vala
colinkiama: You're right, the private variable doesn't need to be there

src/Services/Document.vala
colinkiama: Nope. It's possible to just use `title`

src/Widgets/DocumentView.vala
colinkiama: We can get the document from the child of the `TabPage` instead. `search_for_document_in_tab` was a supposed to be a temporary workaround for issues I had when I started working on this pull request.

src/Services/Document.vala
colinkiama: Only `title` is used now

src/Services/Document.vala
colinkiama: Private `_loading` variable is gone now

jeremypw (changes requested):
Thanks for working on this. It mostly works now, just a couple of things I notice:
1) The tab background colour does not follow the "solarized light" theme (unlike `main`).  
2) The `tab_history` button is not connected up.  This button does not seem to be working in `main` either but if we are going to add it again here we might as well have it working.

colinkiama:
I've addressed the additional feedback you had.

colinkiama:
@jeremypw pinging because it has been a while

src/Widgets/DocumentView.vala
```diff
@@ -482,4 +628,41 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
...
+        close_tab_section.append (_("Close Tabs to the Right"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_CLOSE_TABS_TO_RIGHT);
+        close_tab_section.append (_("Close Other Tabs"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_CLOSE_OTHER_TABS);
+        close_tab_section.append (_("Close Tab"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_CLOSE_TAB + "::");
+
+        var open_tab_section = new Menu ();
+        open_tab_section.append (_("Open in New Window"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_MOVE_TAB_TO_NEW_WINDOW);
+        open_tab_section.append (_("Duplicate"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_DUPLICATE_TAB);
+
+        menu.append_section (null, close_tab_section);
+        menu.append_section (null, open_tab_section);
```
danirabbit, line 645: It looks like these sections are the other way around in Files:

Screenshot from 2024-05-28 11 23 55
Screenshot from 2024-05-28 11 24 08

src/Widgets/DocumentView.vala
```diff
@@ -482,4 +628,41 @@ public class Scratch.Widgets.DocumentView : Granite.Widgets.DynamicNotebook {
...
+        var menu = new GLib.Menu ();
+
+        var close_tab_section = new Menu ();
+        close_tab_section.append (_("Close Tabs to the Right"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_CLOSE_TABS_TO_RIGHT);
+        close_tab_section.append (_("Close Other Tabs"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_CLOSE_OTHER_TABS);
+        close_tab_section.append (_("Close Tab"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_CLOSE_TAB + "::");
+
+        var open_tab_section = new Menu ();
+        open_tab_section.append (_("Open in New Window"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_MOVE_TAB_TO_NEW_WINDOW);
+        open_tab_section.append (_("Duplicate"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_DUPLICATE_TAB);
```
danirabbit, line 642: This is "Duplicate Tab" in Files. Should be consistent :)
```suggestion
        open_tab_section.append (_("Duplicate Tab"), MainWindow.ACTION_PREFIX + MainWindow.ACTION_DUPLICATE_TAB);
```

danirabbit (changes requested):
Looks like we're missing a couple of accelerators that are present in Files:

Screenshot from 2024-05-28 11 23 55


Otherwise this looks and works good to me!

jeremypw (changes requested):
Thanks  for your continued work! The theming and closed tab button issues are now fixed and accels are showing.


One remaining behavioural issue I noticed is that the "Close Other Tabs" option is disabled when the selected tab is the first one. This should be enabled whenever there are more than one tab regardless of which is selected.

I'll use this branch for a while to check for any more regressions.

src/MainWindow.vala
```diff
@@ -182,6 +192,7 @@ namespace Scratch {
             action_accelerators.set (ACTION_GO_TO, "<Control>i");
             action_accelerators.set (ACTION_SORT_LINES, "F5");
             action_accelerators.set (ACTION_NEW_TAB, "<Control>n");
+            action_accelerators.set (ACTION_DUPLICATE_TAB, "<Control><Alt>t" );
```
jeremypw, line 195: This clashes with the "ACTION_TOGGLE_TERMINAL" accelerator.

colinkiama: I've updated the accelerator to use `CTRL + SHIFT + K`. This matches GNOME Web and Microsoft Edge. Is this okay?

jeremypw:
Requesting review from @danirabbit regarding what accels to use consistently for tab menu between Code, Files, Terminal and Web (if they need to be consistent).

danirabbit:
Looks like Chrome and Edge are also using "Ctrl Shift K". Safari uses Super Shift K. So it makes sense for us to be in line with what appears to be a de facto standard here. We should probably add this shortcut to Files too

jeremypw:
OK, I'll leave it as is here and modify Files and Terminal to match.

jeremypw (approved):

danirabbit (approved):
Let's go! :rocket:

jeremypw:
Just found another regression - dropping a file from Files onto the tabbar (empty space) no longer opens it in a new tab.  @colinkiama   Would you like to submit a new PR fixing that? If not I'll try to fix it some time.

colinkiama:
I won't be working on this until next week. Feel free to fix this in the meantime.

jeremypw:
@colinkiama I could not get dropping onto the empty tabbar space to work in my Terminal branch https://github.com/elementary/terminal/pull/760 - there was no obvious way to access a widget for that area. I could only implement dropping onto the add button and tab labels.  In Code, dropping onto the add button already works.  I can implement dropping onto tab labels easily enough.

jeremypw:
Before converting to Hdy.TabView, Code did not support dropping onto the tab labels, I guess because it is not clear what should happen? Should the contents of the dropped file replace the contents of the document (destructive), should it be appended to the document,  should the document be closed and a new one created from the dropped file, replacing the old document or should a new tab be created?

jeremypw:
Gedit only implements drop on the add button anyway so maybe we can just leave this (unless someone can find a way implementing drop on empty space

PR #1428, opened 2024-04-14
[GTK4 Prep] Replace menu items with menu model

colinkiama:
I replaced as much as I possibly could.

The SourceView requires GTK4. We'd have access to [`Gtk.TextView.extra_menu`](https://valadoc.org/gtk4/Gtk.TextView.extra_menu.html).

Also, the PopoverMenus used for the replaced Gtk.MenuItems currently don't show keyboard shortcuts. As far as I know, this should be fixed after moving over to GTK4.

Part of #1157.


src/FolderManager/FileItem.vala
```diff
@@ -27,43 +27,18 @@ namespace Scratch.FolderManager {
...
-            open_in_menu.add (files_menuitem);
-
-            var contractor_menu = new Gtk.Menu ();
+        public override GLib.Menu? get_context_menu () {
+            var open_in_terminal_pane_item = new GLib.MenuItem (_("Open in Terminal Pane"),
+                                                                MainWindow.ACTION_PREFIX
+                                                                + MainWindow.ACTION_OPEN_IN_TERMINAL
+                                                                + "::"
+                                                                + file.path);
+
```
danirabbit, line 36: Code style for wrapping should be like:
```suggestion
            var open_in_terminal_pane_item = new GLib.MenuItem (
                _("Open in Terminal Pane"),
                MainWindow.ACTION_PREFIX
                + MainWindow.ACTION_OPEN_IN_TERMINAL
                + "::"
                + file.path
            );
```

There's actually a function for creating detailed action names as well: https://valadoc.org/gio-2.0/GLib.Action.print_detailed_name.html

danirabbit (changes requested):
Haven't tested this yet or looked into it too much, but just made a note about code style and detailed action names.

Since there's issues in Gtk3 with PopoverMenu, you can use Gtk.Menu with a model https://valadoc.org/gtk+-3.0/Gtk.Menu.Menu.from_model.html and then it's a pretty easy drop-in replacement during the GTK 4 port to change that

Since this is quite a huge branch I wonder if it might be better to do this in smaller steps, like just a branch to switch to using GLib.Actions I imagine is already a huge review in itself. It might help to get your work merged in faster if branches are more incremental

colinkiama:
Sure, I'll split this up into multiple pull requests.

PR #1434, opened 2024-04-26
changing the order of the right click new items menu

MobileAZN:
I am updating the order of the 'right-click' menu items in Code when right -clicking on a folder in the file pane.  The items were out of alphabetical order, but also, I feel people are more likely to create a new file vs a new folder, thus making navigating/mousing this menu easier.

I also swapped the order of the Gtk.MenuItem in the source (lines 218-223) to maintain the order following the menu creation and readability, however it isn't necessary.  Only a swap of lines 225 and 226 are truly needed for this change.


danirabbit:
Hey thanks for your branch. I'm -1 on this change because it would introduce an inconsistency between Files and Code. Menu items are almost never in alphabetical order

Closing as "Design conflict" since it would introduce inconsistency between apps

PR #1438, opened 2024-06-17
Define terminal menu items using GLib.MenuModel

colinkiama:
Replaces usage of Gtk.MenuItem with GLib.MenuModel in the Terminal widget as #1157

### DIFF
--- a/src/Widgets/Terminal.vala
+++ b/src/Widgets/Terminal.vala
@@ -5,13 +5,23 @@
  */
 
 public class Code.Terminal : Gtk.Box {
+    public const string ACTION_GROUP = "term";
+    public const string ACTION_PREFIX = ACTION_GROUP + ".";
+    public const string ACTION_COPY = "action_copy";
+    public const string ACTION_PASTE = "action_paste";
+
     private const double MAX_SCALE = 5.0;
     private const double MIN_SCALE = 0.2;
     private const string LEGACY_SETTINGS_SCHEMA = "org.pantheon.terminal.settings";
     private const string SETTINGS_SCHEMA = "io.elementary.terminal.settings";
 
-    private GLib.Pid child_pid;
     public Vte.Terminal terminal { get; construct; }
+    public SimpleActionGroup actions { get; construct; }
+
+    private GLib.Pid child_pid;
+    private SimpleAction copy_action;
+    private SimpleAction paste_action;
+    private Gtk.Clipboard current_clipboard;
 
     construct {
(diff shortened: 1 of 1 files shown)


src/Widgets/Terminal.vala
```diff
@@ -36,29 +46,44 @@ public class Code.Terminal : Gtk.Box {
...
                 menu.select_first (false);
                 menu.popup_at_pointer (event);
             }
             return false;
         });
 
+        realize.connect (() => {
+            current_clipboard = terminal.get_clipboard (Gdk.SELECTION_CLIPBOARD);
+            copy_action.set_enabled (terminal.get_has_selection ());
+
```
danirabbit, line 80: extra whitespace

src/Widgets/Terminal.vala
```diff
@@ -74,6 +99,14 @@ public class Code.Terminal : Gtk.Box {
...
     }
 
+    public void copy () {
+        terminal.copy_clipboard ();
+    }
+
+    public void paste () {
+        terminal.paste_clipboard ();
+    }
+
```
danirabbit, line 109: Is there a reason to add new public methods here? Seems like this is simple enough we could just keep it as lambdas as before

src/Widgets/Terminal.vala
```diff
@@ -5,13 +5,23 @@
...
     private const string LEGACY_SETTINGS_SCHEMA = "org.pantheon.terminal.settings";
     private const string SETTINGS_SCHEMA = "io.elementary.terminal.settings";
 
-    private GLib.Pid child_pid;
     public Vte.Terminal terminal { get; construct; }
+    public SimpleActionGroup actions { get; construct; }
+
+    private GLib.Pid child_pid;
+    private SimpleAction copy_action;
+    private SimpleAction paste_action;
```
danirabbit, line 23: Do we need to hold a reference for these actions here? It seems like they're only used in the construct method right?

danirabbit (changes requested):
Looks good and works as expected! Just a couple of comments

src/Widgets/Terminal.vala
colinkiama: No reason. I moved the logic back in to lambdas.

src/Widgets/Terminal.vala
colinkiama: Thanks for catching that! We don't. These have been moved to the construct block.

danirabbit (approved):
Nice work :tada:

PR #1441, opened 2024-06-23
Plugins and sidebar now use glib menu model for actions

colinkiama:
As part of "Replace Gtk.MenuItem with GLib.MenuModel" task in #1157


danirabbit (approved):
This looks great. Way to clean up key press handling!

PR #1444, opened 2024-06-27
[GTK4 Prep] Add FileView actions

colinkiama:
Adds actions to FileView to prepare to use GLib.MenuModel for context menus in FileView items. As part of #1157


src/FolderManager/FileView.vala
```diff
@@ -54,12 +76,47 @@ public class Scratch.FolderManager.FileView : Code.Widgets.SourceList, Code.Pane
 
         git_manager = Scratch.Services.GitManager.get_instance ();
 
+        rename_folder_action = new SimpleAction (ACTION_RENAME_FOLDER, null);
```
danirabbit, line 79: Is there a reason this is a separate SimpleAction instead of part of ActionEntries?

src/FolderManager/FileView.vala
```diff
@@ -22,11 +22,33 @@
...
+        { ACTION_DELETE, action_delete, "s" },
+        { ACTION_NEW_FILE, add_new_file, "s" },
+        { ACTION_NEW_FOLDER, add_new_folder, "s"},
+        { ACTION_CLOSE_FOLDER, action_close_folder, "s"},
+        { ACTION_CLOSE_OTHER_FOLDERS, action_close_other_folders, "s"}
+    };
+
+    public SimpleActionGroup actions { get; construct; }
+    public ActionGroup toplevel_action_group { get; private set; }
+    public SimpleAction rename_folder_action { get; private set; }
```
danirabbit, line 51: Is there a reason for these to be public?

src/FolderManager/FileView.vala
```diff
@@ -22,11 +22,33 @@
...
+    private const ActionEntry[] ACTION_ENTRIES = {
+        { ACTION_RENAME_FILE, action_rename_file, "s" },
+        { ACTION_DELETE, action_delete, "s" },
+        { ACTION_NEW_FILE, add_new_file, "s" },
+        { ACTION_NEW_FOLDER, add_new_folder, "s"},
+        { ACTION_CLOSE_FOLDER, action_close_folder, "s"},
+        { ACTION_CLOSE_OTHER_FOLDERS, action_close_other_folders, "s"}
+    };
+
+    public SimpleActionGroup actions { get; construct; }
```
danirabbit, line 49: `construct` should be for construction properties. It doesn't look like this is set from inside a constructor

src/FolderManager/FileView.vala
```diff
@@ -22,11 +22,33 @@
...
  */
 public class Scratch.FolderManager.FileView : Code.Widgets.SourceList, Code.PaneSwitcher {
+    public const string ACTION_GROUP = "file-view";
+    public const string ACTION_PREFIX = ACTION_GROUP + ".";
+    public const string ACTION_RENAME_FILE = "rename-file";
+    public const string ACTION_RENAME_FOLDER = "rename-folder";
+    public const string ACTION_DELETE = "delete";
+    public const string ACTION_NEW_FILE = "new-file";
+    public const string ACTION_NEW_FOLDER = "new-folder";
+    public const string ACTION_CHANGE_BRANCH = "change-branch";
```
danirabbit, line 32: I don't see this action here?

danirabbit (changes requested):
A few inline comments

It also looks like none of these actions are used at all in this branch? Gtk.MenuItem is a Gtk.Actionable so it is possible to use GLib.Actions with Gtk.MenuItem without having to also port everything to GLib.Menu in the same commit.

Really the reason I'm asking for things to be broken up in separate commits to make sure they they are not only reviewable, but so that in a possible future git blame we can track changes more easily. So if we're introducing a bunch of dead code that's not really reviewable either even though the commit is smaller and it will make it harder to traverse history

src/FolderManager/FileView.vala
```diff
@@ -246,6 +303,30 @@ public class Scratch.FolderManager.FileView : Code.Widgets.SourceList, Code.Pane
         plugins.hook_folder_item_change (source, dest, event);
     }
 
+    public void rename_file (string path) {
```
danirabbit, line 306: Is there a reason this method is public?

colinkiama: Nope

src/FolderManager/FileView.vala
colinkiama: I must have thought it wasn't possible months ago for some reason. I've managed to include the action in ActionEntries now

src/FolderManager/FileView.vala
colinkiama: Not right now. Only one I've left as `public` is `toplevel_action_group` because it was already `public`.

src/FolderManager/FileView.vala
colinkiama: I've turned `actions` into a private field now.

src/FolderManager/FileView.vala
colinkiama: I've removed it now.

colinkiama:
I've made changes that should address all your comments.

FileView actions are now used by Gtk.MenuItems in FileView items.

src/FolderManager/FileItem.vala
```diff
@@ -120,35 +120,16 @@ namespace Scratch.FolderManager {
...
-                            view.unselect_all ();
-                            return Source.REMOVE;
-                        }
-                    });
-                }
-            });
+            var rename_item = new Gtk.MenuItem.with_label (_("Rename")) {
+                action_name = FileView.ACTION_PREFIX + FileView.ACTION_RENAME_FILE,
+                action_target = new Variant.string (file.path),
+                sensitive = view.rename_request (file)
```
danirabbit, line 126: We should set the action `enabled` property rather than setting the widget sensitivity. This will automatically set widget sensitivity but also make sure that if the action is used elsewhere we're only setting it in the one place and it affects any keyboard shortcuts too

src/FolderManager/FileView.vala
```diff
@@ -246,6 +297,57 @@ public class Scratch.FolderManager.FileView : Code.Widgets.SourceList, Code.Pane
...
+    }
+
+    private void rename_folder (string path) {
+        var folder_to_rename = find_path (root, path) as FolderItem;
+        if (folder_to_rename == null) {
+            critical ("Could not find folder from given path to rename: %s", path);
+        }
+
+        folder_to_rename.selectable = true;
+            if (start_editing_item (folder_to_rename)) {
```
danirabbit, line 331: indentation mistake here?

src/FolderManager/FileView.vala
```diff
@@ -246,6 +297,57 @@ public class Scratch.FolderManager.FileView : Code.Widgets.SourceList, Code.Pane
...
+                this.unselect_all ();
+                return Source.REMOVE;
+            }
+        });
+    }
+
+    private void rename_folder (string path) {
+        var folder_to_rename = find_path (root, path) as FolderItem;
+        if (folder_to_rename == null) {
+            critical ("Could not find folder from given path to rename: %s", path);
```
danirabbit, line 327: return here?

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -127,14 +127,16 @@ namespace Scratch.FolderManager {
...
+                action_target = new Variant.string (file.path)
+            };
 
-            var close_all_except_item = new Gtk.MenuItem.with_label (_("Close Other Folders"));
-            close_all_except_item.activate.connect (() => { close_all_except (); });
-            close_all_except_item.sensitive = view.root.children.size > 1;
+            var close_all_except_item = new Gtk.MenuItem.with_label (_("Close Other Folders")) {
+                action_name = FileView.ACTION_PREFIX + FileView.ACTION_CLOSE_OTHER_FOLDERS,
+                action_target = new Variant.string (file.path),
+                sensitive = view.root.children.size > 1
```
danirabbit, line 138: Same here with sensitivity, this should be handled by setting the action sensitive when the number of children change

danirabbit (changes requested):
Nice! Everything is working here afaict, just a few more comments

colinkiama:
I've now made changes to address your latest comments

src/FolderManager/FileView.vala
```diff
@@ -54,12 +73,44 @@ public class Scratch.FolderManager.FileView : Code.Widgets.SourceList, Code.Pane
...
+        if (path == null || path == "") {
+            return;
+        }
+
+        var project_item = find_path (root, path) as ProjectFolderItem;
+        if (project_item == null) {
+            return;
+        }
+
+        project_item.close_all_except ();
```
danirabbit, line 111: It looks like this sends a signal that is connected to in this class on line #476. So perhaps that whole block of code should just live here instead of sending signals back and forth?

src/FolderManager/FileView.vala
```diff
@@ -266,6 +369,83 @@ public class Scratch.FolderManager.FileView : Code.Widgets.SourceList, Code.Pane
...
+
+        if (path == null || path == "") {
+            return;
+        }
+
+        rename_folder (path);
+    }
+
+
+     private void action_delete (SimpleAction action, Variant? param) {
```
danirabbit, line 425: Extra whitespace and an indentation error?

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -127,14 +127,20 @@ namespace Scratch.FolderManager {
...
-            close_all_except_item.sensitive = view.root.children.size > 1;
+            var close_all_except_item = new Gtk.MenuItem.with_label (_("Close Other Folders")) {
+                action_name = FileView.ACTION_PREFIX + FileView.ACTION_CLOSE_OTHER_FOLDERS,
+                action_target = new Variant.string (file.path),
+            };
+            var close_other_folders_action = Utils.action_from_group (
+                FileView.ACTION_CLOSE_OTHER_FOLDERS,
+                view.actions
+            );
+            close_other_folders_action.set_enabled (view.root.children.size > 1);
```
danirabbit, line 143: This should be in `View` instead and be something like a function that is connected to `root.child_added` and `root.child_removed` and sets the action enabled state there

danirabbit (commented):
Couple more things here, but I think that's the last bit!

src/FolderManager/FileView.vala
colinkiama: This is done now

src/FolderManager/FileView.vala
colinkiama: Fixed!

danirabbit (approved):
Nice job!

PR #1446, opened 2024-06-30
Fix "go to line" behaviour when column number is not explicitly set

colinkiama:
When the value of `offset` in the SourceView widget's `go_to_line` method is less than 0, use `0` as the character position instead of `offset` to prevent cursor being moved to the line before the intended line number to go to.


jeremypw (approved):
Makes sense; the functions calling this could have offset = -1, at least theoretically.

PR #1447, opened 2024-07-05
Limit selection highlighting to single line and avoid crash

### LINKED ISSUE #1443: Code crashes when deleting lines
### What Happened?

### Steps to Reproduce

3. Press backspace

### Expected Behavior

### OS Version

7.x (Horus)

### Software Version

### Log Output

```shell
Segmentation fault
```

### Hardware Info

_No response_

jeremypw:
Fixes #1443 

This stops the highlight word selection plugin operating and sometimes causing a crash when selecting and deleting lines of text.

The exact cause of the crash was not determined as it was sporadic and did not yield a useful backtrace, but this seems to stop it (to be confirmed).

### DIFF
--- a/plugins/highlight-word-selection/highlight-word-selection.vala
+++ b/plugins/highlight-word-selection/highlight-word-selection.vala
@@ -53,17 +53,24 @@ public class Scratch.Plugins.HighlightSelectedWords : Peas.ExtensionBase, Peas.A
     public void on_selection_changed (ref Gtk.TextIter start, ref Gtk.TextIter end) {
         var window_search_context = main_window != null ? main_window.search_bar.search_context : null;
 
+        if (current_search_context != null) {
+            // Cancel existing search
+            current_search_context.set_highlight (false);
+            current_search_context = null;
+        }
+
+
+        // Perform plugin selection highlighting only when there is no ongoing and successful search in the window search bar
         if (window_search_context == null ||
             window_search_context.settings.search_text == "" ||
             window_search_context.get_occurrences_count () == 0) {
-            // Perform plugin selection when there is no ongoing and successful search 
-            current_search_context = new Gtk.SourceSearchContext (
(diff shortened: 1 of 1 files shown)


danirabbit:
I can't reproduce the original crash. Is there a sample file you can provide that reliably crashes?

jeremypw:
@danirabbit It did not seem particularly difficult to reproduce the crash provided the recipe above is followed - the important thing is that the selection includes the line ending and there are multiple highlights. I'll push a file in a minute though.

jeremypw:
@danirabbit Try this file. Place the cursor at the beginning of (e.g) `Stack trace of thread 2085:` then press `<Shift><Down>` to select the whole line (including the `CR/LF`).  Several similar line should highlight.  Press `BackSpace`.  Code crashes.

Make sure you have the `master` version installed (as the it is a plugin problem) and have the highlighting extension active.

[CodeDeleteTest.txt](https://github.com/user-attachments/files/16401205/CodeDeleteTest.txt)

danirabbit (changes requested):
Ah okay, so I can still reproduce the crash if I select a line by double clicking it, but only sometimes. I can't find good reproduction conditions.

But also, this breaks highlighting matches when selecting with the keyboard

So I dunno about this fix. It seems to only stop crashing sometimes and it breaks the functionality of the plugin, but also only sometimes

jeremypw:
@danirabbit Yes, its difficult to fix properly when the cause is obscure.  Must some kind of race I guess - I can reliably reproduce on my system and the fix seem reliable here too.

I'll convert to draft pending the cause being identified better.

zeebok:
This did not fix it for me. It is odd, it doesn't seem to be even hitting the plug-in code before it crashes, so this may be an issue with gtksourceview.

jeremypw:
Closing in favour of #1470

PR #1451, opened 2024-07-13
PreferencesDialog: make whitespace strings consistent

davidmhewitt:
We use "whitespace" here, so this should maybe be the same? Or maybe that one should be changed? Or maybe they should both have a hyphen? :laughing: 

But I think they should at least be consistent.

https://github.com/elementary/code/blob/a03ec88be58ec125d6e009799fd522c63a47f344/src/Dialogs/PreferencesDialog.vala#L25

### DIFF
--- a/src/Dialogs/PreferencesDialog.vala
+++ b/src/Dialogs/PreferencesDialog.vala
@@ -93,7 +93,7 @@ public class Scratch.Dialogs.Preferences : Granite.Dialog {
             }
         });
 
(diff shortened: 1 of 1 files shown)


danirabbit:
Looked around to try to find some kind of more definitive answer or styleguide here but the best I could find is that generally "whitespace" (as a single word) can refer to any character that is not visible (like tab or space) whereas "white space" refers more generally to the concept of leaving a blank space. So single word is probably appropriate here since we're talking about stripping trailing whitespace characters

danirabbit (approved):

PR #1454, opened 2024-07-18
Jeremypw/fix cancel folder creation

### LINKED ISSUE #1436: Cancelling new folder creation in project causes parent folder to disappear from sidebar
### What Happened?

After opening an empty project folder, a new subfolder was created using the project context menu. While in rename mode after creation of "untitled folder" the `Esc` key was pressed to cancel the operation. Both the subfolder and the project folder disappeared from the sidebar although the project was still active according to the project chooser button. The project folder still existed on disk.  Closing and restarting Code caused the project folder to reappear in the sidebar.

### Steps to Reproduce

1. Open an empty project folder
2. Start to create a new folder but cancel while in rename-after-create mode.
3. Observe parent folder disappears too

### Expected Behavior

The newly created folder should disappear if cancel is pressed before completing the create/rename operation, but not the parent folder.

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1436 

The dummy entry for an empty folder gets removed during new file creation/renaming but was not being replaced if the file never got created resulting the parent not being visible.  This PR fixes that.

Also fixes issue in OS8 where an empty file no longer has a text content type so was not being shown in the sidebar.

### DIFF
--- a/src/FolderManager/FolderItem.vala
+++ b/src/FolderManager/FolderItem.vala
@@ -252,8 +252,8 @@ namespace Scratch.FolderManager {
 
             view.ignore_next_select = true;
             ((Code.Widgets.SourceList.ExpandableItem)this).remove (item);
-            // Add back dummy if empty unless we are removing a rename item
-            if (!(item is RenameItem || has_dummy || n_children > 0)) {
+            // Add back dummy if empty
+            if (!(has_dummy || n_children > 0)) {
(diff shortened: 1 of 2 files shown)


danirabbit (approved):
Can confirm the fix. Nice work!

PR #1455, opened 2024-07-24
[GTK4 Prep] Move External App Actions Logic To Utils

colinkiama:
Part of of "Replace Gtk.MenuItem with GLib.MenuModel" task in #1157


src/Utils.vala
```diff
@@ -184,4 +184,109 @@ namespace Scratch.Utils {
...
+        var file = GLib.File.new_for_path (path);
+        var file_list = new List<GLib.File> ();
+        file_list.append (file);
+        var app_info = new GLib.DesktopAppInfo (app_id);
+        try {
+            app_info.launch (file_list, null);
+        } catch (Error e) {
+            warning (e.message);
+        }
+    }
```
danirabbit, line 246: `file_type` appears to be unused here. This seems to be a duplicate of `launch_app_with_file` but just with the one liner to create a GLib.File from path. Should this function exist?

src/Utils.vala
```diff
@@ -184,4 +184,109 @@ namespace Scratch.Utils {
...
+                    GLib.VariantType.STRING,
+                    { file.get_path (), app_id, file_type }
+                )
+            };
+
+            menuitem.add (menuitem_grid);
+            menu.add (menuitem);
+        }
+    }
+
```
danirabbit, line 223: extra space

src/Utils.vala
```diff
@@ -184,4 +184,109 @@ namespace Scratch.Utils {
...
+                menu.add (menu_item);
+            }
+        } catch (Error e) {
+            warning (e.message);
+        }
+
+        return menu;
+    }
+
```
danirabbit, line 272: extra space

danirabbit (changes requested):
Seems sensible. Just one question and a couple comments about extra whitespace

src/Utils.vala
colinkiama: No, it doesn't need to. It has been removed now

colinkiama:
I've made new changes that should resolve all the issues now.

src/Utils.vala
```diff
@@ -184,4 +184,94 @@ namespace Scratch.Utils {
...
+        } catch (Error e) {
+            warning (e.message);
+        }
+    }
+
+    public Gtk.Menu create_contract_items_for_file (GLib.File file, string file_type) {
+        var menu = new Gtk.Menu ();
+
+        try {
+            var contracts = Granite.Services.ContractorProxy.get_contracts_by_mime (file_type);
```
danirabbit, line 239: Should this be https://valadoc.org/granite-7/Granite.Services.ContractorProxy.get_contracts_for_file.html so we don't have to manually get file type?

src/Utils.vala
```diff
@@ -184,4 +184,94 @@ namespace Scratch.Utils {
...
+        }
+
+        return menu;
+    }
+
+    public void execute_contract_with_file_path (string path, string contract_name, string file_type) {
+        var file = GLib.File.new_for_path (path);
+
+        try {
+            var contracts = Granite.Services.ContractorProxy.get_contracts_by_mime (file_type);
```
danirabbit, line 264: Same here

src/Utils.vala
colinkiama: Yeah. That makes the code more concise. Thanks!

src/Utils.vala
colinkiama: This has been updated now too.

danirabbit (approved):
Nice work!

PR #1456, opened 2024-07-25
[GTK4 Prep] Use actions for New Window menu item in FileItem

colinkiama:
Part of of "Replace Gtk.MenuItem with GLib.MenuModel" task in https://github.com/elementary/code/issues/1157

After this, the none of menu items in FileView context menus will be activated using the `activate` signal. The switch to GLib.MenuModel will be simple after this.


danirabbit (approved):
Very straightforward :+1:

PR #1458, opened 2024-07-25
[GTK4 Prep] Create FileView Context Menu Items using GLib.MenuModel

colinkiama:
Completes the "Replace Gtk.MenuItem with GLib.MenuModel" task in #1157

### DIFF
--- a/src/FolderManager/FileItem.vala
+++ b/src/FolderManager/FileItem.vala
@@ -54,7 +54,7 @@ namespace Scratch.FolderManager {
             if (contractor_items.get_n_items () > 0) {
                 external_actions_section.append_submenu (
                     _("Other Actions"),
-                    Utils.create_contract_items_for_file (file.file)
+                    contractor_items
                 );
             }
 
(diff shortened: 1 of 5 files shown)


colinkiama:
A side effect I've noticed with this change is that keyboard accelerators are no longer displaying in the FileView context menus. Is this what was expected to happen?

danirabbit (changes requested):
Just one thing that looks off, otherwise seems like everything is good. Nice work! That was a lot of refactoring!

danirabbit:
@colinkiama Ah so looking at it before it looks like the menu items with accels didn't pass a path argument and you've added a path to the menuitems here? For example the New Branch item will show the accel if you change the detailed action to:

```
                GLib.Action.print_detailed_name (
                    MainWindow.ACTION_PREFIX + MainWindow.ACTION_NEW_BRANCH,
                    ""
                )
```

colinkiama:
I've fixed the issue with the accelerators by using the "accel" attribute supported by GLib.MenuModel.

danirabbit (approved):
works for me. Nice work!

Issue #1440, opened 2024-06-21
Opening folders with same name and on different partitions as projects causes Code to crash


jeremypw:
### What Happened?

Attempted to open two versions of a project stored on different partitions in the sidebar.  Code crashed on opening the second one and was unstable until the first project was also closed.

### Steps to Reproduce

1. Have two folders with the same name on different partitions 
2. Try to open both as projects 
3. Crash

### Expected Behavior

No crash.

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

aguillon:
Hi! Please note that I'm not one of the project maintainers.

Do you remember the name and paths of the two folders you added? I have been able to reproduce this by adding, e.g., `/usr` and `/home/arthur/usr`. I believe the issue appears when the two folders and their ancestors have the same names, pairwise, until we reach the root /.

Alternatively, would you have the time to try my fix?

jeremypw:
@aguillon 

In my case I was opening the same project on two different partitions so the paths were something like:
1) `/home/jeremypw/Project1`
2) `/media/jeremypw/<UID>/home/jeremypw/Project1`

Thanks for having a try at fixing this!  Welcome to elementaryOS developing! I'll have a look at your PR.

PR #1460, opened 2024-08-11
Fix the app crashing when two folders and their ancestors have the same names

aguillon:
Hi!

I think this fixes https://github.com/elementary/code/issues/1440. As mentioned there, a way to reproduce is to open the folders `/usr` and `$HOME/usr`. In this case, the issue is not so much that the two folders and their ancestors have the same names, but that `find_unique_path` does not handle recursion up to the root.

This is my first contribution to a Vala project, ever. In particular, I don't know how to call a code formatter or how to write unit tests (which would make a lot of sense with the present issue). If you show me how, I will gladly update my PR.

### DIFF
--- a/src/Utils.vala
+++ b/src/Utils.vala
@@ -150,10 +150,27 @@ namespace Scratch.Utils {
             return false;
         }
 
+        if (!f1.has_parent (null) || !f2.has_parent (null)) {
+            path1 = f1.get_basename ();
+            path2 = f2.get_basename ();
+            return true;
+        }
+
         var f1_parent = f1.get_parent ();
         var f2_parent = f2.get_parent ();
 
         while (f1_parent.get_relative_path (f1) == f2_parent.get_relative_path (f2)) {
+            // If f1 == /a/b and f2 == /.../a/b we still need to disambiguate with
+            // the one parent name that we have
+            // Both conditions cannot simultaneously be true
+            if (!f1_parent.has_parent (null)) {
+                f2_parent = f2_parent.get_parent ();
+                break;
+            }
+            if (!f2_parent.has_parent (null)) {
+                f1_parent = f1_parent.get_parent ();
+                break;
+            }
             f1_parent = f1_parent.get_parent ();
             f2_parent = f2_parent.get_parent ();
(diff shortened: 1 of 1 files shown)


jeremypw:
@aguillon I use `vala-lint` to flag up elementaryOS code style issues but no automatic code formatting at the moment. The CI will fail if there are serious code style errors (as it does currently).

You can get the source code of `vala-lint` from https://github.com/vala-lang/vala-lint.git.

The project does not current implement any unit tests so don't worry about that.  You can look at other projects like Files and Terminal for examples of how it is done.

I have quickly tried out your fix and it seems to be working!  I'll test it a bit more then it should be good to merge.

jeremypw (changes requested):
This fixed a the situation mentioned in the linked issue.  As `get_parent ()` can return `null` I think you need to perform a null check before referencing that parent.

aguillon:
Thank you for `vala-lint`.

> As get_parent () can return null I think you need to perform a null check before referencing that parent.

Thank you, indeed this could happen if we add the root as a project. This should work now.

src/Utils.vala
```diff
@@ -150,10 +150,27 @@ namespace Scratch.Utils {
...
+            if (!f1_parent.has_parent (null)) {
+                f2_parent = f2_parent.get_parent ();
+                break;
+            }
+            if (!f2_parent.has_parent (null)) {
+                f1_parent = f1_parent.get_parent ();
+                break;
+            }
             f1_parent = f1_parent.get_parent ();
             f2_parent = f2_parent.get_parent ();
```
jeremypw, line 175: Should these be null checked too? These parents are referenced in the loop condition.

jeremypw (changes requested):

src/Utils.vala
aguillon: If both previous `has_parent` return `true`, I assume `get_parent` cannot return `null`, no?

jeremypw: Ah yes, sorry.

jeremypw (approved):
Lets merge this now!

PR #1461, opened 2024-08-20
Add saved files to "recent files". Fix #1281.

### LINKED ISSUE #1281: Editing a file does not update Recent in Files
### What Happened?

A text file (that was not in recent://) was opened, edited and saved with Code.  The file was still not in recent://.  However, when the same file was opened, edited and saved with LibreOffice Writer, it did then appear in recent://.

### Steps to Reproduce

1. Create a text file
2. Open and edit it in Code
3. Save it
4. Open Files and navigate to Recent.  The file does not appear.

### Expected Behavior

Code should behave like other document editors and documents recently changed by them should appear in recent://

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

aguillon:
Hi!

I tried to fix https://github.com/elementary/code/issues/1281 using the [Gtk RecentManager](https://valadoc.org/gtk+-3.0/Gtk.RecentManager.html) but I am only partially satisfied and would like some advice:
* I don't understand why operations on files appear in 3 different files (`Services/Document.vala`, `FolderManager/FileItem.vala` and `FolderManager/File.vala`). I get that renaming a file from the view is a bit different than saving it, and the logic is quite different as well, but shouldn't all of this be refactored together somehow?
* The present code seems to work for "Save" and "Save as" (as well as automatic save) but not renames. Using another program, I have checked that renames were correctly saved in the manager but they don't seem to appear in the recent files, and I don't know why.

I don't know how to fix this for renames. Any advice?


aguillon:
When I say that it "does not work with renames", I mean "adding the same code to the two rename operations in the two other files" does not work, of course.

I just found out about https://github.com/elementary/files/issues/2274. I'll try again with Nautilus tomorrow.

jeremypw:
@aguillon Thanks for looking into this!  At present saving the file is the responsibility of the Document service.  Renaming the file, however is down in the sidebar (FolderManager) in one place only, that is, `Item.rename ()`.  The rest of the code is to do with constructing the context menu and coordinating with the corresponding document to determine whether it is renameable.   So you should be able to add RecentManager code to the Item rename function to get it to work with renames.

I agree it could be simpler to move the rename function to the tab context menu and/or the document context menu.  It seems to be a convention to have the rename option in the sidebar however (at least GEdit does).

jeremypw:
@aguillon Are you still working on this? I have pushed a draft PR that fixes adding to Recent on renaming sidebar items.  If you wish you can copy from this as you were the first to start work on this.

aguillon:
@jeremypw No, you can go ahead and take it. Sorry but I cannot find the time to do Linux development these days.

jeremypw:
Closing in favor of #1467

PR #1465, opened 2024-09-19
Reliably open documents from sidebar

### LINKED ISSUE #1403: Sidebar: First select fails to open document after branch change which removes a sidebar item
### What Happened?

After changing branches (which causes some of the loaded documents to change, selecting another document in the foldermanager failed to open it.

### Steps to Reproduce

1. Open a git project
2. Prepare (if necessary) a branch that contains a file not existing in another branch and switch to that branch
3. Change the branch so that the file is removed from the sidebar
4. Try to open another doc from the sidebar.

An easier way to demonstrate the problem is to create a temporary file item in the sidebar and then delete it.  The next select then fails to open a document.

### Expected Behavior

Selecting a document in the sidebar should open it under these circumstances

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1403 

Using the `item-selected` signal from the SourceList to open documents causes problems when documents are removed for any reason because this causes unwanted documents to selected and opened.   In `master` this is fixed by a hack involving some select signals being ignored but this is not reliable.  Quite often clicking on an item in the sidebar fails to open it because the signal is falsely ignored.

Instead, this PR uses item activation to trigger document opening/tab selection.  To maintain the same single click behaviour the SourceList is modified so `activate_on_single_click` is accessible.


colinkiama (approved):
Works great!

colinkiama:
@elementary/maintainers seems like something is wrong with the CI checks in this pull request.

All the checks have already passed but GitHub insists that there's one more required lint check remaining.

jeremypw:
Seems the linting check has stalled for some reason.

jeremypw:
Hmm, after removing automerge and re-running CI it is now failing on Flatpak build.  It seems to be looking for Gtk.SourceView which we do not use any more, having incorporated the code.

jeremypw:
Looks like the gitlab server is not available atm so SourceView source code not available for building the Flatpak.

jeremypw:
@danirabbit Any idea what is blocking this from merging? CI seems to be passing.  Is it OK to bypass the requirements?

danirabbit:
Sorry about that. Made a mistake migrating to the new branch protection rulesets

PR #1470, opened 2024-10-07
Increase maximum selection length for highlighting

### LINKED ISSUE #1443: Code crashes when deleting lines

### LINKED ISSUE #1469: Certain word selections are not highlighted by plugin
### What Happened?

Some strings of characters do not highlight as expected when selected.

### Steps to Reproduce

1. Open a document and type `Module libzstd.so.1 from deb libzstd-1.5.5+dfsg2-2build1.amd64` at two different points.
2. Ensure word selection highlighting plugin is active
3. Progressively select from the first occurrence - the corresponding words in the second occurrence should highlight.
4. At the point the "+" character is included in the selection, the highlighting fails.

### Expected Behavior

The selected string should be highlighted when it occurs elsewhere in the document regardless of its contents.

### OS Version

7.x (Horus)

### Session Type

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1469 

The previous limit on selection highlighting length was based on selecting only a single word but the plugin actually works on multi-word selections so the limit of less than one line (potentially) does not make a lot of sense.  I have increased it to 255 for now.


zeebok:
Hmmm maybe I misunderstood the issue. The later example in #1443 involves highlighting just a word and causing the crash. I am not sure if I have run into an issue with multi-line deletion issue, or at least don't run into either issue on OS8 stable :frowning_face:

jeremypw:
I hadn't seen that last example in #1443.  I've never seem a crash with just one word selected so I am not sure whether this PR fixes that.  #1443 is sepecifically about multi-line selections anyway.

An alternative approach to this would be to limit selection highlighting to just one word (which the original length limit implies was the original intention) - that would automatically exclude multiple lines.

zeebok:
> I hadn't seen that last example in #1443. I've never seem a crash with just one word selected so I am not sure whether this PR fixes that. #1443 is specifically about multi-line selections anyway. I can reproduce that issue every time.

At the moment I have the opposite luck with repro lol. I can always get the single word crash (with the highlight plugin on) but not a multiline.... is it pretty consistent with a certain doc that I can peek at?

jeremypw:
@zeebok Thanks for looking into this - its strange that you get different symptoms!  If you take a look at the comments to https://github.com/elementary/code/pull/1447, there is a text file there that I posted on July 27 for Danielle to use to reproduce.  I may have to reduce the scope of this to fixing #1469 if it does not fix the other issue for everyone.

jeremypw:
I don't think it relies on a particular document - I never get a crash with single word deletion of a selected, highlighted word (with BackSpace) whereas as you and others do.  It must be some obscure race that depends on details of the particular machine being used.

jeremypw:
@zeebok I've just noticed that after trying to reproduce the single word crash,  the document no longer responds to keyboard input, although Code overall responds as normal - is this the issue you have?

zeebok (approved):
This at the very least fixes the issue in the description, so we should just merge it in

PR #1472, opened 2024-10-09
Store and restore active-project-path explicitly

### LINKED ISSUE #1471: Active project dropdown sometimes out of sync on restart
### What Happened?

On reloading it was sometimes found that the active project as indicated by dropdown was different to the active project indicated by the open documents, expanded sidebar folder and terminal widget

### Steps to Reproduce

1. Open two projects
2. Open a document from each project
3. Open the Terminal widget
4. Set as active the project that appears second in the sidebar
5. Close and reopen Code
6. Observe that the sidebar, documents and terminal are restored as expected but the dropdown shows the first project as active, not the second.

### Expected Behavior

On restoring the dropdown should correctly show the active project.

### OS Version

7.x (Horus)

### Session Type

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1471 

 - [x] Save and restore the active (git) project
 - [x] Ensure the project chooser button and the terminal show the active project
 - [x] Active project only changes in response to user choice 
 - [x] Global search continues to search in the project associated with the focussed document  
 - [x] When other folders are closed with the context menu then the remaining folder becomes the active one
 - [x] The menu option "Open in Terminal Pane" becomes "Set as Active Project" for git repos.   

See also #1518 - "Open in Terminal Pane" is a distinct action from "Set as Active Project" - it just changes the current directory in the Terminal Pane to the folder path whereas "Set as Active Project" has a number of effects including setting the current Terminal Pane directory to the default *build* path (not necessarily the folder (root) path).


danirabbit (dismissed):
I think this is actually worse because the project isn't updating when I switch project files, so project-wide search will continue to search in the old selection instead of the project that relates to the open files

jeremypw:
@danirabbit Hmm, OK. I was trying to make the "active project" more stable and avoid conflict between what the terminal build directory is and what the active project is.  May have to rethink the global search.  It is not clear what to do if the user opens documents in more than one project at a time. I guess the search should be done in the project associated with the currently focussed document?  I am reluctant to have the active project keep switching depending on which document is focussed!  The active project, in my mind, is the one that is being developed and built. Agreed you might want to have documents from other projects open for reference but that does not change the focus of development.

jeremypw:
Converting to draft while fixing a bug in choosing active projects with hyphens

jeremypw:
@danirabbit I would be grateful if you would reconsider the revised version of this PR. The project-wide search now follows the project associated with the focussed document as before, which I believe was you main concern.  I have some uncertainty about what to do with "project" folders that are not git repos but this is probably not blocking as it pre-exists.  Do we need to distinguish these as far as setting the terminal path or project button for example.  At present in this PR the context menu for the project folder does differ but it would be simpler if it didn't so I am happy to revert that.

src/FolderManager/FileView.vala
```diff
@@ -119,10 +116,30 @@ public class Scratch.FolderManager.FileView : Code.Widgets.SourceList, Code.Pane
...
         }
 
+        //Make remaining project the active one
+        git_manager.active_project_path = path;
+
+        write_settings ();
+    }
+
+    private void action_set_active_project (SimpleAction action, GLib.Variant? parameter) {
+    warning ("set active project");
```
zeebok, line 130:
```suggestion
        warning ("set active project");
```

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -134,16 +134,28 @@ namespace Scratch.FolderManager {
...
             }
 
-            var open_in_terminal_pane_item = new GLib.MenuItem (
-                _("Open in Terminal Pane"),
-                GLib.Action.print_detailed_name (
-                    MainWindow.ACTION_PREFIX + MainWindow.ACTION_OPEN_IN_TERMINAL,
-                    new Variant.string (
-                        Services.GitManager.get_instance ().get_default_build_dir (path)
+            MenuItem set_active_folder_item;
+            if (is_git_repo) {
```
zeebok, line 138: Do we want to allow both menu items since we don't provide git actions within Code?

src/MainWindow.vala
```diff
@@ -1200,8 +1210,10 @@ namespace Scratch {
         }
 
         private void action_find_global (SimpleAction action, Variant? param) {
+            var selected_text = "";
+            var search_path = "";
+
             var current_doc = get_current_document ();
-            string selected_text = "";
```
zeebok, line 1204: Why change this from strongly typed to loose?

zeebok (approved):
Looks good. Just a couple questions due to my ignorance of the codebase. I can test if you like before merge but I'm comfortable with just merging.

src/MainWindow.vala
jeremypw: It is elementaryos code-style to use `var` or `unowned var` when the type is implied. It's just neater I suppose since the variable names align. I don't think it makes any difference to the C code in this case.

src/FolderManager/FileView.vala
jeremypw: Removed

src/FolderManager/ProjectFolderItem.vala
zeebok: Maybe I am just misunderstanding where this menu comes in. What I am reading with this code is if I right click a project in the Project (left) Pane I will only see the option to open in terminal if the project is not a git repo? My comment off of that thinking is that even as a git repo I would want to open terminal since we don't provide git operations beyond the current branch stuff.

But it sounds like I am just missing something with this code.

jeremypw: Ah, setting a project as the active project also sets the Terminal path to the build directory (set in Preferences).  For a non-git folder the Terminal is just set to the path of that folder.

zeebok (approved):

PR #1484, opened 2024-10-22
Fix highlighting after reselection of same word

### LINKED ISSUE #1483: Highlight word selection:  Does not work when reselecting the same word
### What Happened?

Selected a word - others were highlighted.  Then clicked elsewhere in the doc so the selection (and highlighting) was lost.  Reselected the same word - no highlighting occurred.  Selected a different word - highlighting occurred.

### Steps to Reproduce

1. Open a doc with multiple occurrences of a word
2. Turn on word highlighting plugin
3. Select target word
4. Deselect by clicking elsewhere
5. Reselect target word
6. Observe that highlighting no longer occurs.

### Expected Behavior

Highlighting of other occurrences of selected word happens regardless of past actions.

### OS Version

7.x (Horus)

### Session Type

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1483 

 - Ensure selection is updated in plugin when selection changes by clicking elsewhere
 - Simplify and DRY some code

Throttling the update-spaces function was tried but the resulting lag was found to be distracting so now they are updated straight-away as before. This is inefficient but any improvement can be left for another PR.


zeebok (dismissed):
This works as expected though it seems slightly slower, particularly with removing the highlight after clicking elsewhere. Not sure if there is a way to fix that before merging but I think it could also be a separate PR if you wanted

jeremypw:
@zeebok I see what you mean.  I have now separated the throttle for the selection change signal so that can be faster.

zeebok:
@jeremypw I will try to take a look at this review again this week!

jeremypw:
Converting to draft as there seems to be some code from the dogfood fork accidentally merged here!

PR #1488, opened 2024-11-25
Silence some warnings

jeremypw:
Silence some critical warnings on startup and on renaming sidebar items

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -618,7 +618,10 @@ namespace Scratch {
                     title = _("%s - %s").printf (doc.get_basename (), base_title);
 
                     toolbar.set_document_focus (doc);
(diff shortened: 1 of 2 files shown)


src/MainWindow.vala
```diff
@@ -601,7 +601,7 @@ namespace Scratch {
                 content_stack.visible_child = view_grid;
                 toolbar.document_available (true);
                 set_widgets_sensitive (true);
-                update_find_actions ();
+                update_find_actions ();jeremypw/silence-some-warnings
```
zeebok, line 604: Should that be there?

jeremypw: Oops! Fixed.

leonardo-lemos (approved):

PR #1489, opened 2024-11-25
Jeremypw/minor code improvements

jeremypw:


src/FolderManager/File.vala
```diff
@@ -78,31 +83,29 @@ namespace Scratch.FolderManager {
...
-            }
+        // If parent is hidden then inherit validity from parent
+        private bool? _is_valid_directory = null;
+        public bool is_valid_directory {
+            get {
+                if (_is_valid_directory == null) {
+                _is_valid_directory = info != null &&
+                                      !info.get_is_backup () &&
+                                      info.get_file_type () == FileType.DIRECTORY;
+                }
```
ryonakano, line 94:
```suggestion
                if (_is_valid_directory == null) {
                    _is_valid_directory = info != null &&
                                          !info.get_is_backup () &&
                                          info.get_file_type () == FileType.DIRECTORY;
                }
```

Because we have 4 spaces indentation anywhere we use the if statement

src/FolderManager/FolderItem.vala
```diff
@@ -304,15 +302,14 @@ namespace Scratch.FolderManager {
                         var path_item = find_item_for_path (source.get_path ());
                         if (path_item == null) {
                             var file = new File (source.get_path ());
-                            if (file.is_valid_directory ()) {
+                            if (file.is_valid_directory) {
                                 path_item = new FolderItem (file, view);
-                            } else if (!file.is_temporary) {
+
```
ryonakano, line 307:
```suggestion
```

Maybe an extra line added unintentionally?

src/FolderManager/File.vala
jeremypw: Done

src/FolderManager/FolderItem.vala
jeremypw: Removed

src/FolderManager/File.vala
```diff
@@ -67,8 +67,13 @@ namespace Scratch.FolderManager {
...
 
-                _icon = GLib.ContentType.get_icon (info.get_content_type ());
-                return _icon;
+                if (info != null) {
+                    _icon = GLib.ContentType.get_icon (info.get_content_type ());
+                    return _icon;
+                } else {
+                    _icon = new ThemedIcon ("missing-image");
+                    return _icon;
+                }
```
ryonakano, line 76: Tiny thing, but we don't need to write return statement twice here.

```suggestion
                if (info != null) {
                    _icon = GLib.ContentType.get_icon (info.get_content_type ());
                } else {
                    _icon = new ThemedIcon ("missing-image");
                }

                return _icon;
```

ryonakano (commented):
Sorry for my late review, another tiny thing and otherwise looks good.

jeremypw:
@ryonakano  Thanks for review! I've fixed that obviously unnecessary repetition.

ryonakano (approved):

PR #1487, opened 2024-11-24
851 fix word completion logic

jeremypw:
Fixes #851 

 - [ ] Handle text being inserted at the start of a word
 - [ ] Handle [delete_range](https://valadoc.org/gtk+-3.0/Gtk.TextBuffer.delete_range.html) signal, removing unused words from the prefix tree
 - [ ] Allow numeric delimiters as long as they aren't at the start of a word (allows variables with numbers in them to be suggested)

PR #1490, opened 2024-12-02
Rework word completion: Use Gee.TreeMap instead of custom prefix tree

jeremypw:
An experiment arising from #1487, an attempt to address some of the shortcomings of the current plugin. Much of the extra code is commentary.  Also, this PR works entirely with `unichar` and the builtin definition of DELIMITERS rather than sometimes using TextIter functions to find word boundaries and sometimes using `split_set` which do not necessarily agree.

Not sure how this will scale up to very large documents but Code is intended for code files which are generally modest in size.

Pushing for visibility but will probably make a new cleaner PR from scratch judged an improvement.


jeremypw:
Closing as although it is almost working, it turned out not to have material advantages.

PR #1496, opened 2024-12-21
doc: bump libvala dependency

kaixoo:
OS 8 doesn't seem to have `libvala-0.48-dev` anymore

```bash

user@pc:~$ sudo apt install libvala-0.48-dev
Lecture des listes de paquets... Fait
Construction de l'arbre des dépendances... Fait
Lecture des informations d'état... Fait      
E: Impossible de trouver le paquet libvala-0.48-dev
E: Impossible de trouver de paquet correspondant à l'expression rationnelle « libvala-0.48-dev »
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ You'll need the following dependencies:
 * libhandy-1-dev >= 0.90.0
 * libpeas-dev
 * libsoup2.4-dev
-* libvala-0.48-dev (or higher)
+* libvala-0.56-dev (or higher)
 * libvte-2.91-dev
 * valac
 


jeremypw (commented):
Should be OK as OS7 has v0.56.  The libvala dependency automatically adjusts to the version of vala compiler installed so we should probably specify the same minimum version for `valac`

kaixoo:
> so we should probably specify the same minimum version for valac


I didn't want to touch the minimum version. Just adjusting the *recommended* version because 0.48 no longer exists in OS 8.

jeremypw:
I don't know what the true minimum version of vala required to compile the code is now tbh. But we only *support* compiling on the current release of elementary (8) and the previous one (7) anyway so we generally bump the dependencies accordingly.

PR #1500, opened 2024-12-24
Do not allow simultaneous opening of projects with parent-child relationship

### LINKED ISSUE #1494: Code crashes when both a folder and its parent are open as projects and a document present in both is opened
### What Happened?

Code crashed when opening a document in a parent project when the child folder was also open as a project.

### Steps to Reproduce

1. Close all docs and projects
2. Open a folder as a project
3. Open the parent of that folder as a project
4. Navigate in the sidebar to a document present in the child folder and click to open it.
5. Code crashes

### Expected Behavior

Code should not crash - the document should open.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1494 

 - Check parents and children before opening a new project folder
 - Fix consequent race condition between restoring project folders and restoring open documents

Because restoring a project folder now occurs in a separate thread it is necessary to ensure that all project folders are restored before starting to restore open documents otherwise documents are not correctly assigned to a project on startup.


zeebok (approved):
Looks good to me!

PR #1501, opened 2024-12-25
Migrate to libpeas-2

### LINKED ISSUE #1498: Refactor the plugin system to not use libpeas
### Problem

See https://github.com/elementary/code/pull/1497#issuecomment-2558162725

### Proposal

For ease of porting to Gtk4 and to remove an unnecessary dependency we should implement the plugin system in the same way that Files does

### Prior Art (Optional)

See the relevant code in Pantheon Files.

In Code it will be necessary to implement activating and deactivating individual plugins and adjust the PreferencesDialog accordingly as in Files there are only core plugins that are always active.

jeremypw:
Fixes #1498 

 - Reproduces as far as possible changes to current appearance and behaviour
 - Needed to ship a `.vapi` for libpeas-2 for now as is not currently supplied by the libpeas-2-dev package available (2.0.0) but it is supplied by 2.0.5 which is published in Ubuntu 24.10 (Oracular) and later. Not sure how to get it into elementary8 which is based on Ubuntu 24.04LTS.
 - Could not get Flatpak manifest to compile because of dependency issues so leaving that for now. Code does not currently work properly as a Flatpak anyway and is supplied as `.deb`

Used a ListBox bound to the Peas.Engine list_model for the Preferences Dialog, but as the model does not change while the ListBox in use may be simpler to construct and sort it manually.


kaixoo:
> Needed to ship a .vapi for libpeas-2 as is not currently supplied by the libpeas-2-dev package available

Can we send this upstream?

jeremypw:
> Can we send this upstream?

Probably, it should just be a matter of deb packaging.  However, I am not convinced libpeas-2 is the way to go - it provides a lot of stuff we don't need like loading JavaScript plugins which has to be included in the Flatpak (unless there are some compile options I haven't found yet).  I am going to try to produce another PR removing libpeas altogether for comparison.

jeremypw:
Libpeas may provide small advantages in code organization, but as far as I am aware the main point is to allow third-party plugins written in other languages and separately packaged in other projects to be used (they just need to be installed into the right directory).

kaixoo:
I don't see a problem in allowing to load Javascript plugins, but okay.

jeremypw:
Its not a problem, just unnecessary.  Need to identify functional advantages of libpeas to justify the extra dependencies.

jeremypw:
@danirabbit Will it be possible to ship libpeas-2 version 2.0.5 in elementary 8 if it is not in Ubuntu 24.04?

jeremypw:
The main advantage of keeping the libpeas dependency for now is, especially if we can lose the need for the `.vapi`, is the smaller diff.

jeremypw:
Hmm - as far as I can make out,, even Gnome is not using `libpeas-2` for their `gedit` flatpak :disappointed:

jeremypw:
@leonardo-lemos Thanks!  I forgot to check the CI file as everything worked locally.

kaixoo:
@jeremypw if you need help in anything you can tell me

jeremypw:
@kaixoo Thanks for the offer. The limiting factor is getting reviews and re-reviews of existing (simpler, ready) PRs tbh.  I guess people are focussed on finalizing the stable release of OS 8.  I could mark this as ready for review if you like, otherwise I was waiting for the `.vapi` file to be included in the `-dev` package.

I've started a fork of Code that I can use myself at least that contains some of the more useful features that are not in the original.

kaixoo:
Ah, alright, I guess it's fine to wait then.

leonardo-lemos:
@jeremypw Can you resolve the conflicts here, please? I'd like to take a look at this PR ;)

leonardo-lemos (changes requested):
Works as expected, but we need to update dependencies in the README too.

leonardo-lemos (approved):
LGTM!

PR #1502, opened 2024-12-27
Lose libpeas dependency

### LINKED ISSUE #1498: Refactor the plugin system to not use libpeas

jeremypw:
Fix #1498

 - Use `gmodule-2.0` directly
 - Implement only the necessary parts of `libpeas` in `PluginManager` (based on the equivalent class in `io.elementary.files`)
 - Update code style a bit, add some headers
 - Move some code into PastbinDialog.vala where it is actually used
 
NOTE:  All the plugins are loaded and made resident in the same way as Files. They are activated and deactivated (but not unloaded) as required.  When deactivated the interface signals are disconnected.

Need to check all plugins still work as expected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Install Dependencies
       run: |
         apt update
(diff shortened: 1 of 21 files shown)


jeremypw:
Closing in favor of migrating to libpeas-2 for now.

PR #1503, opened 2025-01-08
Always reselect symbol item when clicked

### LINKED ISSUE #1317: Symbol Pane: Does not scroll back to already selected method
### What Happened?

A certain method was located with the symbol pane - the selected method scrolled into view.  The view was then manually scrolled so the method was out of view.  That method was again clicked in the symbol pane - the method did not scroll back into view.

### Steps to Reproduce

 * Open symbol pane and a (long) Vala file
 * Select a method in the symbol pane
 * View scrolls the method into view
 * Manually scroll view away from that method
 * Reselect that method in the symbol pane
 * View does not scroll

### Expected Behavior

Clicking a method in the symbol pane should always scroll that method into view even if it was the previously clicked method.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1317

### DIFF
--- a/data/io.elementary.code.gschema.xml
+++ b/data/io.elementary.code.gschema.xml
@@ -152,6 +152,11 @@
       <summary>Remember the last focused document.</summary>
       <description>Restore the focused document from a previous session when opening Code.</description>
     </key>
+    <key name="active-project-path" type="s">
+      <default>''</default>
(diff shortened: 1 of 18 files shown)


jeremypw:
Closing as based on wrong branch

PR #1510, opened 2025-01-19
Port to Gtk4 and Gtk.SourceView5

jeremypw:


.github/workflows/ci.yml
```diff
@@ -22,7 +22,8 @@ jobs:
     - name: Install Dependencies
       run: |
         apt update
-        apt install -y exuberant-ctags libeditorconfig-dev libgail-3-dev libgee-0.8-dev libgranite-7-dev libgtk-4-dev libgtksourceview-5-dev libgtkspell3-3-dev libadwaita-1-dev libpeas-2-dev libsoup-3.0-dev libvala-dev libvte-2.91-dev meson valac polkitd libpolkit-gobject-1-dev
+        apt install -y exuberant-ctags libeditorconfig-dev libgail-3-dev libgee-0.8-dev libgranite-7-dev libgtk-4-dev libgtksourceview-5-dev libgtkspell3-3-dev libadwaita-1-dev libpeas-2-dev libsoup2.4-dev libvala-dev libvte-2.91-gtk4-dev meson valac polkitd libpolkit-gobject-1-dev libgit2-glib-1.0-dev
```
zeebok, line 25: Why the rollback on libsoup version?

plugins/pastebin/meson.build
```diff
@@ -5,7 +5,7 @@ module_files = [
     'pastebin.vala'
 ]
 
-soup_dep = dependency('libsoup-2.4')
+soup_dep = dependency('libsoup-3.0')
```
zeebok, line 8: This is updated to 3 but the CI workflow installs 2.4?

.github/workflows/ci.yml
jeremypw: Can't remember now! I'll look at this again.

plugins/pastebin/meson.build
jeremypw: Probably a mistake in CI but I'll check.

jeremypw: I've reverted to libsoup-3

jeremypw:
@zeebok Thanks for the review. Its a while since I've thought about this. I've corrected that libsoup dependency in CI. I am not sure why this PR is not in draft as there is a lot of migration still to be done on it. I fear the diff is already too big anyway and changes need to be backported to the Gtk3 version where possible first.

jeremypw:
> Does this mean that you plan on creating a separate `gtk4` or `next` branch that we can gradually push gtk4 changes to before finally merging to `main`?

That would seem easier but it really depends on what @danirabbit wants to do.

jeremypw:
@danirabbit The ports done so far have been relatively small.  I suspect the diff for Files will still be fairly large even when as much prep work as possible has been done.  I am not sure what a "slop" branch is but if you can identify areas of prep work that can still be done in Files I am happy to do them.

jeremypw:
@danirabbit I think Imost of of the event controller stuff I'll check what menu stuff still needs doing.  Do I understand correctly that when all that can be done in Gtk3 has been done (presumably you will determine that?), then you envisage a single PR that converts all remaining to functional Gtk4?  If so, then I guess each commit to that needs to be logically separate like a mini-PR.  I could try squash merging locally from local sub-branches to facilitate that.

jeremypw:
Closing as now too many conflicts and too old.  Will be pushing a fresh attempt in the near future.

PR #1512, opened 2025-01-20
Fix "Open in" submenu when running as Flatpak

### LINKED ISSUE #1511: Open in menu empty when running as Flatpak
### What Happened?

Running Code as Flatpak, I notice that the context menu of Project Items in the sidebar has an empty submenu under "Open in..."

### Steps to Reproduce

 - Run Code as a Flatpak
 - Open a project
 - Secondary click on the sidebar item for the project
 - Choose "Open in..."
 - Note missing apps

### Expected Behavior

The submenu should contain something usable or not appear at all.

### OS Version

7.x (Horus)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #1511


zeebok (approved):
I really wish we could say what app it would open in, and I am guessing we have no way to find that out, so this works well!

zeebok:
Just gotta fix the conflict

jeremypw:
@zeebok Thanks for the review! Yes, it would be better get the default app name if possible but I could not see how to do it in a distro-agnostic way and without perforating the sandbox. There doesn't seem to be a portal for this information.  If I've missed something it can be improved later.

PR #1519, opened 2025-01-24
Sort Apps for "Open in ..."

jeremypw:
Rather than listing applicable apps in indeterminate order, they are shown in locale alphabetical order.

Screenshot from 2025-01-29 12-28-30

### DIFF
--- a/src/Utils.vala
+++ b/src/Utils.vala
@@ -202,12 +202,13 @@ namespace Scratch.Utils {
     }
 
     public GLib.Menu create_executable_app_items_for_file (GLib.File file, string file_type) {
-        List<AppInfo> external_apps = GLib.AppInfo.get_all_for_type (file_type);
-        var files_appinfo = AppInfo.get_default_for_type ("inode/directory", true);
-        external_apps.prepend (files_appinfo);
-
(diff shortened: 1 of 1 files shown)


src/Utils.vala
```diff
@@ -202,13 +202,14 @@ namespace Scratch.Utils {
...
-        var files_appinfo = AppInfo.get_default_for_type ("inode/directory", true);
-        external_apps.prepend (files_appinfo);
-
-        string this_id = GLib.Application.get_default ().application_id + ".desktop";
+        var external_apps = GLib.AppInfo.get_all_for_type (file_type);
+        var this_id = GLib.Application.get_default ().application_id + ".desktop";
         var menu = new GLib.Menu ();
 
+        external_apps.sort ((a, b) => {
+            return strcmp (a.get_name (), b.get_name ());
```
ryonakano, line 210: Maybe we can use [`string.collate ()`](https://valadoc.org/glib-2.0/string.collate.html) here, which looks like follow the current locale?

jeremypw: Good point - thanks!  Done.

ryonakano (approved):

PR #1520, opened 2025-01-24
Fix highlighting after reselecting same word

jeremypw:


jeremypw:
Blocked until libpeas-2 fix has been merged as plugin not working otherwise.

PR #1491, opened 2024-12-04
Dynamically update word completions

jeremypw:
Fixes #851 

 - [x] Plugin: Handle 'insert-text' and 'delete-range' signals to update prefix-tree before buffer changes
 - [x] Engine: Only use `split-set` with custom delimiters to identify words (do not use textiter functions) see #1526
 - [x] PrefixNode: Split out into separate file
 - [x] PrefixNode: Add functionality to record multiple occurrences of word.
 - [x] PrefixNode: Add link to parent node
 - [x] PrefixNode: Move code to insert, find, remove and list completions here
 - [x] PrefixTree: Manage reaping of unused words, interface with engine.
 - [x] CompletionProvider: Off-load some functions to parser for consistency
 - [x] CompletionProvider: Completion overwrites remainder of word when inserting in middle 
 - [x] Discontinue key press handling (no need to force tree update?)
 - [x] General: Update code style and headers. Add extra comments and debug output.
 - [x] General: Adjust minimum length of word and minimum length of prefix
 - [x] General: Introduce maximum length of word stored
 - [x] Show completions after deleting with backspace

For simplicity, insertions and deletions are all handled the same way rather than separate functions depending on where the insertion/deletion occurred.  This results in some temporary "words" being added and removed in quick succession while typing but on current average hardware (i5 processor, SSD) and modest sized files (3000 lines) no performance issues were noticed.  However, more efficiency could be achieved at the cost of greater complexity by queueing the changes (like Files) and ignoring those that cancel out. 

To be decided:
 - Best order for completions
 - Best values for min prefix and min word
 - Best value for interactive delay
 - Whether forced update should still be available

However this PR is usable as is in my opinion and can refined later.
 
Based on #1487

PR #1529, opened 2025-01-31
Rework word completion: Split out PrefixNode class

jeremypw:
Towards #1491 

Little value by itself but in order to implement dynamic updating of the tree when text is deleted it is necessary to make this class more complex (e.g. keeps track of number of occurrences of each word) and a lot of code in PrefixTree can be moved here.


danirabbit (approved):

danirabbit:
Makes sense and builds but linter is unhappy :)

PR #1532, opened 2025-02-01
Rework word completion: Start to prepare PrefixNode for dynamic updating

jeremypw:
- Add a property counting how many times a word occurs in a document (used to remove nodes with zero occurrences)
- Distinguish node types: ROOT, CHAR, WORD_END.  (WORD_END type carries occurrences property)
- Make tree bi-directional (nodes have parent property)
- Use Gee.ArrayList instead of GLib.List


plugins/word-completion/prefix-tree-node.vala
```diff
@@ -20,16 +20,51 @@
...
-    public GLib.List<PrefixNode> children;
+    private enum NodeType {
+        ROOT,
+        CHAR,
+        WORD_END
+    }
+
+    private const unichar WORD_END_CHAR = '\0';
+    private unichar? uc = null;
+    private NodeType type = ROOT;
```
danirabbit, line 31: Seems like this should also be a construction property since it's set in constructors?

plugins/word-completion/prefix-tree-node.vala
```diff
@@ -20,16 +20,51 @@
...
 public class Scratch.Plugins.PrefixNode : Object {
-    public GLib.List<PrefixNode> children;
+    private enum NodeType {
+        ROOT,
+        CHAR,
+        WORD_END
+    }
+
+    private const unichar WORD_END_CHAR = '\0';
+    private unichar? uc = null;
```
danirabbit, line 30: Should this not be a construction property since it's set in constructors?

jeremypw: See above reply

plugins/word-completion/prefix-tree-node.vala
jeremypw: Have to jump through some hoops to get it to compile as a property which did not seem worthwhile for a private member but I guess it does ensure that it is immutable I suppose.  I'll push a commit for review.

jeremypw:
Reviewed scope and construction of properties. Type and character made immutable properties; occurrence count made private member.

danirabbit (changes requested):

plugins/word-completion/prefix-tree-node.vala
```diff
@@ -20,16 +20,50 @@
...
 public class Scratch.Plugins.PrefixNode : Object {
-    public GLib.List<PrefixNode> children;
+    public enum NodeType {
+        ROOT,
+        CHAR,
+        WORD_END
+    }
+
+    private const unichar WORD_END_CHAR = '\0';
+    private  uint occurrences; // Only used for WORD_END nodes
```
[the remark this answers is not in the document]
jeremypw: Sorry about that - idk how I keep missing lint.

danirabbit (approved):

PR #1537, opened 2025-02-07
Handle error thrown while checking out a branch

jeremypw:
- Silence warning about unhandled GLib.Error
 - Provide feedback if checkout fails


src/Services/MonitoredRepository.vala
```diff
@@ -191,28 +191,41 @@ namespace Scratch.Services {
...
+                    //Ensure documents match checked out branch (deal with potential conflicts/losses beforehand)
+                    strategy = Ggit.CheckoutStrategy.FORCE
+                };
 
-            git_repo.checkout_head (options);
+                git_repo.checkout_head (options);
+                branch_name = new_branch_name;
+            } catch (Error e) {
+                var dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                    _("An error occurred while checking out %s").printf (new_branch_name),
```
ryonakano, line 213: I guess `new_branch_name` could be a blank string if `new_branch_name = new_head_branch.get_name ();` in L197 throws an error, which the error message results an incomplete sentence: `An error occurred while checking out `

FYI Looking at `vapi/libgit2-glib-1.0.vapi`, `Ggit.Branch.get_name ()` could throw an error:

```c
	[CCode (cheader_filename = "libgit2-glib/ggit.h", type_id = "ggit_branch_get_type ()")]
	public class Branch : Ggit.Ref {
		[CCode (has_construct_function = false)]
		protected Branch ();
		public void @delete () throws GLib.Error;
		public unowned string? get_name () throws GLib.Error;
```

jeremypw: I think it unlikely because the branch has already been looked-up by name in the calling function.  However I have modified the message to not use the name explicitly (it may occur in the error's own message anyway)

jeremypw:
Thanks for the review @ryonakano . I have changed the message in the light of your concerns.

ryonakano (approved):
LGTM, thank you!

PR #1539, opened 2025-02-13
Profiling: Enable profiling in code using google-perftools

jeremypw:
Allow obtaining cpu and/or heap profile on specified parts of the code using tools from google-perftools package. The libraries are optionally linked in according to the `meson_option.txt` setting.  The profilers can be switched on and off in the desired section of code using conditional compilation flags.  Some basic instructions are included in the code.

Profiling the whole app can be done from the commandline without compiling in the libraries but the advantage of having the libraries compiled in is that profiling can be switched on and off in specific parts of the code only. 

By default the libraries are not compiled in so these changes will have no effect on the 'ordinary' user.


leonardo-lemos:
@jeremypw Can we merge it?

zeebok:
@jeremypw I think it is probably safe to merge this now

PR #1545, opened 2025-03-03
Bump platform to 8.1

danirabbit:


ryonakano:
I fixed Code builds and runs successfully with Flatpak both on amd64 and arm64. Also bumped platform from 8.1 to 8.2.

jeremypw (approved):
Builds and runs successfully.

PR #1548, opened 2025-04-03
Clone remote repository.

jeremypw:
- Add clone action and associated menuitem
 - Put "Clone" menuitem in ProjectChooser menu
 - Move "Open Project" menuitem into ProjectChooser menu
 - Use dialog to get remote uri and local target (partially validated)
 - Persist default remote and local projects folder entries
 - Show spinner while cloning is ongoing, do not block UI
 - If cloning fails show error dialog giving opportunity to amend and retry
 - On success, open the new project folder but do not make it active (similar to "Open Project" action)
 
Screenshot from 2025-06-05 12 26 58
Screenshot from 2025-06-05 12 28 23
Screenshot from 2025-06-05 12 29 28

### DIFF
--- a/src/Dialogs/CloneRepositoryDialog.vala
+++ b/src/Dialogs/CloneRepositoryDialog.vala
@@ -6,16 +6,13 @@
  */
 
 public class Scratch.Dialogs.CloneRepositoryDialog : Granite.MessageDialog {
-    public bool can_clone { get; private set; default = false; }
-
-
     // Git project name rules according to GitLab
     // - Must start and end with a letter ( a-zA-Z ) or digit ( 0-9 ).
     // - Can contain only letters ( a-zA-Z ), digits ( 0-9 ), underscores ( _ ), dots ( . ), or dashes ( - ).
     // - Must not contain consecutive special characters.
     // - Cannot end in . git or . atom .
-
     private const string NAME_REGEX = """^[0-9a-zA-Z].([-_.]?[0-9a-zA-Z])*$"""; //TODO additional validation required
+
     private Regex name_regex;
     private Gtk.Label projects_folder_label;
     private Granite.ValidatedEntry remote_repository_uri_entry;
@@ -25,6 +22,7 @@ public class Scratch.Dialogs.CloneRepositoryDialog : Granite.MessageDialog {
     private Gtk.Spinner spinner;
     private Gtk.Revealer revealer;
 
+    public bool can_clone { get; private set; default = false; }
     public string suggested_local_folder { get; construct; }
     public string suggested_remote { get; construct; }
 
@@ -51,7 +49,16 @@ public class Scratch.Dialogs.CloneRepositoryDialog : Granite.MessageDialog {
     construct {
         transient_for = ((Gtk.Application)(GLib.Application.get_default ())).get_active_window ();
         image_icon = new ThemedIcon ("git");
+        badge_icon = new ThemedIcon ("emblem-downloads");
         modal = true;
+
+        ///TRANSLATORS "Git" is a proper name and must not be translated
+        primary_text = _("Create a local clone of a Git repository");
+        secondary_text = _("The source repository and local folder must exist and have the required read and write permissions");
+
+        var cancel_button = add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
+        clone_button = (Gtk.Button)add_button (_("Clone Repository"), Gtk.ResponseType.APPLY);
+        set_default (clone_button);
 
         try {
             name_regex = new Regex (NAME_REGEX, OPTIMIZE, ANCHORED | NOTEMPTY);
@@ -59,16 +66,10 @@ public class Scratch.Dialogs.CloneRepositoryDialog : Granite.MessageDialog {
             warning ("%s\n", e.message);
         }
 
-        var cancel_button = add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
-
-        ///TRANSLATORS "Git" is a proper name and must not be translated
-        primary_text = _("Create a local clone of a Git repository");
-        secondary_text = _("The source repository and local folder must exist and have the required read and write permissions");
-        badge_icon = new ThemedIcon ("emblem-downloads");
-
         remote_repository_uri_entry = new Granite.ValidatedEntry () {
             placeholder_text = _("https://example.com/username/projectname.git"),
-            input_purpose = URL
+            input_purpose = URL,
+            activates_default = true
         };
         remote_repository_uri_entry.changed.connect (on_remote_uri_changed);
         remote_repository_uri_entry.text = suggested_remote;
@@ -109,7 +110,9 @@ public class Scratch.Dialogs.CloneRepositoryDialog : Granite.MessageDialog {
 
         });
 
-        local_project_name_entry = new Granite.ValidatedEntry ();
+        local_project_name_entry = new Granite.ValidatedEntry () {
+            activates_default = true
+        };
         local_project_name_entry.changed.connect (validate_local_name);
 
         var content_box = new Gtk.Grid ();
@@ -135,12 +138,9 @@ public class Scratch.Dialogs.CloneRepositoryDialog : Granite.MessageDialog {
         custom_bin.add (stack);
         custom_bin.show_all ();
 
-        clone_button = (Gtk.Button)add_button (_("Clone Repository"), Gtk.ResponseType.APPLY);
         bind_property ("can-clone", clone_button, "sensitive", DEFAULT | SYNC_CREATE);
         spinner.bind_property ("active", clone_button, "visible", INVERT_BOOLEAN);
         spinner.bind_property ("active", cancel_button, "visible", INVERT_BOOLEAN);
-        //Do not want to connect to "is-valid" property notification as this gets changed to "true" every time the entry
-        //text changed. So call explicitly after we validate the text.
         can_clone = false;
 
         // Focus cancel button so that entry placeholder text shows


src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+        repository_host_uri_entry = new Granite.ValidatedEntry.from_regex (url_regex) {
+            activates_default = false
+        };
+        repository_user_entry = new Granite.ValidatedEntry.from_regex (name_regex) {
+            activates_default = false
+        };
+        repository_name_entry = new Granite.ValidatedEntry.from_regex (name_regex) {
+            activates_default = false
+        };
+        repository_local_folder_entry = new Granite.ValidatedEntry.from_regex (local_folder_regex) {
```
danirabbit, line 80: This should use a file chooser because we need to make sure Code has access in the documents portal to this folder

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+    private void on_is_valid_changed () {
+        can_clone = repository_host_uri_entry.is_valid &&
+                    repository_user_entry.is_valid &&
+                    repository_name_entry.is_valid &&
+                    repository_local_folder_entry.is_valid;
+    }
+
+    private class CloneEntry : Gtk.Box {
+        public CloneEntry (string label_text, Gtk.Widget entry) {
+            var label = new Gtk.Label (label_text) {
```
danirabbit, line 160: Need to make sure to set `mnemonic_widget` property so that entries are labeled for the screen reader

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+            image_icon: new ThemedIcon ("git")
+        );
+
+        repository_local_folder_entry.text = local_folder;
+        repository_host_uri_entry.text = "https://github.com";
+        repository_user_entry.text = "elementary";
+        repository_name_entry.text = "";
+        repository_local_name_entry.text = "";
+
+        on_is_valid_changed ();
```
danirabbit, line 53: This should all happen in the `construct` block. Only things necessary for the `Object` call should happen in a constructor

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+
+    public CloneRepositoryDialog (string local_folder) {
+        Object (
+            transient_for: ((Gtk.Application)(GLib.Application.get_default ())).get_active_window (),
+            image_icon: new ThemedIcon ("git")
+        );
+
+        repository_local_folder_entry.text = local_folder;
+        repository_host_uri_entry.text = "https://github.com";
+        repository_user_entry.text = "elementary";
```
danirabbit, line 49: This should be placeholder text, not pre-filled text

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+        repository_name_entry.text = "";
+        repository_local_name_entry.text = "";
```
danirabbit, line 51: Text is empty by default. I'm not sure why you're setting it empty again?

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+        clone_button.has_default = true;
+        clone_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
+
+        repository_host_uri_entry.notify["is-valid"].connect (on_is_valid_changed);
+        repository_user_entry.notify["is-valid"].connect (on_is_valid_changed);
+        repository_name_entry.notify["is-valid"].connect (on_is_valid_changed);
+        repository_local_folder_entry.notify["is-valid"].connect (on_is_valid_changed);
+
+        clone_button.sensitive = can_clone;
+        bind_property ("can-clone", clone_button, "sensitive");
```
danirabbit, line 116: Can use `SYNC_CREATE` here instead of setting separately

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+        };
+        repository_local_folder_entry = new Granite.ValidatedEntry.from_regex (local_folder_regex) {
+            activates_default = false,
+            width_chars = 50
+        };
+        repository_local_name_entry = new Granite.ValidatedEntry.from_regex (name_regex) {
+            activates_default = false,
+        };
+
+        set_as_active_check = new Gtk.CheckButton.with_label (_("Set as Active Project")) {
```
danirabbit, line 88: Probably need to clarify what this does. Why would/wouldn't someone want to do this?

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+            margin_top = 12,
+            active = true
+        };
+
+        var content_box = new Gtk.Box (VERTICAL, 12);
+        content_box.add (new Granite.HeaderLabel (_("Source Repository")));
+        content_box.add (new CloneEntry (_("Host URI"), repository_host_uri_entry));
+        content_box.add (new CloneEntry (_("User"), repository_user_entry));
+        content_box.add (new CloneEntry (_("Name"), repository_name_entry));
```
danirabbit, line 97: "Name" seems ambiguous here. What is this the name of?

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+        };
+
+        set_as_active_check = new Gtk.CheckButton.with_label (_("Set as Active Project")) {
+            margin_top = 12,
+            active = true
+        };
+
+        var content_box = new Gtk.Box (VERTICAL, 12);
+        content_box.add (new Granite.HeaderLabel (_("Source Repository")));
+        content_box.add (new CloneEntry (_("Host URI"), repository_host_uri_entry));
```
danirabbit, line 95: Is this just https or ssh too?

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+        add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
+        primary_text = _("Create a local clone of a Git repository");
+        ///TRANSLATORS "Git" is a proper name and must not be translated
+        secondary_text = _("The source repository and local folder must exist and be accessible");
+        badge_icon = new ThemedIcon ("download");
+
+        repository_host_uri_entry = new Granite.ValidatedEntry.from_regex (url_regex) {
+            activates_default = false
+        };
+        repository_user_entry = new Granite.ValidatedEntry.from_regex (name_regex) {
```
danirabbit, line 74: can we pre-fill this from Environment? https://valadoc.org/glib-2.0/GLib.Environment.get_user_name.html

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+            activates_default = false
```
danirabbit, line 75: why are these set to `activates_default = false`? Usually you can activate from any part of the form in case you need to go back and make a correction

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+            local_folder_regex = new Regex (LOCAL_FOLDER_REGEX, RegexCompileFlags.OPTIMIZE);
+            url_regex = new Regex (URL_REGEX, RegexCompileFlags.OPTIMIZE);
+        } catch (RegexError e) {
+            warning ("%s\n", e.message);
+        }
+
+        add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
+        primary_text = _("Create a local clone of a Git repository");
+        ///TRANSLATORS "Git" is a proper name and must not be translated
+        secondary_text = _("The source repository and local folder must exist and be accessible");
```
danirabbit, line 68: what does "be accessible" mean in this context?

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+            name_regex = new Regex (NAME_REGEX, RegexCompileFlags.OPTIMIZE);
+            local_folder_regex = new Regex (LOCAL_FOLDER_REGEX, RegexCompileFlags.OPTIMIZE);
+            url_regex = new Regex (URL_REGEX, RegexCompileFlags.OPTIMIZE);
+        } catch (RegexError e) {
+            warning ("%s\n", e.message);
+        }
+
+        add_button (_("Cancel"), Gtk.ResponseType.CANCEL);
+        primary_text = _("Create a local clone of a Git repository");
+        ///TRANSLATORS "Git" is a proper name and must not be translated
```
danirabbit, line 67: This comment is probably meant to be above the preceding line

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+            activates_default = false
+        };
+        repository_name_entry = new Granite.ValidatedEntry.from_regex (name_regex) {
+            activates_default = false
+        };
+        repository_local_folder_entry = new Granite.ValidatedEntry.from_regex (local_folder_regex) {
+            activates_default = false,
+            width_chars = 50
+        };
+        repository_local_name_entry = new Granite.ValidatedEntry.from_regex (name_regex) {
```
danirabbit, line 84: This should probably auto fill when the repository URL is changed

danirabbit (changes requested):
If I'm cloning a repository, I'm almost always either typing the full path or going to this page and hitting the copy button:

Screenshot from 2025-05-27 09 37 44

Screenshot from 2025-05-27 09 38 42

But this form seems to expect different entries for the remote URL and repository path. Is that right? I'm not exactly sure how to fill out the form as written tbh

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,176 @@
...
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+* General Public License for more details.
+*
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA.
+*
+* Authored by: Jeremy Wootten <jeremywootten@gmail.com>
+*/
```
danirabbit, line 20: For new files it would be nice to use the SPDX header:

```suggestion
/*
 * SPDX-License-Identifier: GPL-2.0-or-later
 * SPDX-FileCopyrightText: 2025 elementary, Inc. <https://elementary.io>
 *
 * Authored by: Jeremy Wootten <jeremywootten@gmail.com>
 */
```

src/Widgets/Sidebar.vala
```diff
@@ -60,7 +60,16 @@ public class Code.Sidebar : Gtk.Grid {
...
             action_target = new Variant.string (""),
             always_show_image = true,
-            label = _("Open Folder…")
+            label = _("Open …"),
+            xalign = 0.0f
+        };
+
+        var clone_button = new Gtk.Button.from_icon_name ("folder-open-symbolic", Gtk.IconSize.SMALL_TOOLBAR) {
+            action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_CLONE_REPO,
+            always_show_image = true,
```
danirabbit, line 69: probably shouldn't construct new image buttons in this way since you can't do this in GTK4. You have to put an image and a label in a box and pack that

src/Widgets/Sidebar.vala
```diff
@@ -60,7 +60,16 @@ public class Code.Sidebar : Gtk.Grid {
             action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER,
             action_target = new Variant.string (""),
             always_show_image = true,
-            label = _("Open Folder…")
+            label = _("Open …"),
+            xalign = 0.0f
+        };
+
+        var clone_button = new Gtk.Button.from_icon_name ("folder-open-symbolic", Gtk.IconSize.SMALL_TOOLBAR) {
```
danirabbit, line 67: What do you think about making these model buttons at the end of the project selection popover?

jeremypw:
@danirabbit Thanks for the detailed review! I'll get working on your comments. Regarding the dialog, I was trying to reduce the amount of typing required to specify the remote URL especially if you are cloning an elementary repository.  I too usually copy paste the URL from github because I use the terminal command `git clone ..` anyway.  I was trying to make the UI easier by only having to type the name of the project for the default situation but if it confuses then maybe I should just have one entry for the URL?  I'll rethink the dialog.

src/Widgets/Sidebar.vala
jeremypw: Yes, I wasn't too happy with having too many things in the actionbar.  Most sidebars only have one or two things.

src/Dialogs/CloneRepositoryDialog.vala
jeremypw: Replaced entry with filechooser button

src/Dialogs/CloneRepositoryDialog.vala
jeremypw: This is intended to be pre-filled so the user does not have to keep typing them.  The intention is that these will persist.  Maybe use a dropdown list of common hosts (Github, Gitlab etc)?

src/Dialogs/CloneRepositoryDialog.vala
jeremypw: OK, omitted.

src/Dialogs/CloneRepositoryDialog.vala
jeremypw: No, I don't think so - this is the user name on the remote host, not on the local host. I'll try to improve the variable names

src/Dialogs/CloneRepositoryDialog.vala
jeremypw: I have removed these, but I am worried that pressing Enter can now initiate a premature clone (which is the default action) when the user intended to carry out further edits to the default values.  Should I not make "Clone" the default action now?

src/Widgets/Sidebar.vala
```diff
@@ -60,7 +60,16 @@ public class Code.Sidebar : Gtk.Grid {
             action_name = Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_OPEN_FOLDER,
             action_target = new Variant.string (""),
             always_show_image = true,
-            label = _("Open Folder…")
+            label = _("Open …"),
```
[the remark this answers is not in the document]
jeremypw: Fixed

danirabbit:
@jeremypw I'm not a fan of the dropdown idea tbh. It feels like trying to be too clever and seems more restrictive and then I have to delete parts of my copied string from the web interface and figure out where the pieces go. It sounds painful to me :/

danirabbit:
I think it would be clearer and easier to just have one entry for repository URL instead of trying to split it up into three entries. That seems to against the way code hosts present the URLs for cloning and makes it 3 steps to enter the remote URL instead of one paste step

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,192 @@
...
+        // Set default values.
+        //TODO Persist user choices for these
+        //TODO Use a dropdown of common/recent hosts?
+        //TODO Use a dropdown of recent user names?
+        remote_host_uri_entry.text = "https://github.com";
+        remote_user_name_entry.text = "elementary";
+
+        on_is_valid_changed ();
+
+        show_all ();
```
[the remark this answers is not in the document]
jeremypw: Ah I wondered what was causing that ...

src/Utils.vala
```diff
@@ -337,4 +337,17 @@ namespace Scratch.Utils {
...
     }
+
+    public Gtk.Button make_button_with_icon_and_label (string icon_name, string text) {
+        var box = new Gtk.Box (HORIZONTAL, 0);
+        var image = new Gtk.Image.from_icon_name (icon_name, BUTTON) {
+            margin_end = 3
+        };
+        var label = new Gtk.Label (text);
+        box.add (image);
+        box.add (label);
```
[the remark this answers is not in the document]
jeremypw: Fixed

jeremypw:
Gotta go now - will continue tomorrow ...

stsdc:
GNOME Builder has something like that:
obraz

You can see suggested GNOME projects to clone on the start page.
But also there is a clone repository page:

obraz

jeremypw:
@stsdc Thanks - this is something to work towards.  Probably we'll just produce the simplest minimal functionality in this PR and then we can iterate.

jeremypw:
Removed from potential branches for v8.0.0 as I do not want to hold that up too much and this needs some design thought.

danirabbit:
>  I am trying to ease my own workflow which is mostly cloning projects from the same host/users

@jeremypw I hear what you're saying and that makes sense. What do you think about if we saved that host/org portion of the URL after the first time someone clones a branch? I think that might be a good compromise between assuming that folks are primarily cloning elementary repos and also trying to save time when you're cloning lots of repos from the same org.

So for example:
1. We start with a single empty url entry
2. someone clones for example "https://github.com/elementary/files.git"
3. The next time they open the clone dialog, the url entry is now pre-filled with "https://github.com/elementary/"

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,181 @@
...
+            image_icon: new ThemedIcon ("git"),
+            suggested_local_folder: _suggested_local_folder
+        );
+
+    }
+
+    construct {
+        try {
+            name_regex = new Regex (NAME_REGEX, RegexCompileFlags.OPTIMIZE);
+            url_regex = new Regex (URL_REGEX, RegexCompileFlags.OPTIMIZE);
```
danirabbit, line 37: We should probably make sure that the URL has three parts. So it's only valid with host/user/project and not with just host.

I don't know how difficult that is to do with regex vs using `Glib.Uri.parse ()`?

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,221 @@
...
+                host != null && //e.g. github.com
+                userinfo == null &&  //User is first part of pat
+                (port < 0 || port == 443)) { //TODO Allow non-standard port to be selected
+
+                if (path.has_prefix (Path.DIR_SEPARATOR_S)) {
+                    path = path.substring (1, -1);
+                }
+
+                var parts = path.split (Path.DIR_SEPARATOR_S);
+                valid = parts.length == 2 && parts[1].has_suffix (".git");
```
danirabbit, line 198: We should use `parts[1]` to pre-fill `local_project_name_entry` so that it always reflects the name of the target folder

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,221 @@
...
+                if (path.has_suffix (".git")) {
+                    path = path.slice (0, -4);
+                }
+
+                local_name = Path.get_basename (path);
+            } catch (UriError e) {
+                warning ("Could not parse remote uri");
+                can_clone = false;
+            }
+        }
```
danirabbit, line 150: We could remove this if we pre-filled the entry when we parse the uri

danirabbit (changes requested):
I think we're really close! I have a couple more outstanding comments and this needs some kind of error handling

src/MainWindow.vala
```diff
@@ -1040,6 +1042,47 @@ namespace Scratch {
...
+                                //TODO Make active according to dialog checkbox
+                            }
+                        } catch (Error e) {
+                            warning ("Unable to clone '%s'. %s", uri, e.message);
+                        }
+                    }
+                );
+            } else {
+                //TODO Give feedback
+            }
```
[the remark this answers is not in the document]
danirabbit: `present ()` isn't blocking, so I think as written this will just run immediately

src/Dialogs/CloneRepositoryDialog.vala
jeremypw: Done

src/Dialogs/CloneRepositoryDialog.vala
jeremypw: Done

src/MainWindow.vala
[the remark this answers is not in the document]
jeremypw: Yes, on reflection there isn't an issue in calling an async method with `.begin ()` inside the closure (and the `clone ()` method isn't actually async anyway at the moment.  I have tried to provide some feedback during cloning in #1587 but maybe it is sufficient for this PR to let the UI block for a while while the cloning is progress and deal with the progress after some design consideration? While cloning elementary-sized repos it blocks for 2 - 3 seconds for me. This feature won't be in version 8.0 anyway.

jeremypw:
@danirabbit I've added cloning error handling and persisting the remote and projects folder entries.

src/Services/GitManager.vala
```diff
@@ -112,5 +112,38 @@ namespace Scratch.Services {
 
             return build_path;
         }
+
+        //TODO Make this a real async function that does not block the main loop.
```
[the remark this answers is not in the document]
jeremypw: Looks like we need to use a real GLib.Thread.  Async functions run in the same thread as the main loop so can block the UI.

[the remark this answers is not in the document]
jeremypw: Now using a GLib.Thread to wrap the Ggit.clone ().  Seems to work (as judged by spinner operating smoothly during cloning)

danirabbit (changes requested):
Somewhere along the way we seem to have lost "activates default" on entries

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,259 @@
...
+        content_box.attach (new CloneEntry (_("Name of Clone"), local_project_name_entry), 0, 2);
+        content_box.attach (revealer, 1, 2);
+        content_box.show_all ();
+
+        custom_bin.add (content_box);
+        custom_bin.show_all ();
+
+        clone_button = new Gtk.Button.with_label (_(CLONE_REPOSITORY));
+        clone_button.show ();
+        add_action_widget (clone_button, Gtk.ResponseType.APPLY);
```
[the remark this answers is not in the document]
jeremypw: Included in my commit

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,259 @@
...
+public class Scratch.Dialogs.CloneRepositoryDialog : Granite.MessageDialog {
+    public bool can_clone { get; private set; default = false; }
+
+
+    // Git project name rules according to GitLab
+    // - Must start and end with a letter ( a-zA-Z ) or digit ( 0-9 ).
+    // - Can contain only letters ( a-zA-Z ), digits ( 0-9 ), underscores ( _ ), dots ( . ), or dashes ( - ).
+    // - Must not contain consecutive special characters.
+    // - Cannot end in . git or . atom .
+    private const string CLONE_REPOSITORY = N_("Clone Button");
```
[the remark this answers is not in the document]
jeremypw: Sorry, left over from testing...

src/Dialogs/CloneRepositoryDialog.vala
```diff
@@ -0,0 +1,259 @@
...
+                spinner.stop ();
+            }
+
+            revealer.reveal_child = value;
+        }
+
+        get {
+            return revealer.reveal_child;
+        }
+    }
```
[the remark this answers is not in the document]
jeremypw: OK, the spinner at this point is a proof of principle that the cloning thread was not blocking the UI.  I wasn't sure how to show it in practise.  I'll show another page as you suggest.

jeremypw:
There doesn't seem to be a way of cancelling an ongoing cloning process afaict.

PR #1551, opened 2025-04-20
Fix regressions in search bar behaviour

### LINKED ISSUE #1549: Search entry is no longer highlighted after pressing Ctrl F
### What Happened?

When pressing Ctrl + F the current search is no longer hightlighted. This is really disruptive when I have something in my clipboard that I want to replace the current search with or just want to start a different search since there's no easy way to clear the searchbar with just the keyboard

### Steps to Reproduce

1. Start a search by pressing Ctrl F
2. Do some work (optionally copy something to your clipboard)
3. Start a search again by pressing Ctrl F
4. Have to navigate to the search bar with the mouse to clear it so that you can move on to the next task

### Expected Behavior

Restore the previous behavior of selecting search contents when pressing Ctrl F. It's easy to cancel a selection if you don't want to clear the old search. It's much harder to clear the search bar if you want to start a new search

### OS Version

8.x (Circe)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

### LINKED ISSUE #1550: Manually overwritten search term is reverted on immediately pressing `Ctrl-F`
### What Happened?

Tried to change a search term by deleting the old entry and writing a new one. On pressing `Ctrl-F` the old entry reappeared.  This only happens if `Ctrl-F` is the next action and it is not necessary to press `Ctrl-F` as the new search has already happened. Nevertheless it is unexpected and annoying.

### Steps to Reproduce

1. Perform a search
2. Focus the search entry and replace the search term using the search term
3. Immediately press `Ctrl-F`
4. The previous search term reappears

### Expected Behavior

The old search term should be forgotten if it is manually overwritten.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #1549
Fixes #1550 

It is not clear why the search entry no longer automatically selected the contents on focus in - this should be the default behaviour of the widget. It is now done explicitly after focusing signal handlers have finished.

The previous search term was sometimes reappearing unexpectedly because it was still selected in the document.  This is fixed by not using the selected text as the search term when the search entry is already focused.


leonardo-lemos (approved):
LGTM, nice job!

PR #1554, opened 2025-05-17
Use 'development-target' tag of docker image to build Gettext catalog updates

leonardo-lemos:
We are using an old container image to build gettext updates, it's now failing to build because the image contains a old version of GLib, so using 'development-target' image fixes the Gettext Catalogs build.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -8,7 +8,7 @@ jobs:
   build:
     runs-on: ubuntu-22.04
     container:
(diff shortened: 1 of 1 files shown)


ryonakano:
This would work, but maybe we may want to update `next-unstable` to point to `circe-unstable` instead of `horus-unstable`?

https://github.com/elementary/docker/blob/master/next-unstable

leonardo-lemos:
@ryonakano I totally agree! Closing this PR in favor of https://github.com/elementary/docker/pull/47

PR #1565, opened 2025-05-22
Jeremypw/vala symbols/improve tooltip

### LINKED ISSUE #1410: Add more information to Symbol plugin tooltips
### Problem

The symbol plugin tooltips aren't super useful currently. They just show the name of the symbol which is very redundant

### Proposal

Tooltips should show as much information as possible for that symbol to help identify it. For example what type of symbol it is and any code comment associated with that symbol

### Prior Art (Optional)

Check out the tooltips in the sidebar on Valadoc.org: https://valadoc.org/granite-7/Granite.SettingsPage.html

Screenshot from 2024-01-19 13 39 13

jeremypw:
Fixes #1410 

 - Extract the source definition from the document and use as tooltip
 - Fix some consequential timing issues on restoring documents with the outline open
 - Update some code style of ValaSymbolItem and move some code from ValaSymbolItem into the construct clause

It proved simpler and better to copy the definition from the document rather than trying to reconstruct from the parsed symbols.  The tooltip includes text to the end of the first line pointed to by the symbol.source_reference as this may contain extra useful information such as parameters and/or comments.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -579,7 +579,6 @@ namespace Scratch {
             });
 
             document_view.realize.connect (() => {
-                document_view.update_outline_visible ();
                 update_find_actions ();
             });
 
@@ -694,6 +693,7 @@ namespace Scratch {
             }
 
             document_view.request_placeholder_if_empty ();
+            document_view.update_outline_visible ();
             restore_override = null;
             if (focused_file != null) {
                 folder_manager_view.expand_to_path (focused_file.get_path ());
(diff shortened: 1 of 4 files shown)


jeremypw:
Getting crash when starting up with certain documents/conditions - diagnosing ...

jeremypw:
Looks like the crashing was due to accessing the source view from a separate parsing thread without locking it. This is now fixed.  The parsing thread is now made to yield to the main thread after each child construction and this allows other documents to load and the UI to be used even though there is a document (e.g. a vapi) that takes a long time to finish parsing.  Because of this the time allowed to construct the symbol pane has been increased.

PR #1566, opened 2025-05-25
Jeremypw/plugins/fix libpeas2

### LINKED ISSUE #1560: Bracket completion no longer works
### What Happened?

Bracket completion seems to no longer work at all for any type of bracket

### Steps to Reproduce

1. Open a code file
2. Type in a starting bracket like `(`, `{`, or `[`

### Expected Behavior

The ending bracket should be automatically added

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #1560 

In PluginManager:
 - Make namespaces explicit
 - Make ExtensionSet a construct property

In master, plugins are getting deactivated as soon as loaded for reasons that are obscure.  After much trial, error and elimination it appears that the above two changes fix the issue.

View the diff hiding whitespace changes.


danirabbit (changes requested):
Definitely fixes the issue! A couple nitpicks

danirabbit (approved):
:rocket:

PR #1567, opened 2025-05-25
Increase target size in plugin list

jeremypw:
Check boxes present quite a small target so change to the larger Gtk.Switch.

Screenshot from 2025-05-25 14 04 17


danirabbit (approved):
Looks good :rocket:

PR #1522, opened 2025-01-28
Load children asynchronously

jeremypw:

PR #1571, opened 2025-05-26
Fix global search - load children synchronously when expanding sidebar to path

### LINKED ISSUE #1570: Global search does not always show all matches
### What Happened?

Performed a global search (case insensitive, partial words) for a common word - on first attempt only some of the expected files in the sidebar showed badges.  Repeating the same search showed successively more badges.

### Steps to Reproduce

1. Perform a global search for a term appearing multiple times in a lot of files
2. Notice that on first attempt not all the expected files are badged.

### Expected Behavior

All expected matches should appear as badges on the first attempt at a global search.

### OS Version

8.x (Circe)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #1570 

Also fixes expanding to focused document on startup.

 - [x] Add a method to load children synchronously to FolderItem
 - [x] Use synchronous method in `expand to folder`
 - [x] DRY loading code in FolderItem

This means that the speed gain from #1522 is lost if a large folder needs to be loaded automatically to expand to the focused document but manually expanding large folders will still be fast.


leonardo-lemos (approved):
Works as expected

PR #1580, opened 2025-05-27
SourceView: remove unused variables

danirabbit:
Fix a couple compilation warnings by removing dead code


jeremypw (approved):
Thanks! - I've been meaning to do that.

PR #1581, opened 2025-05-28
Move buttons from actionbar to projectchooser

jeremypw:


src/Widgets/Sidebar.vala
```diff
@@ -74,23 +67,15 @@ public class Code.Sidebar : Gtk.Grid {
...
-            orientation = HORIZONTAL,
+        var project_more_button = new Gtk.MenuButton () {
             hexpand = true,
-            selection_mode = NONE,
-            column_spacing = 0,
-            max_children_per_line = 2
+            use_popover = false,
+            menu_model = project_menu_model,
+            label = _("Manage project folders…"),
+            xalign = 0.0f
```
danirabbit, line 75: There's no `xalign` property for buttons in GTK4, so you have to make a label separately and pack it

jeremypw:
I think this has been superceded now?

PR #1582, opened 2025-05-29
Highlight selection plugin: Fix highlighting when same word reselected

jeremypw:
#1520 

 - Simplify workings following recent changes to master 
 - `deselected` signal not required
 -


jeremypw:
Blocking as conflicts with #1622 which needs to be merged first.

PR #1608, opened 2025-06-24
Clone repository: Show progress

### LINKED ISSUE #1628: Git clone success dialog should be a toast
### What Happened?

When a git repository is successfully cloned it shows a whole dialog box stating that the clone was successful

### Steps to Reproduce

1. Clone a git repository
2. See a dialog box when it succeeds 

### Expected Behavior

This should be a Granite.Toast instead if the window is focused or a system notification if the window is not focused.

Toasts are less intrusive and more appropriate than a dialog when an additional decision doesn't need to be made, and notifications are more helpful for long-running tasks where I might have navigated to another workspace while I'm waiting for the task to be completed.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #1628 

Less intrusive solution to indicating cloning in progress (spinner) and finished (toast).


jeremypw:
I know I have left progress machinery in the dialog at the moment in case this can be re-used elsewhere.

jeremypw:
I am going to close this as the preferred solution advised by @danirabbit is to just show cloning completion with a toast. It is not therefore necessary to use remote callbacks to get detailed information about the progress of the clone.

PR #1613, opened 2025-06-25
Bump libgit2-glib dependency

jeremypw:
As Code already cannot compile on OS7 (because of libpeas-2) we can bump the version of libgit2-glib used.  On OS8 we no longer need to carry the `.vapi` either as it is now in `/usr/share/vala`


leonardo-lemos (changes requested):
Can you also update the lib version in the Flatpak manifest, please?

jeremypw:
@leonardo-lemos Thanks - I am working on that but having trouble getting the flatpak to build atm.  I'll revert to draft.

jeremypw:
@leonardo-lemos Turns out that for some unreason the flatpak does not build with platform v8.1 or v8.2 but it builds with v7.1 and v8 so I bumped the platform to v8.   I haven't bumped other modules like libgit2 or libssh2 so far.

io.elementary.code.yml
```diff
@@ -1,6 +1,6 @@
-app-id: io.elementary.code
+app-id: io.elementary.code8
```
danirabbit, line 1: Can we revert the ID change here please :)

danirabbit (changes requested):

io.elementary.code.yml
jeremypw: Oops, that was supposed to be a temporary change

danirabbit (approved):

PR #1616, opened 2025-06-29
Implement create new from template

### LINKED ISSUE #297: New file from template
The new file menu in the folder plugin should have a Templates submenu the same as Files does

jeremypw:
Fixes #297 

Adapted from Files code. 

Some undecided issues:  

(1) Should templates automatically enter rename mode (currrently not). I am assuming templates will be named something like "Application.vala" so will not need renaming.

(2) Should templates be automatically opened (currently not).  I have kept the same behaviour as new file and folder but selected the new item so it can be easily opened if required.

(3) How to make where to put the template files more discoverable or configurable.

Screenshot from 2026-04-27 20 56 10


jeremypw:
Need to fix the path from where the templates are taken.

jeremypw:
@danirabbit 
I now notice that there is some quite extensive old code already implementing a "TemplateManager".  but it is not clear how this was intended to work and afaict it has never actually worked? It appears to relate to "Project Templates" so maybe similar to the unmerged PR #996?   Hopefully this PR does not duplicate existing code.

jeremypw:
On second thoughts, I think we ought to enter renaming mode so the actual desired name can be entered more conveniently. I'll implement this soon.

jeremypw:
> I don't think we should have a setting for the template folder

The idea is to keep Code specific templates separate (I am not sure why templates for different apps should all be in the same folder tbh). However, I'll bow to majority opinion and revert that.

jeremypw:
Checked this does not crash with too many templates and already uses GLib.Menu. The menu appears a little different as it does not have a separate Template submenu or "-too many templates" item. Will update to make same as Files in appearance.

jeremypw:
The menu now appears the same as Files.

danirabbit (approved):
Nice job!

PR #1618, opened 2025-06-29
Restart shell in terminal pane if it exits

### LINKED ISSUE #457: Exiting process running in terminal leaves terminal useless
When the terminal is open and the process running inside of it ends, the terminal is left in an unusable state until application is restarted. Ideally I would think this should restart bash, or at least have a way to restart it.

This can be reproduced by opening Code's terminal and typing `exit`. You can open and close the terminal afterwards and it will continue to be in the exited state.

jeremypw:
Fixes #457 

There was already code to hide the terminal if it's child exited but it had regressed and was not working.

This PR makes the terminal hide again if the shell exits.

It also restarts the shell at the last saved location and clears the screen ready for it to be used again. To enable this, the shell location is saved to settings every time it changes through app action instead of just at closing the app. (NOTE: Change of location through the shell `cd` command is not detected so not saved)

### DIFF
--- a/src/Widgets/Terminal.vala
+++ b/src/Widgets/Terminal.vala
@@ -130,8 +130,7 @@ public class Code.Terminal : Gtk.Box {
         Posix.kill (child_pid, Posix.Signal.TERM);
         terminal.reset (true, true);
         spawn_shell (dir);
-        var settings = new Settings (Constants.PROJECT_NAME + ".saved-state");
-        settings.set_string ("last-opened-path", dir);
+        Scratch.saved_state.set_string ("last-opened-path", dir);
     }
 
     private string get_shell_location () {


src/Widgets/Terminal.vala
```diff
@@ -44,7 +44,13 @@ public class Code.Terminal : Gtk.Box {
...
 
         terminal.child_exited.connect (() => {
-            GLib.Application.get_default ().activate_action (Scratch.MainWindow.ACTION_PREFIX + Scratch.MainWindow.ACTION_TOGGLE_TERMINAL, null);
+            //Hide the exited terminal
+            var win_group = get_action_group (Scratch.MainWindow.ACTION_GROUP);
+            win_group.activate_action (Scratch.MainWindow.ACTION_TOGGLE_TERMINAL, null);
+            //Get ready to resume at last saved location
+            spawn_shell (Scratch.saved_state.get_string ("last-opened-path"));
+            //Clear screen of new shell
+            terminal.feed_child ("clear -x\n".data);
```
danirabbit, line 53: Is there a reason we don't add this to the action?

jeremypw: The normal toggle-terminal action does not exit the current shell (or any foreground process running in it) so we do not want to respawn in that case.  This is for when the shell exits unexpectedly or through the user typing `exit`.

src/Widgets/Terminal.vala
```diff
@@ -125,6 +130,8 @@ public class Code.Terminal : Gtk.Box {
         Posix.kill (child_pid, Posix.Signal.TERM);
         terminal.reset (true, true);
         spawn_shell (dir);
+        var settings = new Settings (Constants.PROJECT_NAME + ".saved-state");
+        settings.set_string ("last-opened-path", dir);
```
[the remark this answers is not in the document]
jeremypw: Good question!

jeremypw:
Looks like the shortcut for `action-open-in-terminal` (`<Ctrl><Alt><Shift>t`) is not distinguished from the shortcut for `action-toggle-terminal` (`<Ctrl><Alt>t`).  Will have to assign a different shortcut.

jeremypw:
@lenemter I have used the existing setting object as suggested, thanks.  Regarding the `open-in-terminal` action, it does seem to be working.  When the shortcut is used it navigates the terminal pane to the current project *build path* but when used with the "Open in Terminal Pane" option in the sidebar folder context menu it navigates to the actual folder specified.  Not sure why there is this difference, but that is outside the scope of this PR.

If you want to open the folder in the Terminal app, then you have to use the context menu - there is no shortcut.

I am going to propose a PR to clarify the naming of this action.

lenemter (approved):
Oh, I got confused by the action name. I thought it would open terminal pane in the directory of currently opened file, but it opens the directory of the current git project

PR #1619, opened 2025-06-29
Minimal install of codecore library; use current version

jeremypw:
This PR discontinues providing a package for codecore together with `.vapi` files etc as we no longer support third party plugins and the core plugins do no need them.  

In passing the shared library version tracks the project version

The flatpak continues to build and run.


tintou:
The behavior of `install_dir` is specific in Vala, see https://mesonbuild.com/Vala.html#building-libraries

src/meson.build
```diff
@@ -84,14 +84,13 @@ codecore = library(
     config_header,
     dependencies: dependencies,
     install: true,
-    install_dir: [true, true, true],
-    version: '0.0'
+    version: meson.project_version ()
```
tintou, line 87:
```suggestion
    soversion: '0',
    version: meson.project_version ()
```

jeremypw:
@tintou Thanks for the info - the meson manual seems a bit contradictory/confusing with regard to the `install_dir` keyword parameter.  Is issue #635 a valid issue that needs addressing? Do some versions of meson not accept the array of booleans? 

Beyond that, do we need to publish the codecore library as a package at all? We do not support third-party plugins now.  As far as I can tell the core plugins only need to find the `.so` file - they do not need the `,vapi`, `.deps` or `h` files.   My last commit removes the installation of these and Code still builds and runs normally.   Or is this a step too far?  @danirabbit thoughts?

tintou:
array of booleans is supported since a very long time (I think since Meson supports Vala so almost its inception)

jeremypw:
@tintou  Thanks for clarifying and closing the issue.  I'll leave this as draft for @danirabbit to comment on whether we want to continue publishing the codecore package.

jeremypw:
@tintou Just to check - if we are not publishing codecore as a Vala library do we need the array of booleans - or are there other advantages to keeping the array?

tintou:
If we are not allowing out-of-tree plugins there is no need to install the header/gir/vapi indeed

PR #1622, opened 2025-06-30
Fix crash from highlight selection plugin when deleting selection

### LINKED ISSUE #1443: Code crashes when deleting lines

jeremypw:
Fixes #1443 

The cause seems to be a clash between the handlers for the "selection-changed" and "deselected" signals in the plugin.  One is throttled and the other not.

For simplicity the "deselected" signal is lost (only used by the plugin) and a deselection now triggers a throttled "selection-changed" signal.


zeebok (approved):
Code looks good. If you would like testing, let me know

jeremypw:
@zeebok Looks like that issue you identified also occurs with master so I'll fix that separately. Thanks for checking!

PR #1625, opened 2025-07-02
Disable some actions in secondary window

jeremypw:
Based on #1624 

Limit certain behaviour that would be complicated to synchronize/reconcile with the primary window.

 - [x] Access to terminal
 - [x] Change active project (ChooseProjectButton)
 - [ ] Change active project (context menu)
 - [x] Clone repository
 - [ ] Checkout/create branch

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -155,13 +155,13 @@ namespace Scratch {
             if (active_window == null) {
                 if (location_jump_manager.has_selection_range () && location_jump_manager.has_override_target ()) {
                     RestoreOverride restore_override = location_jump_manager.create_restore_override ();
-                    add_window (new MainWindow.with_restore_override (true, restore_override));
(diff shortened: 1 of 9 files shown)


danirabbit:
I don't like this at all :( We should definitely make active project per window for example

jeremypw:
@danirabbit Funnily enough I came to the same conclusion and am implementing that now.  It should be possible to allow most actions to be done in either window but I do not want to start saving state per window atm so there may be some restrictions.

I envisage being able to work on two *different* projects without restrictions but I think we will have to avoid having the same active project in both windows unless git actions are disabled in one of them.

jeremypw:
I think I'll close this and submit a more limited PR implementing per-window active project, based on #1574 as that touches the same code.

Then we can think about any necessary limitations as and when specific issues arise.

PR #1627, opened 2025-07-03
Keyboard accessibility: Set an accelerator for Close Other Tabs action

jeremypw:
For convenience.

Proposed shortcut is `<Shift><Control>W`

Ideally needs to be paired with a "Pin Tab" functionality so docs can be opened temporarily and then closed without closing the doc actually being worked on.


danirabbit (approved):
Seems in line with what other platforms do for this 👍

PR #1635, opened 2025-07-17
SymbolPane: Fix filter for C symbols

### LINKED ISSUE #1597: Symbol pane filter should work for C symbols too
### What Happened?

When working on Vala files it is useful to filter the symbol list to find the ones needed.  However trying this when on a C file did not filter the list (even though the filter entry was still there).

### Steps to Reproduce

1. Open a C file
2. Open the symbol pane
3. Type something in the filter entry
4. The list is not filtered

### Expected Behavior

The C symbols are filtered in the same way as Vala symbols

### OS Version

8.x (Circe)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #1597

The CtagsSymbol object was not declaring that it implements the SymbolItem interface so was being ignored by the filter function.

### DIFF
--- a/src/SymbolPane/C/CtagsSymbol.vala
+++ b/src/SymbolPane/C/CtagsSymbol.vala
@@ -16,7 +16,7 @@
  *
  */
 
-public class Scratch.Services.CtagsSymbol : Code.Widgets.SourceList.ExpandableItem {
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
Works as expected, nice job!

PR #1636, opened 2025-07-17
SymbolPane: improve tooltips for C tags

jeremypw:
We implemented more informative tooltips for Vala symbols recently.  This does the same for C tags (as far as the parser allows).

The parser does not provide start and end lines for the tags, only the line on which the tag name occurs. As C methods often have the type on the previous line, the start line is decremented for methods.  It is necessary to use Markup.escape_text as some C symbol definitions contain markup characters.


zeebok (approved):
Glad both previous and same line scenarios are covered!

PR #1640, opened 2025-08-12
Fix open-folder shortcut

jeremypw:
- Resolve shortcut conflict
 - Add target to open-folder shortcut
 - Use separate action for open project with no target else accel label does not show accel string

It seems that Granite.AccelLabel does not show accelerator strings for accelerators containing a default target so a separate action with no target is created for the cases where a file chooser will be shown and this is used instead.


lenemter (approved):
All shortcuts appear to work as intended so LGTM!

PR #1641, opened 2025-08-12
Cloning: Show progress in sidebar

### LINKED ISSUE #1628: Git clone success dialog should be a toast

jeremypw:
Fixes #1628

During cloning the dialog hides and a spinner appears in the sidebar.  On a successful clone a toast appears in the sidebar.  If the clone is unsuccessful the behaviour is as before.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1055,34 +1055,24 @@ namespace Scratch {
                 // Persist last entries (not necessarily valid)
                 Scratch.settings.set_string ("default-remote", clone_dialog.get_remote ());
                 Scratch.settings.set_string ("default-projects-folder", clone_dialog.get_projects_folder ());
-                // Clone dialog show spinner during cloning so keep visible
                 //TODO Show more information re progress using Ggit callbacks
                 if (res == Gtk.ResponseType.APPLY && clone_dialog.can_clone) {
-                    clone_dialog.cloning_in_progress = true;
+                    sidebar.cloning_in_progress = true;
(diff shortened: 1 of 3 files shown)


jeremypw:
Note: The "Revert shortcut clash" commit was included in error and reverted. It belongs in #1640.

jeremypw:
Just noticed the issue also mentioned sending a notification if the window is not focused.  This PR does not implement that at present - it is targeted on the toast aspect.

jeremypw:
Now implemented notification if not focused.  Do we need to withdraw notification (like Terminal) if Code window becomes active?

jeremypw:
@zeebok  Thanks for the review!  I'll merge and hope that will stimulate further suggestions for appearance improvement.  I'll look into  progress indication in the new dock.

PR #1652, opened 2025-10-31
Fix typo in .desktop file

decathorpe:
Without this change, desktop-file-validate considers the .desktop file to be invalid:

```
(...)/usr/share/applications/io.elementary.code.desktop: error: file contains key "icon" in group "Desktop Action NewWindow", but keys extending the format should start with "X-"
```


danirabbit:
Thanks for noticing this!

PR #1656, opened 2025-11-05
Option to keep projects sorted

### LINKED ISSUE #568: Sort open project folders

jeremypw:
Fixes #568 

 - Convert ACTION_ORDER_FOLDERS into a toggle action
 - Convert "Manage Project Files menu child into a popover 
 - Use ModelButton for "Alphabetize" action
 - Modify Fuzzy Finder plugin to add a button with accel label to the project menu
 - Add a "order-folders" setting for persistence

Ordering the project folders in SourceList is not very efficient but as there are relatively few and that widget may well be replaced in Gtk4 it does not seem worth doing anything more complicated.

BEFORE

Screenshot from 2025-11-05 14 37 23

AFTER

Screenshot from 2025-11-05 14 41 08


danirabbit (changes requested):
Gtk.ModelButton is going away in GTK4 so it's not a good idea to use it in new code. I don't see anything here that can't be done in GLIb.Menu?

PR #1663, opened 2025-11-09
Focus terminal on enter, strip trailing spaces on doc focus out

### LINKED ISSUE #581: Automatically clear tabs/spaces from empty lines?

jeremypw:
Fixes #581 

It is considered potentially problematic to strip spaces on autosave so this solution is proposed.  Stripping only occurs when the document is unfocused on the assumption that in that case the user is no in mid edit.

It is inconvenient to have to both move and click the pointer to focus the terminal and the same to refocus the document.  Also lint errors are often caused by failing to force save the document first (trailing spaces are not stripped during autosave).

 - Implement focus follow between the terminal and the current document (does this need a setting?)
 - Strip trailing spaces from the document when focused out (subject to the setting). Leave whether this change is saved to the AutoSave setting,


ryonakano (commented):
I'm not sure if we should reuse `strip-trailing-on-save` setting because this is not triggered on save.

zeebok (approved):
Code looks good. Is Code not on Gtk4 yet?

jeremypw:
@zeebok Now that Terminal has been ported I'll start to prepare Code for Gtk4 (contributions welcome!).  Will need to get the number of outstanding PRs down, one way or another, before actually doing the port though.

Issue #1659, opened 2025-11-06
Add an option in Preferences to disable syntax highlighting globally


lobre:
### Problem

Currently, Code always applies syntax highlighting based on the detected or selected language.

The only way to disable it for a given file is to manually change its language mode to “Plain Text”, but it must be done per file, every time.

For users who prefer a minimal, distraction-free interface, this is not ideal.

Personally, I’ve developed the habit of working without syntax highlighting — I find it helps me focus on code structure and readability rather than colors. While I understand this is a subjective preference, a global toggle would make Code more flexible for different workflows.

### Proposal

Add a setting in Preferences as as:

[ ] Disable syntax highlighting

PR #1664, opened 2025-11-09
Disable-able Syntax Highlighting Preference

lobre:
Hey,

I have tried to answer my self-created issue:

https://github.com/elementary/code/issues/1659

As a sum-up, I have added a new syntax-highlighting GSettings key plus a toggle under `Preferences -> Interface` so users who prefer monochrome editing can disable highlighting globally while leaving the existing language picker active for commenting, Markdown shortcuts, etc.

The SourceView simply toggles `Gtk.SourceBuffer.highlight_syntax`, so search hits and other non-language effects keep working, and the language dropdown now says "Document language" to better reflect its broader role.

Happy to adjust any of this if you think there is a better way to expose the preference, or if there is anything with the implementation.

### DIFF
--- a/data/io.elementary.code.gschema.xml
+++ b/data/io.elementary.code.gschema.xml
@@ -106,6 +106,11 @@
       <summary>Highlight Matching Brackets</summary>
       <description>Whether Code should highlight matching brackets.</description>
     </key>
+    <key name="syntax-highlighting" type="b">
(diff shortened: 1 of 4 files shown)


jeremypw:
I'll let you complete your PR as you started first and look like you're on the right lines.

PR #1602, opened 2025-06-19
elementary high contrast (light) stylescheme

jeremypw:
Fixes #548 

Based on the elementary-light theme but tweaked to produce better contrast, using the elementary palette mainly.

 - Make most colors darker/more saturated
 - Make most backgrounds lighter
 - In some cases make style bold
 - Improve contrast between line number panel and text panel background (which is what the issue was about)
 - Make the current line background different from the text background

Preferably needs testing by a user who prefers or requires a high-contrast theme, and on as many different languages as possible. I checked on Vala, C, XML and Meson mainly.

The theme probably works better with a bold font but for comparison with "classic" below, the same regular font is used.

Comparison between "elementary-high-contrast" (on the left) and "classic" scheme (on the right)
Screenshot from 2025-10-27 16 38 07

PR #1678, opened 2025-11-18
Feature: Option to use bold font with high-contrast theme

jeremypw:
Option to increase the readability of the high contrast scheme (see #1602) beyond changing the colors without having to choose a new font.


jeremypw:
Bump to next feature release

jeremypw:
Closing as on reflection, it is not worth having a toggle that is likely to set once and can be done by choosing the bold version of the required font in existing UI.

PR #1680, opened 2025-11-24
Feature: Add action to find matching bracket

jeremypw:
Builtin highlight matching brackets only works over a limited number of characters.  This works over whole document, scrolls to match and checks whether indent is the same.

This action (but not accelerator) is now exposed in the document context menu


src/Widgets/SourceView.vala
```diff
@@ -756,5 +765,153 @@ namespace Scratch.Widgets {
...
+                return;
+            }
+
+            if (start_indent != end_indent || !found) {
+                var min_line = int.min (start_line, end_line);
+                var max_line = int.max (start_line, end_line);
+                var parent_window = get_toplevel () as Gtk.Window;
+                var dialog = new Granite.MessageDialog (
+                    found ? _("Matching bracket has different indent") : _("No matching bracket found"),
+                    _("You may have omitted a required bracket or inserted extra brackets between lines %i and %i, or you may need to adjust the indents").printf (min_line, max_line),
```
zeebok: Would it be better to have a similar `found` ternary op so the full message matches the summary?

zeebok (dismissed):
Just one optional comment but looks good to me! Excited to make use of this feature!

src/Widgets/SourceView.vala
jeremypw: Yes, I'll do that.

jeremypw:
I'd like to add extra functionality to find where the error occured if possible but that will have to wait for another PR.  I'd also like to expose the shortcut somehow but the document context menu does not currently show accels. Again, another PR.

zeebok (approved):

zeebok:
> I'd like to add extra functionality to find where the error occured if possible but that will have to wait for another PR. I'd also like to expose the shortcut somehow but the document context menu does not currently show accels. Again, another PR.

I wonder if adding a Cheat Sheet option in the gear menu that gives a bunch of helpful shortcuts and features is a decent enough way to expose it?

jeremypw:
> I wonder if adding a Cheat Sheet option in the gear menu that gives a bunch of helpful shortcuts and features is a decent enough way to expose it?

Yeah, this idea has been raised for a number of apps and always been put off in the hope a global system might avoid reinventing this wheel for each app. I think the approved way at the moment is to show shortcuts in tooltips and menus so I'll have to see whether this can be achieved here.  But I will leave that to a follow up PR as it covers a number of actions.

PR #1683, opened 2026-01-01
One Pastebin upload dialog

### LINKED ISSUE #949: Pastebin Dialog: Can open infinite format dialogs
Looks like it's not blocking so you can open an infinite amount of them.

callerun:
Block creation of multiple Pastebin dialogs at the same time. 

If one dialog is already open and the user press "Upload to Pastebin", the already open dialog gets focused.

I think this should fix #949


jeremypw (approved):
Makes sense and fixes the linked issue.  Thanks for your contribution!

PR #1684, opened 2026-01-03
Ensure window not stuck fullscreen

jeremypw:
It was found possible for Code to get stuck in a state where it always restores fullscreen regardless of settings.  

This PR fixes the issue by ensuring the window is not fullscreen unless dictated by the settings.


zeebok (approved):
Strange this can happen since a size is requested but this seems to be a good fix

PR #1687, opened 2026-02-06
Option to keep folders sorted

### LINKED ISSUE #568: Sort open project folders

jeremypw:
Fixes #568 

Supercedes #1656 

Uses SettingsAction in Sidebar and only changes one MenuItem. Eliminates MainWindow action (there is no accel).


jeremypw:
Looks like some external factor is causing flatpak CI to fail now (after merging `master`). Launchpad down?

PR #1696, opened 2026-02-23
Feature: Separate terminal zoom

### LINKED ISSUE #1492: Independent font size for terminal
### Problem

At present the terminal font size is linked with the document font size. This is not always what is desired.  It would be good to be able to change these separately, e.g. by Control-scrolling on the terminal

### Proposal

Uncouple the terminal and document font sizes (or allow the ratio between the two to be changed) so that the terminal font size can be changed separately.

### Prior Art (Optional)

_No response_

jeremypw:
Fixes #1492 

Allows separate adjustment of terminal and sourceviews to desired font size.


jeremypw:
Needs rewriting to not use `scroll-event`

zeebok (approved):

PR #1701, opened 2026-02-26
ProjectFolderItem: code style

jeremypw:
Similar to danirabbit's work on Files
 
- SPDX header
-  Inline namespace
- public signals and properties before private
- public methods before protected methods before private methods 

The "authored by" line is retained although many other authors now exist.  Not sure of the policy on this.
Not sure of correct position of the ``static construct` clause relative to the instance creator function.

No change in functionality.


jeremypw:
Closing as easier to redo rather than fix conflicts

PR #1710, opened 2026-04-15
SourceList: use key controller

danirabbit:


jeremypw (commented):
The code is good, but in practice this feature is almost impossible to use because it hard to get keyboard focus into the sidebar.   The only way I found was to start a rename using right-click and then cancel it - the keyboard focus then remains in the sidebar.

If it is intended to make the sidebar (and other widgets other than the textview and terminal) operable with the keyboard *alone*, in a discoverable way,  then there is work to do.

jeremypw:
Files uses `<Control><Tab>` to switch focus from view to sidebar but this does not work in Code as this shortcut is used by the textview.

danirabbit:
Yeah I had a hard time testing this and I think the keyboard navigation experience definitely needs to be addressed. My goal here however is just to prepare for GTK4 porting

jeremypw:
OK, it sounds like the goal is that the sidebar should be accessible/controllable with the keyboard but that can be achieved later.

jeremypw (approved):

PR #1713, opened 2026-05-03
SourceView: Save and restore selection when focus out and in

jeremypw:
It was noticed that since #1663, it is possible to accidentally stray into the terminal pane when selecting text and this causes the selection to be lost.  

There does not seem to be an advantage in losing the selection when the document focuses out so this PR saves and restores the selection on focus out/in.

Opinion sought whether this behaviour is undesirable under some circumstances and whether it is worth making optional.

Note that the selection is also now maintained when switching tabs.


jeremypw:
CLosing as stripping no longer takes place if there is a selection.

PR #1719, opened 2026-05-17
Feature/sidebar focus

### LINKED ISSUE #1711: Sidebar difficult to focus using keyboard
### What Happened?

The sidebar sourcelist code supports keyboard events but there is no obvious way of getting keyboard focus onto the sidebar

### Steps to Reproduce

Try to move keyboard focus onto the sidebar using only the keyboard.

### Expected Behavior

It shoul be possible to move keyboard focus onto the sidebar using the keyboard without compromising existing keyboard functionality in the text view.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

Dipanshusinghh:
### Summary
This PR fixes the keyboard accessibility issue where the project sidebar could not be focused via keyboard navigation.

### Changes Made
- Made `Code.Sidebar` focusable by setting `can_focus = true`.
- Added `ACTION_FOCUS_SIDEBAR` (`action-focus-sidebar`) to `MainWindow`.
- Bound `<Control><Alt>Left` to focus the sidebar directly (and opens it automatically if it is currently hidden).

Fixes #1711


jeremypw (changes requested):
Unfortunately this does not seem to work yet - focus remains with the document for me.  Not sure why yet, but it *does* work if you focus the `choose_project_button` instead. So its something to do with making sure the treeview within the SourceList gets (and retains) the final focus.

Also may need a matching hotkey to return focus to the document but that can be addressed later.

I notice that if you do focus the choose_project_button then you can reach other widgets in the headerbar by using the standard `Tab` and `<Shift>Tab` keys. But not the actionbar.

src/Widgets/Sidebar.vala
```diff
@@ -45,6 +45,7 @@ public class Code.Sidebar : Gtk.Grid {
     construct {
         orientation = Gtk.Orientation.VERTICAL;
         get_style_context ().add_class (Gtk.STYLE_CLASS_SIDEBAR);
+        can_focus = true;
```
jeremypw: I am not sure this right - Sidebar is a container (Grid) which cannot have keyboard focus itself

Dipanshusinghh:
@jeremypw Thanks for the detailed review and for pointing out the root cause! 

You are completely right. `Sidebar` is a `Gtk.Grid` container and shouldn't have `can_focus = true`. The issue with the focus not reaching the actual treeview happens because calling `grab_focus()` on the `SourceList` (which is a `Gtk.ScrolledWindow`) doesn't automatically pass the focus down to its child `Gtk.TreeView`

I have addressed the requested changes:-.
1. Removed `can_focus = true;` from `Code.Sidebar`
2. Overridden `grab_focus()` in `Code.Widgets.SourceList` to explicitly call `tree.grab_focus()`, ensuring the underlying `Gtk.TreeView` gets and retains the final keyboard focus

Regarding the shortcut to return focus to the document and the Actionbar tab navigation, I agree that those can be addressed in a separate follow-up PR as enhancements 

Thanks again!

jeremypw:
@Dipanshusinghh Still not working for me I am afraid.  It *is* possible to make the sourcelist have keyboard focus by starting and then cancelling a rename operation on one of the items.  After that you can move the cursor up and down with the keyboard.  However, then pressing the `action-focus-sidebar` shortcut causes the tree to lose keyboard focus for some reason.

src/Widgets/SourceList/SourceList.vala
```diff
@@ -2937,5 +2937,9 @@ public class SourceList : Gtk.ScrolledWindow {
 
         return null;
     }
+
+    public override void grab_focus () {
```
jeremypw: This needs to be `public new void grab_focus ()`

src/Widgets/Sidebar.vala
```diff
@@ -195,4 +195,12 @@ public class Code.Sidebar : Gtk.Grid {
     public void notify_cloning_success () {
         cloning_success_toast.send_notification ();
     }
+
+    public void focus_sidebar () {
+        if (stack.visible_child != null) {
+            stack.visible_child.grab_focus ();
```
jeremypw: You need to check the visible child is a SourceList and then cast it to that type in order to ensure the modified grab_focus method works as expected.

jeremypw (changes requested):
A have made a couple of suggestions that should get this working (as long as there is a selected item in the sidebar at least.

Dipanshusinghh:
@jeremypw I have updated the PR with the exact changes you suggested
1. Changed `grab_focus()` in `SourceList.vala` to `public new void grab_focus()` to properly shadow the parent container method
2. Updated `focus_sidebar()` in `Sidebar.vala` to explicitly check `if (stack.visible_child is Code.Widgets.SourceList)` and cast it before calling `grab_focus()`. This successfully delegates the final keyboard focus to the internal `Gtk.TreeView`

Tested and verified. Thanks again for the excellent guidance and help in getting this right

jeremypw (changes requested):
This works as expected now - thanks!  I think we ought to implement the reverse action in this PR though on second thoughts.

There are some other issues with keyboard focus control - in particular switching from doc to terminal and back but they can wait. 

I notice that once you are out of the doc/terminal focus pair, other widgets can be reached with the standard Tab/Shift-Tab keys, so that's good.

jeremypw:
You need to merge in the updated master branch - I cannot do that.

Dipanshusinghh:
@jeremypw I have updated the PR as requested:

1. Added `ACTION_FOCUS_DOCUMENT` (`action-focus-document`) bound to `<Control><Alt>Right` in `MainWindow.vala` to implement the reverse action (returning focus from the project sidebar back to the active text document).
2. Merged the updated `master` branch into `feature/sidebar-focus`.
Everything is fully up-to-date and ready to merge!

Dipanshusinghh:
Hi @jeremypw, just checking the current status of this PR.

Since the latest updates (including the recent fixes in 3f704fe / 0d87050) have been applied and the functionality seems to be working as expected,
 I wanted to ask if there is still any remaining blocker preventing this from being merged.
If there are any issues, concerns, or improvements needed in the current implementation, 
please let me know — I’m happy to address them and will make the required changes promptly.
Thanks again for your time and review!

jeremypw (approved):
Code looks good and works as expected - thanks!

Issue #1699, opened 2026-02-24
Trailing whitespace is stripped from string literal  containing a line break


jeremypw:
### What Happened?

Using a string literal such as 
```
var test = """This is a string literal ................            
.............with trailing space""";

(where .... represents whitespace) with strip trailing space on results in the trailing space after "literal" to be removed which changes the value
of the string.

### Steps to Reproduce

See above

### Expected Behavior

Trailing white space between triple quotes should not be stripped.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

Dipanshusinghh:
@jeremypw  I think this likely needs the trailing-whitespace cleanup logic to skip lines that are inside multiline string literals (`""" ... """`), since in this case the whitespace is part of the actual string value rather than editor formatting. I
’ll take a look through the strip-trailing-space path locally and see where that distinction can be handled safely.

PR #1720, opened 2026-05-20
Fix 1698

### LINKED ISSUE #1698: Stripping of trailing whitespace affects only some files; no UI indication which types
### What Happened?

The AppMenu has an option "Strip trailing whitespace on save" which implies that this happens for all documents, which is not the case.

### Steps to Reproduce

Activate the stripping option
Create a `.txt` file with trailing whitespace
Save it
Observe whitespace not stripped

### Expected Behavior

The description of the option should include an indication of which file types it applies too or, better, there should be an option for the user to specify which types it applies to (see #617)

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

Dipanshusinghh:
Fixes #1698 

Preferences:- Exclude Markdown and YAML from whitespace stripping
This fix the trailing whitespace stripping behavior on save:--
- Exclude Markdown and YAML files where trailing whitespace can be semanticly meaningful.
- Adds description/subtext in settings switch to clarify that it only apply to syntax-highlighted files, accept Markdown and YAML.
Fixes #1698../


Dipanshusinghh:
Hi @jeremypw,

I've implemented the exclusion list and updated the settings switch description to list Markdown and YAML as exclusions, matching the 'smart cut/copy' style.

Btw, earlier I was planning to solve both #1698 and #1699 together, but decided to separate them to keep the PR clean. I will submit the PR for #1699 after this is reviewed.
Really enjoying working on elementary OS issues, it is a great learning experience. Looking forward to contributing more.
Let me know if any changes are needed. Thanks!

jeremypw:
I agree those two issues should be considered in separate PRs, they are not sufficiently interdependent to deal with together.  Will review this one shortly.

Dipanshusinghh:
will create separate PRs for both issues to avoid any conflicts. Thanks for the feedback, looking forward to your review!

jeremypw (changes requested):
This works as expected.

The only thing I am unsure about is the wording of the label subtext.  "Syntax highlighting" may not be turned on and some source languages may not have any visible highlighting yet trailing spaces are still stripped.  It may be sufficient just to say "Except Plain Text, MarkDown and YAML"?

Dipanshusinghh:
@jeremypw 
Updated the wording to make it clearer and avoid the syntax highlighting ambiguity. Thanks for catching that, the simpler phrasing makes more sense here.

Let me know if there’s anything else that should be adjusted.

Dipanshusinghh (commented):
This looks good overall and the behavior works as expected.

Only small thing i was wondering about is the wording of the subtext. Since trailing whitespace stripping can still apply even when syntax highlighting is not visibly enabled, maybe something simpler like:

`Except Plain Text, Markdown and YAML`

might describe the behavior a bit more accurately.

The actual exclusion logic in `Document.vala` looks clean to me.

jeremypw:
> This looks good overall and the behavior works as expected.
> 
> Only small thing i was wondering about is the wording of the subtext. Since trailing whitespace stripping can still apply even when syntax highlighting is not visibly enabled, maybe something simpler like:
> 
> `Except Plain Text, Markdown and YAML`
> 
> might describe the behavior a bit more accurately.
> 
> The actual exclusion logic in `Document.vala` looks clean to me.

You appear to have reviewed your own PR?  Is there some AI involvement here? Note that elementaryos only accepts manually written code.  Presumably AI assistance for review and translation is OK though.

jeremypw:
I don't want to discourage you contributing, which is welcome, just clarify elementaryos policy in this area.

Dipanshusinghh:
@jeremypw
Sorrry about the confusion — I didn’t mean to make it look like a self-review. I was just trying to repeat the wording suggestion while updating the PR, but I ended up phrasing it badly on my side.

The code and changes in this PR were written by me manually. I do understand and respect elementary OS’s policy around AI-generated code.

And thanks again for the review and feedback throughout this PR — I really appreciate it.

jeremypw:
@Dipanshusinghh Looks like the updated wording has not been pushed to this PR yet?

Dipanshusinghh:
Hi @jeremypw,

srry for the delay, I've updated the wording to "Except Plain Text, Markdown and YAML" now and pushed the change.

Let me know if this looks gud to you. Thannks!

jeremypw (approved):
Looks good now - thanks!

PR #1721, opened 2026-05-23
Fix 1699

### LINKED ISSUE #1699: Trailing whitespace is stripped from string literal  containing a line break

Dipanshusinghh:
Fixes #1699

This PR prevents trailing whitespace from being stripped inside string literals (like multiline strings). Currently, `strip_trailing_spaces()` splits the buffer into lines and runs a regex match to strip whitespace on everything, which changes the semantic value of the strings.

To fix this we check the context class of the iterator at the line end using `iter_has_context_class()`. If the iterator has the `"string"` context class, we skip stripping the trailing space. this is similar to how we already skip stripping inside comments in the `detect-indent` plugin.

### Testing
Tested this locally with multiline string literals in a test file. trailing whitespaces inside triple quotes are now preserved, while normal lines of code stil get striped corectly.


jeremypw (approved):
Nice concise solution! Thanks

Dipanshusinghh:
@jeremypw 
thanks for the review and merge! glad I could help .if anything else comes up or you spot something that could use a fix, feel free to let me know—I'd be happy to contribute.

PR #1722, opened 2026-05-24
Use user accent for selection background

### LINKED ISSUE #1289: Current match styling hard to see with all three stylesheets
### What Happened?

When performing a search all matches are highlighted, the currently selected match is highlighted differently but with all three stylesheets is hard to distinguish due to poor contrast with either unselected matches or with the background.  Where there is a contrast with unselected matches it is the *unselected* matches that are dominant contrary to what would be expected.

Screenshot from 2023-03-25 17 22 38
Screenshot from 2023-03-25 17 22 24
Screenshot from 2023-03-25 17 22 08

### Steps to Reproduce

1. Load a documents with several search matches, one of which is current
2. Load all three stylesheets.

### Expected Behavior

The currently selected match should clearly stand out from both the background and the unselected matches in all three stylesheets.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #1289

The provided user accent colors contrast better with text and hightlight styles than the present hard-coded colors.  Also the user has control over them and it improves consistency with selections elsewhere in the app and other apps.

Gtk already uses the accent color for selections by default so we just stop overriding it.

### DIFF
--- a/data/styles/elementary-dark.xml
+++ b/data/styles/elementary-dark.xml
@@ -26,7 +26,7 @@
 
   <!-- Global Settings -->
   <style name="text"                        foreground="fg-0" background="bg-2"/>
-  <style name="selection"                   foreground="fg-1" background="bg-4"/>
+  <style name="selection"                   foreground="fg-1"/><!-- The user accent color is used for background -->
   <style name="cursor"                      foreground="fg-0"/>
   <style name="secondary-cursor"            foreground="fg-1"/>
   <style name="current-line"                background="bg-1"/>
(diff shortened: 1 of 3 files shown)


zeebok (approved):
Seems reasonable. I see 2 of the tests failed, kind of surprised by that.

jeremypw:
@zeebok Yes - its something to do with building the flatpak in CI though - pretty sure not caused by this PR anyway.  I'll have a look into flatpak building locally but as it is not shipped as flatpak atm it is not critical.

PR #1730, opened 2026-06-16
Gtk4 prep: Avoid using focus event

jeremypw:
In these cases, it is sufficient to listen to changes to the "in-focus" property of the widget (there are no focusable child widgets). This code should compile in Gtk4 without change.

### DIFF
--- a/src/Services/Document.vala
+++ b/src/Services/Document.vala
@@ -256,24 +256,17 @@ namespace Scratch.Services {
 
             this.source_view.buffer.create_tag ("highlight_search_all", "background", "yellow", null);
 
-            // Focus in event for SourceView
-            // Check if file changed externally or permissions changed
-            this.source_view.focus_in_event.connect (() => {
-                if (!locked && !is_file_temporary) {
-                    check_undoable_actions ();
-                    check_file_status.begin ();
+            this.source_view.notify["is-focus"].connect (() => {
+                if (source_view.is_focus) {
+                    if (!locked && !is_file_temporary) {
+                        check_undoable_actions ();
+                        check_file_status.begin ();
+                    }
+                } else {
+                    if (!locked && Scratch.settings.get_boolean ("autosave")) {
+                        save_with_hold.begin ();
+                    }
                 }
-
-                return false;
-            });
-
-            // Focus out event for SourceView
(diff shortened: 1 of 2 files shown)


src/Widgets/SearchBar.vala
```diff
@@ -359,18 +363,12 @@ namespace Scratch.Widgets {
             update_search_widgets ();
         }
 
-        private bool on_search_entry_focused_in (Gdk.EventFocus event) {
-            if (text_buffer == null) {
-                return false;
-            }
-
+        private void on_search_entry_focused_in () requires (text_buffer != null) {
```
danirabbit: Should we also add requires searchentry.isfocus and get rid of the conditional above?

src/Services/Document.vala
```diff
@@ -256,24 +256,17 @@ namespace Scratch.Services {
...
-                if (!locked && Scratch.settings.get_boolean ("autosave")) {
-                    save_with_hold.begin ();
+            this.source_view.notify["is-focus"].connect (() => {
+                if (source_view.is_focus) {
+                    if (!locked && !is_file_temporary) {
+                        check_undoable_actions ();
+                        check_file_status.begin ();
+                    }
+                } else {
+                    if (!locked && Scratch.settings.get_boolean ("autosave")) {
```
danirabbit: Should this not be else if?

src/Services/Document.vala
```diff
@@ -256,24 +256,17 @@ namespace Scratch.Services {
...
-
-                return false;
-            });
-
-            // Focus out event for SourceView
-            this.source_view.focus_out_event.connect (() => {
-                if (!locked && Scratch.settings.get_boolean ("autosave")) {
-                    save_with_hold.begin ();
+            this.source_view.notify["is-focus"].connect (() => {
+                if (source_view.is_focus) {
```
danirabbit, line 261: Seems like both of these require !locked so maybe we should `return_if_reached (locked)`?

danirabbit (changes requested):

jeremypw:
@danirabbit Thanks for the review. I retained that questionable `text_buffer == null` test as changing that is outside the scope of this PR and I am unsure whether it is needed.

danirabbit (approved):

PR #1731, opened 2026-06-16
Gtk4 prep: Avoid connecting directly to key press event

jeremypw:
#### SearchBar: 
Attaching a EventControllerKey to a Gtk.Entry does not work (in Gtk3 at least) as the events are swallowed by the widget regardless of the propgation phase.  So we handle the key presses in the existing MainWindow key controller.

#### Plugins
Similarly key events are swallowed by the source view so we attach a key controller to the window in `hook_window ()`  This has the advantage of not requiring us to disconnect an deactivation but we have to ensure that only key events directed at the sourceview are processed.
Some modification of the handlers was required to avoid accessing Gdk.EventKey and deprecated code such as `event.str`


zeebok (commented):
Does `on_search_entry_focused_in` still get called somewhere or can it be removed?

jeremypw:
@zeebok  `on_search_entry_focused_in` doesn't seem to exist in this PR? When the search entry is focused the `notify["is-focus"]` signal handler will run.

zeebok:
Oops, I think I accidentally renamed `on_search_entry_key_press` in my head when I went to make the comment. But I do see that you remove the function connected to it.

zeebok (approved):

PR #1732, opened 2026-06-18
Gtk4 Prep: Use extra_menu model for sourceview context menu

jeremypw:
As far as possible moves towards the code that will be required for sourceview-5 (Gtk4) with regards to the context menu.  While in Gtk3 we still have to create Gtk.MenuItems as that is what is required for sourceview-4.  However we create these items from an extra_menu MenuModel which in sourceview-5 will be automatically used.  

Actions required by the extra items are moved from MainWindow into the SourceView class which makes more sense. Their associated accelerators are also moved, using a keycontroller.  

Rather than setting the actions enabled/disabled  only on popping up the context menu, they are constantly updated according to various property notifications as will be required for sourceview-5.

Overall this saves a few lines and porting the context menu will require little more than deleting some of these.
 
#### In passing ####
Stripping whitespace is inhibited if there is a selection when the context menu is invoked to avoid the selection being lost when the menu is focused.


src/Widgets/SourceView.vala
```diff
@@ -192,7 +194,67 @@ namespace Scratch.Widgets {
                 }
             });
 
-            populate_popup.connect_after (on_context_menu);
+            // Actions and menumodel for additional context menu items
+            var sort_action = new SimpleAction ("sort-lines", null);
```
danirabbit, line 181: It looks like we already have actions for these that we were using previously. Is there a reason to create new actions instead of reusing the existing ones?

jeremypw: I am in the process of removing those actions and associated functions from MainWindow. It makes more sense to handle them locally, especially in Gtk4.

Note that this pattern has already been used for the Terminal pane and Sidebar for example.

danirabbit (approved):

PR #1733, opened 2026-06-19
Gtk4 Prep: User EventController and PopoverMenu for Terminal context menu

jeremypw:
Use Gtk4 compatible code to show the context menu in the Terminal pane.

No obvious, simple way of having the menu positioned with the top-left corner at the click pointer.  The arrow cannot be removed in Gtk3 but can in Gtk4.


src/Widgets/Terminal.vala
```diff
@@ -146,17 +151,17 @@ public class Code.Terminal : Gtk.Box {
...
-            }
-            return false;
+        secondary_button_controller = new Gtk.GestureMultiPress (terminal) {
+            propagation_phase = CAPTURE,
+            button = Gdk.BUTTON_SECONDARY
+        };
+        secondary_button_controller.pressed.connect ((n, x, y) => {
+            paste_action.set_enabled (current_clipboard.wait_is_text_available ());
+            menu.pointing_to = Gdk.Rectangle () {x = (int)x, y = (int)y, height = 1, width = 1};
+            menu.popup ();
```
danirabbit: We actually need to look for [`event.triggers_context_menu`](https://valadoc.org/gdk-3.0/Gdk.Event.triggers_context_menu.html) and not just blindly connect to secondary button: https://github.com/elementary/granite/blob/1792919cefc83d192f069ec00ab81e2276c43c69/lib/Widgets/ListItem.vala#L166

danirabbit (changes requested):

src/Widgets/Terminal.vala
jeremypw: This is true, although the same flaw already exists in the main code.  According to ValaDoc, the secondary button event *always* triggers a context menu.  So this code will not show an unintended menu.  However, with some configurations the context menu could also be triggered by a different button so really we need to listent to all button events to check whether they should also trigger a context menu.

danirabbit (approved):

PR #1734, opened 2026-06-19
Gtk4Prep: Use button controller and PopoverMenu in SourceList

jeremypw:


jeremypw:
Converting to draft while deal with button release event too.

src/Widgets/SourceList/SourceList.vala
```diff
@@ -2301,15 +2304,14 @@ public class SourceList : Gtk.ScrolledWindow {
                     // Cancel any editing operation going on
                     stop_editing ();
 
-                    if (event.button == Gdk.BUTTON_SECONDARY) {
-                        popup_context_menu (item, event);
-                        return true;
-                    } else if (event.button == Gdk.BUTTON_PRIMARY) {
+                    var button = button_controller.get_current_button ();
+                    if (button == Gdk.BUTTON_SECONDARY) {
```
danirabbit: Probably should be `triggers_context_menu`

danirabbit (changes requested):

src/Widgets/SourceList/SourceList.vala
jeremypw: Ah, I'll have to try and remember that! It seems a bit that we still have to access the raw event even using an eventcontroller :disappointed:

danirabbit (approved):
LGTM 🚀

PR #1736, opened 2026-06-22
Jeremypw/admin/bump copyright date

jeremypw:
Copy rights updated if required according to last commit on each file.  

In some cases not clear when elementary copyright starts - used first commit by recognised elementary dev.

Question: Can/should all copyrights for elementary LLC be transfered to elementary Inc. even if pre-2018?

Format of copyright unchanged - outside intended scope of the PR.  Will be addressed later.


plugins/pastebin/pastebin.vala
```diff
@@ -3,6 +3,10 @@
   BEGIN LICENSE
 
   Copyright (C) 2011-2012 Giulio Collura <random.cpp@gmail.com>
+                2013 Mario Guerriero <mefrio.g@gmail.com>
+                2014 Viko Adi Rahmawan <vikoadi@gmail.com>
```
danirabbit: We shouldn't add additional copyrights here. It's assumed these contributors assigned their copyright to the existing copyright holders

danirabbit (commented):
Yeah LLC copyrights can be changed to Inc since LLC was dissolved and all its assets were transfered to Inc (which includes IP)

plugins/pastebin/pastebin.vala
jeremypw: OK, thanks. In that case does it matter if there is a gap in the dates or should the earlier date range be extended?

jeremypw: I have extended the earlier copyrights where necessary to fill the gap as you say the copyright is assumed to have been assigned to the earlier author.  This commit can be reverted if necessary.

danirabbit (approved):

PR #1738, opened 2026-06-22
Allow secondary windows to have different pane state

### LINKED ISSUE #1737: Changing pane visibility in one window affects all windows
### What Happened?

Second windows are often (mainly?) used to view two or more documents at once i.e. as a read-only view. In these circumstances it is helpful to have maximum area devoted to the document view.  However, hiding the sidebars and terminal also causes them to hide in the main (first) window which may not be what is wanted.

### Steps to Reproduce

Open a second window
Open and close sidebars and terminal
Observe both windows affected

### Expected Behavior

Secondary windows should have independent pane states (not necessarily perstent).

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #1737 

In order to have consistent predictable behaviour, we have to assert a distinction between the first opened (primary) window and any subsequent (secondary) windows.  Only the primary window synchronises with settings with regard to the panes.
This raises the question as to whether secondary windows should be visually different to primary windows but that is left to another time.

The alternative is to treat all windows as equal in which case we would have to have independent settings for each window.


zeebok (approved):
This definitely solves the issue. When you test this, say have panes open in the primary window and hidden in the secondary, and then re-open Code, are the panes all practically minimum size? That happened for me, but if it is only me then go ahead with merging!

jeremypw:
@zeebok Thanks for the review! I don't have a problem with reopening the primary window but I found an issue with the terminal pane in the second window which does not show initially even though it is "on".   I guess the second window should initially follow the pane state of the primary window although it could be argued that it should start with no panes showing (i.e. just showing the chosen document) when it is the result of the "open document in new window" action.

zeebok (approved):

PR #1739, opened 2026-06-22
Gtk4Prep: Replace dialog.run () and filechooser.run ()

jeremypw:
The run () functions is replaced with show () together with a handler for the "response" signal as will be needed in Gtk4.  The dialogs are explicitly created modal.  In some cases where there is code that must wait for the response, it is necessary to use async functionality and `yield` as dialog.show () is not blocking.

#### In passing ####
An unused dialog is removed
`show_all ()` is moved into the Preferences dialog class where it would normally be.


danirabbit:
Any reason you're using `show` instead of present?

jeremypw:
Valadoc says `Gtk.Window.present` should not be used in Gtk3 because of timestamp issues. It will be OK in Gtk4 though.   

Using show () does mean `modal` must explicitly be set to `true` in order to get a modal dialog.  I have now gone through and made all the custom dialogs created modal and shown in the same way.

src/FolderManager/ProjectFolderItem.vala
```diff
@@ -512,7 +513,7 @@ namespace Scratch.FolderManager {
                 dialog.destroy ();
             });
 
-            dialog.run ();
+            dialog.show ();
```
danirabbit, line 517: show is non-blocking so the rest of this block executes immediately. It probably needs to be moved to a separate function so it can be called in response

danirabbit (changes requested):
Don't forget to update POTFILES since you removed a file

src/FolderManager/ProjectFolderItem.vala
jeremypw: Thanks, I missed that.  It is easier to make the global search async and yield while waiting for the response.

jeremypw: For the same reason the cloning action needed some reworking so the "Retry" functionality works properly.

jeremypw: All show () calls should now be followed by a function return, yield statement, or (in the case of the spell plugin) on code that does not depend on the response.

danirabbit:
Since this is a pretty large branch now, can you break out the async function changes into a separate branch? It'd be nice to have its own commit in git history. If it's too much of a pain, don't worry about it :)

jeremypw:
No problem, it has expanded somewhat ...

jeremypw:
Closing as now made into separate PRs

PR #1747, opened 2026-06-24
GtkPrep: Replace deprecated document focus handlers

jeremypw:
#### Focus out #### 
Only strip trailing spaces when keyboard focus lost rather than pointer leaving.  This way we can lose a deprecated event handler and avoid unnecessary work when the pointer accidentally leaves the document.  Moving the pointer into the terminal pane still strips spaces as the terminal grabs focus.

#### Focus in ####
Use the existing notify is-focus handler and just set the doc view current document directly to avoid an unnecessary signal handler


zeebok (approved):
Code looks good and works as intended!

PR #1748, opened 2026-06-24
Gtk4Prep: ChooseProjectButton - Use actionable interface of project row checkbutton

jeremypw:
Avoid button-press-event by using actionable interface.  Saves code anyway.

### DIFF
--- a/src/Widgets/ChooseProjectButton.vala
+++ b/src/Widgets/ChooseProjectButton.vala
@@ -152,14 +152,6 @@ public class Code.ChooseProjectButton : Gtk.MenuButton {
             }
         });
 
-        project_listbox.row_activated.connect ((row) => {
-            var project_entry = ((ProjectRow) row);
-            toplevel_action_group.activate_action (
-                Scratch.MainWindow.ACTION_SET_ACTIVE_PROJECT,
-                new Variant.string (project_entry.project_path)
-            );
-        });
-
         toggled.connect (() => {
             if (active) {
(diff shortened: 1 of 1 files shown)


lenemter (approved):

src/Widgets/ChooseProjectButton.vala
```diff
@@ -152,14 +152,6 @@ public class Code.ChooseProjectButton : Gtk.MenuButton {
             }
         });
 
-        project_listbox.row_activated.connect ((row) => {
```
danirabbit, line 155: I think you still have to connect here because there's padding around the check button

jeremypw: You are right - I didn't notice the peripheral dead area. Its a shame margins cannot automatically be active.

danirabbit: Yeah in GTK4 we can just make the checkbutton itself click-through and save that little bit of code

jeremypw:
Unfortunately there is a flaw in using checkbutton actionable this way - setting the "active" property of the row prior to popping up causes the action to fire :disappointed:  which sometimes causes the popup to close immediately.

I'll see if there is a fix, otherwise maybe better to leave until the port.

danirabbit:
Yeah it's a one-liner to make it click-through so I'd say just wait for the port instead of spending energy debugging

jeremypw:
Ok closing - will fix in port.

PR #1749, opened 2026-06-24
GtkPrep: Avoid deprecated destroy signal

jeremypw:
No good reason to keep a reference to the preferences dialog around.


leonardo-lemos (approved):

danirabbit (commented):
This was to solve an issue where you can open multiple instances of the preferences dialog

jeremypw:
@danirabbit This should not happen now as the dialog is explicitly modal - i'll double check though.

danirabbit:
Oh you're right. My bad. LGTM

jeremypw:
I cannot find anyway to open more than one preference dialog.

PR #1753, opened 2026-06-26
Gtk4Prep: ProjectChooserButton

jeremypw:
- Use Controller instead of button_release_event
 - Define action and target on each listboxrow instead of deprecated get_action_group/activate_action
 - Do not listen to listbox remove signal - the GitManager is source of truth and we already listen to changes there. There is no way to remove rows (projects) through the UI of this widget.

This should work in Gtk4 with minimal changes for porting purposes.

There are no dead areas when clicking on the listboxrows.


src/Widgets/ChooseProjectButton.vala
```diff
@@ -143,23 +138,6 @@ public class Code.ChooseProjectButton : Gtk.MenuButton {
...
-        project_listbox.remove.connect ((row) => {
-            var project_row = row as ProjectRow;
-            var current_project = Scratch.Services.GitManager.get_instance ().active_project_path;
-            if (project_row.project_path == current_project) {
-                Scratch.Services.GitManager.get_instance ().active_project_path = "";
-                // Label and active_path will be updated automatically
-            }
-        });
-
-        project_listbox.row_activated.connect ((row) => {
```
danirabbit, line 154: Rows can be activated in other ways than just button presses, such as with the keyboard. So we really need to keep this

danirabbit (changes requested):

src/Widgets/ChooseProjectButton.vala
jeremypw: Are you sure? The rows are bound to an action now so it doesn't matter how they are activated, the action will fire.

jeremypw: I don't think itts possible for the checkboxes to get focus and change without activating the row but I'll check.

jeremypw: Ok, it is possible (though not obvious how) so I'll make the checkbuttons non-focusable.

jeremypw:
I've fixed it so that it is possible to operate the chooser with the keyboard (once it has been opened) but because of a previously reported css bug (upstream) there is no visible focus on the rows.  But the whole issue of keyboard accessibility needs revisiting after the port anyway.

src/Widgets/Sidebar.vala
```diff
@@ -1,9 +1,9 @@
 /*-
  * Copyright 2017-2026 elementary, Inc. (https://elementary.io)
  *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation, either version 3 of the License, or
+ * This program is free software: you can redistribute it and/or * it under the terms of the GNU General Public License as published by
+
+ modify * the Free Software Foundation, either version 3 of the License, or
```
danirabbit: Seems like this was an accident?

danirabbit (changes requested):
This works I think. It looks like we can get into a broken state with the selection so we should probably revisit this after the GTK4 port since we'll have some more flexibility, but I'm happy with this for now :)

Just one little thing that looks like an accident

src/Widgets/Sidebar.vala
jeremypw: Oops! Now fixed. Thanks for spotting.

danirabbit (approved):

PR #1758, opened 2026-06-28
Gtk4Prep: MainWindow use dialog.show where simple

jeremypw:
Three more simple changes from #1739 

Async stuff will be in final PR


src/MainWindow.vala
```diff
@@ -989,19 +988,22 @@ namespace Scratch {
...
-                    Utils.last_path = Path.get_dirname (uri);
-                    // Open the file
-                    var file = File.new_for_uri (uri);
-                    var doc = new Scratch.Services.Document (actions, file);
-                    open_document.begin (doc);
+            file_chooser.response.connect ((res) => {
+                var uris = file_chooser.get_uris ();
+                file_chooser.destroy (); // Close now so it does not stay open during lengthy or failed loading
+                if (res == Gtk.ResponseType.ACCEPT) {
+                    foreach (string uri in file_chooser.get_uris ()) {
```
danirabbit: It looks like you've cached the list of URIs but then try to get them from the chooser again here?

jeremypw: Oops, thanks for spotting that!

danirabbit (approved):
🚀

PR #1770, opened 2026-07-13
Cleanup: FolderManager File object

jeremypw:
- Inline namespace
 - Inline single use functions
 - Reorder to static, public, private members, create, construct, public methods, private methods
 - Make path construct only
 - Set GLib.File only in construct clause
 - Load info on construct, including executable attribute
 - Start to make naming less ambiguous

### DIFF
--- a/src/FolderManager/File.vala
+++ b/src/FolderManager/File.vala
@@ -105,7 +105,7 @@ public class Scratch.FolderManager.File : Object {
     public bool is_executable {
         get {
             // We queried this attribute on construction
-            if (info.has_attribute (FileAttribute.ACCESS_CAN_EXECUTE)) {
+            if (info != null && info.has_attribute (FileAttribute.ACCESS_CAN_EXECUTE)) {
                 return info.get_attribute_boolean (FileAttribute.ACCESS_CAN_EXECUTE);
             }
 
--- a/src/FolderManager/File.vala
+++ b/src/FolderManager/File.vala
@@ -1,216 +1,202 @@
 /*-
- * Copyright (c) 2017-2025 elementary LLC. (https://elementary.io),
- *               2013 Julien Spautz <spautz.julien@gmail.com>
- *
- * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


src/FolderManager/File.vala
```diff
@@ -1,216 +1,202 @@
...
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranties of
- * MERCHANTABILITY, SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR
- * PURPOSE. See the GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- *
- * Authored by: Julien Spautz <spautz.julien@gmail.com>, Andrei-Costin Zisu <matzipan@gmail.com>
+* Copyright (c) 2017-2025 elementary LLC. (https://elementary.io),
```
danirabbit, line 2: If we're messing with the header can we make it a SPDX header?

src/FolderManager/File.vala
```diff
@@ -1,216 +1,202 @@
...
-                } else {
-                    _name = info.get_display_name ();
-                }
-
+        return strcmp (a.path.collate_key_for_filename (),
+                       b.path.collate_key_for_filename ());
+    }
+    // returns the path the file
+    public string path { get; construct; }
+    public GLib.File file { get; private set; }
```
danirabbit, line 38: Do we need to watch for file path changes? Was this perhaps the way it was before to account for that?

jeremypw: The FolderManager.FolderItem objects have a monitor for this purpose.

src/FolderManager/File.vala
jeremypw: Used SPDX header, keeping LGPL license for now.

jeremypw:
Changed the header for FolderItem too by mistake, but left it as it is needed anyway.

danirabbit (approved):

PR #1772, opened 2026-07-20
FormatBar: clean up widget construction

danirabbit:
This moves popover construction to `construct` which lets us reduce the scope of some widget declarations. While here, also used latest code style and set missing mnemonic_widget for screen reader support


jeremypw (approved):
Everything seems to work as before. There is a small bug but also present in `master` so I'll fix that separately

PR #1773, opened 2026-07-21
Cleanup: Simplify open document

### LINKED ISSUE #1703: Document opened with range does not scroll to show selected range

jeremypw:
Fixes https://github.com/elementary/code/issues/1703

 - Move creation of Document objects into DocumentView and change first parameter to string (path)
 - Only create new objects when needed
 - Update calls to this function appropriately
 - Lose unneeded intermediate functions in MainWindow and call the Document View function directly 
 - Simplify the open_document function by DRYing and treating all documents the same regarding focus, cursor_position and range, whether newly created or existing .

The foldermanagerview member of window is temporarily made public but after  #1778 this will not be necessary as that function will be moved to GitManager


danirabbit:
@jeremypw can you resolve conflicts please

jeremypw:
Converting back to draft for testing after tricky merge

src/Widgets/DocumentView.vala
```diff
@@ -293,78 +293,98 @@ public class Scratch.Widgets.DocumentView : Gtk.Box {
...
 
-    public async void open_document (Services.Document doc, bool focus = true, int cursor_position = 0, SelectionRange range = SelectionRange.EMPTY) {
-       for (int n = 0; n <= docs.length (); n++) {
-            var nth_doc = docs.nth_data (n);
-            if (nth_doc == null) {
+    // Open document from path to avoid unnecessarily creating Document objects and
+    // to confine Document creation to this class.
+    // Documents are treated the same regarding the focus, cursor_position and range parameters
+    // whether already open or not.
+    // Cursor position may have any value > -2 to set cursor. Lower values are ignored (default)
```
danirabbit, line 325: Why did this change to -2?

src/Widgets/SourceView.vala
```diff
@@ -67,13 +67,17 @@ namespace Scratch.Widgets {
...
 
             set {
-                Gtk.TextIter iter;
-                buffer.get_iter_at_offset (out iter, value);
-                buffer.place_cursor (iter); //Assume invalid offset handled correctly for now
-                Idle.add (() => {
-                    scroll_to_iter (iter, 0.25, false, 0, 0);
-                    return Source.REMOVE;
-                });
+                if (value > -1) {
```
danirabbit: can we write this as an early return instead of nesting?

src/Widgets/DocumentView.vala
jeremypw: Because `-1` and `0` are valid values, sending the cursor to the start or end of the document. So I chose `-2` or anything more negative to mean `leave the cursor where it is`.

src/Widgets/SourceView.vala
jeremypw: Done

src/Widgets/DocumentView.vala
danirabbit: Ah that's interesting. Can you document that? I didn't find that choice immediately obvious

jeremypw: I learnt this from `https://valadoc.org/gtk+-3.0/Gtk.TextBuffer.get_iter_at_offset.html` which is what the sourceview uses to place the cursor.   The interpretation of a position <= -2 is mine as it could be useful to not specify a cursor position.  I'll make the comment clearer.

jeremypw: Looking at it afresh it looks like I set the cursor position twice. I'll fix that.

jeremypw: Actually the "-2" value of the parameter is only used internally when a range override present.

jeremypw:
Probably to much to review - I'll try and split up.

zeebok:
> Probably to much to review - I'll try and split up.

@jeremypw  I don't think that is case. I was about to review it, but if you really want to break it up go ahead.

jeremypw:
@zeebok  Sorry, I have reset it as "ready".  Thanks for having a look at this.

src/Widgets/DocumentView.vala
```diff
@@ -293,78 +293,97 @@ public class Scratch.Widgets.DocumentView : Gtk.Box {
...
+        int cursor_position = -2,
+        SelectionRange range = SelectionRange.EMPTY
+    ) {
+        Scratch.Services.Document? doc = null; // The document to be created and/or focused
+        // Check whether document already open
+        bool found = false;
+        for (int n = 0; n < docs.length (); n++) {
+            doc = docs.nth_data (n);
+            //TODO Is this check necessary?
+            if (doc == null || doc.file == null) {
```
zeebok, line 339: I feel like if we are checking any doc of being null before adding it to the list of docs, or the view, we probably don't need the check

src/Widgets/SourceView.vala
```diff
@@ -67,9 +67,16 @@ namespace Scratch.Widgets {
             }
 
             set {
+                // Values of -1 and greater than number of characters
+                // in the buffer place the cursor at the end of the buffer.
+                // Ignore lower values
+                if (value <= -2) {
```
zeebok, line 73: Do we want to ignore this, or clamp it? What happens if the cursor position is larger than the document supports? We may want to just mimic that.

zeebok (approved):
Just a couple thoughts but it seems to work well with no found regressions!

src/Widgets/SourceView.vala
jeremypw: I am relying on https://valadoc.org/gtk+-3.0/Gtk.TextBuffer.get_iter_at_offset.html which says that `TextBuffer.get_iter_at_offset (out iter, int char_offset)` already handles any offset from -1 upwards (even if beyond the end of the buffer). The value of -2 is our own special case for "leave the cursor alone" used for a range override.

src/Widgets/DocumentView.vala
jeremypw: You are probably right about `doc`; not so sure about `doc.file` - there is nothing to stop this being set to `null` by an external call as it is public.  This will be the subject of future cleanup PRs.  We should not need to keep a separate list of docs - we can already iterate over the tabpage children if needed.  We should make `doc.file` private and stop it becoming `null` then the check can be removed.  I'll leave this for now as outside scope.

Thanks for the review!

PR #1774, opened 2026-07-21
FormatBar: don't subclass MenuButton

danirabbit:
MenuButton is sealed in GTK4


jeremypw (approved):
Looks good and works as before

PR #1778, opened 2026-07-22
Cleanup GitManager: Simplify interface

jeremypw:
Towards an easier porting to Gtk4 when the FolderManager will change significantly

 - Use paths and Glib.Files as parameters for functions instead of FolderManager classes
 - Move some git related functions out of SourceView into Git Manager
 - Also avoid referencing ProjectFolderItem in some other places outside the Sidebar


jeremypw:
Closing pending other changes

PR #1785, opened 2026-08-09
Gtk4Prep: Fuzzy search popover - use ListStore instead of Items array

jeremypw:
ListBox.get_children () will go away in Gtk4 so use a bound ListStore model instead.  This also avoids keeping a separate array of the items.

Hide whitespace when viewing diff

Inspired by @danirabbit 's work on Mail.


jeremypw:
There is a pre-existing issue when using the arrow keys to change the selection such as to cause scrolling - I'll try to address this after this PR is merged.

danirabbit (approved):
LGTM, nice work!

PR #1789, opened 2026-08-10
FuzzySearchPopover: use native keynav features

danirabbit:
* Use psuedo selectors to style keyboard focused row
* `selection_mode = BROWSE` because we want always one selected item
* DO NOT set `can_focus` false without an extremely good reason! This is almost always a bad idea and breaks a11y
* Use Gdk.EVENT constants
* Use `move_cursor` and built-in functions like TAB_BACKWARD

### DIFF
--- a/data/Application.css
+++ b/data/Application.css
@@ -45,7 +45,7 @@ textview.scrubber {
     background-color: transparent;
 }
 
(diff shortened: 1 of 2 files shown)


jeremypw (approved):
Much better and fixes a bug in keyboard navigation I noticed before.

PR #1791, opened 2026-08-10
FuzzySearch/FileItem: use arrow, account for RTL

danirabbit:
Use arrow as a delimiter instead of circle and make sure to put things in correct order for RTL


jeremypw (approved):
Looks better, thanks

PR #1793, opened 2026-08-10
FuzzySearch: fix file names

danirabbit:
Fix file names to the modern naming convention. Add missing popover to POTFILES


jeremypw (approved):
Makes sense

PR #1775, opened 2026-07-21
FormatBar: use Action for lang state

danirabbit:

PR #1796, opened 2026-08-13
Jeremypw/format bar/use language action

jeremypw:
Attempt to finish #1775 after merging master and fixing conflicts

For some reason that is not clear to me, the list row widget only behaved (nearly) as expected if a CheckButton is used instead of an ungrouped RadioButton in that the state reflects whether the target matches the current language id on startup.  With ungrouped RadioButtons they are all on.

There is currently a flaw in that clicking on a selected checkbutton deactivates it but does not change the current language.  We may have to revert to grouped RadioButtons?

A number of other small fixes and cleanups are made which could be merged even if this is abandoned.


jeremypw:
@danirabbit I spent quite a while trying to get this to work and this is the best I can do.  Grouped radiobuttons emit a spurious state changed signal on the deselected item it seems and it is hard to avoid endless loops.  Hopefully it will work better in Gtk4.

jeremypw:
Closing as there does not seem to be any particular advantage to doing this, at least prior to Gtk4

PR #1802, opened 2026-08-19
FuzzySearch: remaining cleanup from #1788

jeremypw:
The remaining cleanups seem to be mainly reordering, reformatting and losing some unnecessary members.  I have removed some additional unused/unnecessary members.

It is noted that the placeholder text in the entry is never shown since the entry is always in focus so the initial view appears rather empty.   Maybe consider adding a place holder in the scrolled window, which after this PR remains in view even when empty.


jeremypw:
@danirabbit I was just going to merge master into #1788 but it proved complicated so I made this PR instead.to see what else needed doing.

PR #1807, opened 2026-08-24
SearchBar: use actions

danirabbit:
Make sure we never use UI as a source of truth. Use actions and set those actions as enabled. Use settings as source of truth where possible

### DIFF
--- a/src/Widgets/SearchBar.vala
+++ b/src/Widgets/SearchBar.vala
@@ -66,6 +66,10 @@ public class Scratch.Widgets.SearchBar : Gtk.Box { //TODO In Gtk4 use a BinLayou
      * "Down", it will go at the start of the file to search for the content
      * of the search entry.
      **/
+    private Granite.SwitchModelButton cycle_search_button;
+    private Gtk.ComboBoxText case_sensitive_search_button;
+    private Granite.SwitchModelButton regex_search_button;
+    private Granite.SwitchModelButton whole_word_search_button;
     private Gtk.SearchEntry search_entry;
     private Gtk.SearchEntry replace_entry;
     private Gtk.Label search_occurence_count_label;
@@ -134,9 +138,9 @@ public class Scratch.Widgets.SearchBar : Gtk.Box { //TODO In Gtk4 use a BinLayou
             )
         };
 
-        var cycle_search_button = new Granite.SwitchModelButton (_("Cyclic Search"));
+        cycle_search_button = new Granite.SwitchModelButton (_("Cyclic Search"));
 
-        var case_sensitive_search_button = new Gtk.ComboBoxText ();
+        case_sensitive_search_button = new Gtk.ComboBoxText ();
         case_sensitive_search_button.append ("never", _("Never"));
         case_sensitive_search_button.append ("mixed", _("Mixed Case"));
         case_sensitive_search_button.append ("always", _("Always"));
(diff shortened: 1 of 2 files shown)


src/Widgets/SearchBar.vala
```diff
@@ -588,15 +583,18 @@ public class Scratch.Widgets.SearchBar : Gtk.Box { //TODO In Gtk4 use a BinLayou
...
         update_search_label_timeout_id = Timeout.add (100, () => {
             update_search_label_timeout_id = 0;
             if (search_context == null) {
                 debug ("update occurrence with null context");
                 replace_tool_button.sensitive = false;
                 replace_all_tool_button.sensitive = false;
-                tool_arrow_up.sensitive = false;
-                tool_arrow_down.sensitive = false;
+                find_next_action.set_enabled (false);
+                find_previous_action.set_enabled (false);
```
jeremypw, line 541: I believe using local variables inside a closure causes a memory leak?  Best to get the actions when needed in this thread or use class members (static?) instead of local variables.

src/Widgets/SearchBar.vala
```diff
@@ -588,15 +583,18 @@ public class Scratch.Widgets.SearchBar : Gtk.Box { //TODO In Gtk4 use a BinLayou
 
     // Update search occurrence label, tool arrows and replace buttons in sync
     private void update_search_widgets () {
+        var find_next_action = Utils.action_from_group (MainWindow.ACTION_FIND_NEXT, window.actions);
+        var find_previous_action = Utils.action_from_group (MainWindow.ACTION_FIND_PREVIOUS, window.actions);
```
jeremypw: These could be class members, static if possible - see comment regarding closure.

src/Widgets/SearchBar.vala
```diff
@@ -81,20 +81,32 @@ public class Scratch.Widgets.SearchBar : Gtk.Box { //TODO In Gtk4 use a BinLayou
...
+
     public SearchBar (MainWindow window) {
         Object (window: window);
     }
 
     construct {
+        var replace_action = new SimpleAction ("replace", null);
+        replace_action.activate.connect (action_replace);
+
+        var replace_all_action = new SimpleAction ("replace-all", null);
```
jeremypw: Could use constants for action names to avoid retyping strings

jeremypw (changes requested):
Everything seems to work as expected.  Just a couple of comments about the code.

danirabbit:
@jeremypw went down the rabbithole of fixing the find_next and find_previous actions and we can save a bunch of keypress handling code as well if we add those actions into the search bar. Let me know if this is getting out of hand and should be done in separate steps

src/Widgets/SearchBar.vala
```diff
@@ -92,33 +118,25 @@ public class Scratch.Widgets.SearchBar : Gtk.Box { //TODO In Gtk4 use a BinLayou
...
-
-        tool_arrow_down = new Gtk.Button.from_icon_name ("go-down-symbolic", SMALL_TOOLBAR) {
-            sensitive = false,
+        var tool_arrow_down = new Gtk.Button.from_icon_name ("go-down-symbolic", SMALL_TOOLBAR) {
+            action_name = ACTION_PREFIX + ACTION_FIND_NEXT,
             tooltip_markup = Granite.markup_accel_tooltip (
-                app_instance.get_accels_for_action (
-                    MainWindow.ACTION_PREFIX + MainWindow.ACTION_FIND_NEXT
-                ),
+                {"<Control>g"},
```
jeremypw: Not sure about having a hard copy of the accelerator string here.  Is there a reason for this? I doubt it will change but still.

danirabbit: Yeah because otherwise it picks up just the local accels and not the global one. I can add a comment if you'd like :)

jeremypw (commented):
This is definitely the way to go and in line with what has already been started in Terminal i.e. move stuff out of MainWindow into the relevant widget where possible.

I made a comment but I will test this further tomorrow as the changes are now more substantial.

src/Widgets/SearchBar.vala
jeremypw: Why does it pick the local accels when you have specified the MainWindow action group?

danirabbit:
I'm gonna break up the settings changes to another branch to make it a little more digestible

src/Widgets/SearchBar.vala
jeremypw: Might be better to disambiguate the action names

danirabbit: Oops yeah I messed that up, fixed

jeremypw (approved):
All seems to work.  

It feels a little odd duplicating action in SearchBar and MainWindow but I guess that is unavoidable in Gtk3 and still saves some code.

PR #1808, opened 2026-08-25
SearchBar: set method → properties, rm unused

danirabbit:
* `reveal ()` and `is_revealed { get; }` become [`search_mode_enabled { get; set; }`](https://valadoc.org/gtk4/Gtk.SearchBar.search_mode_enabled.html)
* `set_search_entry_text ()` and `entry_text { get; } ` become `search_text { get; set; }`
* Remove one line `search_is_focused { get; }` and `replace_is_focused { get; }` which are only used in one place anyways and `on_text_buffer_changed ()` which just calls `update_search_widgets ()`
* Remove unused `transition_time_msec { get; }` and `highlight_none ()`

### DIFF
--- a/src/Widgets/SearchBar.vala
+++ b/src/Widgets/SearchBar.vala
@@ -12,42 +12,33 @@ public enum Scratch.CaseSensitiveMode {
 }
 
 public class Scratch.Widgets.SearchBar : Gtk.Box { //TODO In Gtk4 use a BinLayout Widget
-
     public weak MainWindow window { get; construct; }
 
     public bool is_focused {
         get {
-            return search_is_focused || replace_is_focused;
+            return search_entry.has_focus || replace_entry.has_focus;
         }
     }
 
-    public bool search_is_focused {
-        get {
-            return search_entry.has_focus;
-        }
-    }
-
-    public bool replace_is_focused {
-        get {
-            return replace_entry.has_focus;
-        }
-    }
-
-    public bool is_revealed {
+    public bool search_mode_enabled {
         get {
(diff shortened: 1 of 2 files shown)


src/Widgets/SearchBar.vala
```diff
@@ -389,24 +368,12 @@ public class Scratch.Widgets.SearchBar : Gtk.Box { //TODO In Gtk4 use a BinLayou
...
-        }
-    }
-
-    public void set_search_entry_text (string text) {
-        search_entry.text = text;
-    }
-
     private bool on_key_pressed (uint keyval, uint keycode, Gdk.ModifierType state) {
-        if (!(search_is_focused || replace_is_focused)) {
+        if (!(search_entry.has_focus || replace_entry.has_focus)) {
```
jeremypw, line 372: Can this be replaced by `if (!is_focused) {`?

jeremypw (changes requested):
Works as before afaict.  Just a small code suggestion

src/Widgets/SearchBar.vala
danirabbit: It could but I'm not really a fan of using external API internally when it's just a one-liner like this. If the public function is removed then we'll just have this one liner function hanging on until someone else comes to clean it up again and I don't think it really makes the code more clear personally. But if you feel strongly about it, I'll change it

jeremypw: OK as it is done for a reason I withdraw the suggestion.

jeremypw (approved):

PR #1809, opened 2026-08-25
SearchBar: improve settings handling

danirabbit:
* Recycle the same SearchSettings instead of creating a new one and resetting settings every time document changes
* Use Settings as the source of truth, not widget states


jeremypw:
@danirabbit  After merging master it was necessary to retain a reference to `cyclic_search button` to get it to compile.  You might want to see whether this can avoided.

jeremypw (approved):
Looks good, not regression found.

PR #1822, opened 2026-09-21
Fix stale global search result badges

### LINKED ISSUE #1672: Global search should clear badges as soon as the related search term is deleted or changed
### What Happened?

After performing a global search with matches, the search term in the header bar was deleted.  The badges remained but there is no longer any information as to what search they refer.

### Steps to Reproduce

 - Perform a global search with matches
 - Delete the search term
 - Observe the badges remain

### Expected Behavior

The badges should disappear as soon as the search term to which they relate is changed.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

kahhong:
Fixes (#1672) by clearing all global search result badges in the sidebar when the searched text changes.


jeremypw (approved):
Does the job thanks!

I'll hold this back from merging until the pending release (8.4.0) goes out so that does not have to be revised.

jeremypw:
Blocked unto v 8.4.0 goes out.

